PR #12, opened 2016-01-01
Control nunchuk app over UART

RollingGecko:
Added changes allow the usage of the nunchuk app over UART message.
Necessary for https://github.com/RollingGecko/ArduBoardControler.git


vedderb:
Sorry for the late reply. Automatic merge does currently not work because my code has several other changes, but I have added this to the code I'm working on. I also added accelerometer values to the end of the nunchuk packet just in case, but it should work even if they aren't sent. The next time I push the code the nunchuk command will be implemented (hopefully before the weekend).

RollingGecko:
Thx!

PR #15, opened 2016-02-02
added ADC2 analog brake option

derlucas:
i added a rudimental analog braking option to the ADC App. It is using the EXT2 ADC as a second input.
Please excuse the copy&paste coding. We should make the filtering stuff more beautiful.


vedderb:
For filtering it is probably better to use UTILS_LP_FAST. Also, there should be individual voltage start and voltage end options for the second adc in case a different actuator is used.

PR #21, opened 2016-07-30
README: Remove use of "my".

aarossig:
- This makes it more explicit when users fork the project to contribute.


TimGremalm:
TLDR.. this pull request is not readable. What are you up to?
Wouldn't it be better to send the pull request after all you're done?

aarossig:
Oh, thanks for pointing this out. I didn't realize that I left this pull request open (and on master :/ ). I am closing it now.

I have re-written a bunch of stuff and am in the process of cleaning up the codebase. After hitting a "Toyota bug" with uncontrolled acceleration, I decided to take things into my own hands. You can reproduce it by removing the battery from the Nunchuk remote while at max throttle.

My new design is more modular and has clear isolation of responsibilities between functional units (LEDs, speed regulation, user input). I just added support for unit tests as well. At least one "Toyota bug" has been resolved via duplicate event detection. I have also added initial support for generic LED animations.

Feel free to follow along. I commit/push often.

TimGremalm:
Hehe okay :)

> I have re-written a bunch of stuff and am in the process of cleaning up the codebase. After hitting a "Toyota bug" with uncontrolled acceleration, I decided to take things into my own hands. You can reproduce it by removing the battery from the Nunchuk remote while at max throttle.
Okay, we don't want bugs like that. I appreciate your work of cleaning it up and isolation, looking forward for the finished pull request :)

aarossig:
The unfortunate situation is that pull requests are difficult. I have worked with the original author and there is a large delta between the most recent commit and what exists on his workstation. The development model needs to be moved to "in the open" as opposed to feature drops via commit.

As an example, the WS2811 driver as it stands has a race condition. The DMA is configured in cyclic mode which means that it is constantly banging out the LED data (unnecessary). When you write to the buffer, there is a chance that the LEDs glitch as the DMA transfer writes out the buffer.

The revised driver in my fork has an explicit SendFrame function. The LED data line is only active for a brief time after SendFrame is invoked: https://github.com/aarossig/bldc/blob/master/drivers/ws2812.h#L68

As long as SendFrame is invoked less often than the time it takes to write out a frame, there will be glitch free LEDs.

I guess what I am saying is that there may not be a pull request. I would love to send my changes upstream, but I think this is rewrite compared to a minor patch here and there. Perhaps I can work more closely with the author if they are willing to integrate my changes upstream.

Here's the board I am working on. It represents about 4 months of my time and I have put about 30 miles on my fork of the codebase :)

img_4332
img_4323

TimGremalm:
>As long as SendFrame is invoked less often than the time it takes to write out a frame, there will be glitch free LEDs.
 I guess what I am saying is that there may not be a pull request. I would love to send my changes upstream, but I think this is rewrite compared to a minor patch here and there. Perhaps I can work more closely with the author if they are willing to integrate my changes upstream.

It sounds like an improvment, I would also like to see it upstream as well.
I think Benjamin Vedder also would appreciate it.

Nice layout you got on your drivers :)

PR #22, opened 2016-07-30
packet: remove rx_timeout

aarossig:
- Saves memory (~136 bytes + thread overhead) and CPU time.
- When the value reaches zero, nothing happens.


aarossig:
I should read more closely. rx_state is assigned 0 when the timeout reaches 0.

PR #23, opened 2016-07-30
ws2811: move the gamma table to flash, remove RGBdata

aarossig:
- No one is calling ws2811_get_led_color. We can free the RGB buffer. It
  is best to let the client of the WS2811 driver decide whether or not
  a framebuffer is required. This saves ~116 bytes of RAM, if you have
  28 LEDs.
- The gamma table can easily sit in flash, which reduces boot time, code
  size and saves 256 bytes of RAM.


ws2811.c
```diff
@@ -36,10 +37,44 @@
 #define WS2811_ZERO			(TIM_PERIOD * 0.2)
 #define WS2811_ONE			(TIM_PERIOD * 0.8)
 
+// Gamma table.
```
kalvdans: Please describe in a comment the formula which generated the table.

ledvinap: gamma table is (almost certainly) wrong - it is non-monotonous ... 

It may be possible to generate it using gcc inlining of floating-point functions, which will constant-fold initializer. It's not strictly standard compliant (`__extension__` can be used to silence warnings), but generated code may be easier to use/modify ... 

```
#define T(i) ((int)roundf(powf((float)i / 255.0, 1.0 / 0.45) * 255.0))
static const uint8_t gamma_table[256] = {
T(0), T(1), T(2), ... 
};
#undef T

aarossig:
Thanks for the review, but I think I'll just abandon this. The values may very well be wrong.

FWIW, I stripped out gamma correction in my fork which is probably why I never noticed the wrong values. For the application of brake lights, turn signals and the occasional animation I realized that it just wasn't necessary.

PR #28, opened 2017-02-04
Terminal pluginhooks

dothebart:
Looking for a cheap way to read debug information from my app, I found the terminal facility.
This PR brings a simple API for apps so they can register functions that can be used from the terminal.
A sample app with a sample callback hook is provided.

### DIFF
--- a/applications/app.c
+++ b/applications/app.c
@@ -74,6 +74,9 @@ void app_init(app_configuration *conf) {
 		hw_stop_i2c();
 		app_sten_init();
 #endif
+#ifdef USE_TERMINAL_DEMO_APP
+		app_test_init();
+#endif
 		break;
 
 	default:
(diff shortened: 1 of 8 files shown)


vedderb:
This is a nice addition to the terminal, and I have integrated it in my local version, so your apps should be compatible with the latest code once I release the huge update that I'm working on.

There is only one thing I changed: Instead of allocating memory during runtime when registering new callbacks I made a static buffer with terminal commands. I have avoided using dynamic allocation throughout the whole source code so that the memory usage is known at compile time and to avoid memory-related surprises during runtime. Every terminal callback only has three pointers anyway, so statically allocating 20 callbacks barely takes any RAM anyway.

dothebart:
Yea, saw that already (that there are no mallocs) Probably haven't done enough embedded coding yet ;-)

dothebart:
Hi, Any hints on when the next release will be?
or, can't you share a devel branch with us so we can base subsequent PRs on that?
Any reason why you didn't add any tags from the previous releases?

dothebart:
f14115735812938ccd6142bbdcb7e1799c83d69a added this functionality - thus closing the PR.

PR #29, opened 2017-02-12
don't mix blanks and tabs.

dothebart:
While I personally prefer not to use tabs at all (view varies per editor config) this commit flips the header to be all tab instead of mostly tab.


b264:
I was wondering why also LoL

PR #41, opened 2017-09-18
Update mc_interface.c

Ackmaniac:
Chang maximal brake current calculation so that brakes behave like in the previous version. Because actual code results in unpredictable brakes when motor min and battery min isn't close to each other.


vedderb:
I tested, and the implementation should work. I have heard that some people have problems with this, so for now I have disabled the function completely.

PR #43, opened 2017-09-21
stop script on errors

rpasichnyk:
I added 2 small fixes to the rebuild_all bash script

1) Stop execution of the script if any error occurs (set -e)
2) After executing the script return to the initial directory with popd. For example, if I am in my home directory and run ~/dev/bldc/build_all/rebuild_all, I do not want to end up in bldc/build_all, I want to stay in my home directory after the script finishes

### DIFF
--- a/build_all/rebuild_all
+++ b/build_all/rebuild_all
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e
 
 FWPATH="../"
 DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )


kalvdans:
1. Agree, ``set -e`` is a good thing.

2. The popd has no effect as the script is run in a subshell. Unless you source the script (``. ~/dev/bldc/build_all/rebuild_all``) which will have bad effects anyway, as ``set -e`` will affect your outer shell.

rpasichnyk:
Thanks @kalvdans you are right. I removed pushd / popd change.

kalvdans:
Thanks, looks good to me now!

PR #54, opened 2018-04-04
added proof of concept for detecting the precense of an AS5047 encoder

Shackmeister:
perhaps it should also generate an error after x wrong parities


vedderb:
Something similar is implemented in the last commits. Hopefully it is close to what you had in mind.

PR #58, opened 2018-04-25
Fix overflow when receiving CAN message with large ID

mtzm:
If noticed that the enum CAN_PACKET_ID uses only 1 byte.
This leads to an overflow if CAN messages with large IDs are received from other devices on the same bus.
I don't know how to define the datatype of an enum in C, so I used a workaround (adding a large enumerator at the end of the list). I also added a static assertion to check that CAN_PACKET_ID is 4 bytes long.


mtzm:
Alternatively, one could also split up the enum and typedef to avoid adding a large enumerator at the end of the list.Like:
```
enum {
	CAN_PACKET_SET_DUTY = 0,
	...
};

typedef uint32_t CAN_PACKET_ID;
```

PR #60, opened 2018-05-07
Fix race condition between different command sources

Vajnar:
Since the VESC is able to receive motor commands from numerous sources (e.g. USB, PPM, ...) and there is currently no prioritization of the sources, there can be a race condition between them and if the same command arrives from multiple sources, then the action performed could be chosen at random.

This proposal prefers commands received from the RF transceiver (PPM input) over the other options. I use this behavior in order to automatically switch between the autonomous driving mode (commands send through the USB via the Robot Operating System VESC wrapper - https://github.com/mit-racecar/vesc) of a 1:10 scale model car and a manual override mode (PWM signal generated by a dedicated MCU based on change of output from the RF remote controller) for handling emergency situations (bug in code, system freeze, ...).

Perhaps this change could be useful for other VESC users as well.

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -49,6 +49,8 @@
 #include <stdarg.h>
 #include <stdio.h>
 
+#define SERVO_TIMEOUT	30    /* in ms */
+
 // Threads
 static THD_FUNCTION(detect_thread, arg);
 static THD_WORKING_AREA(detect_thread_wa, 2048);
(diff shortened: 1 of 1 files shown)


commands.c
```diff
@@ -49,6 +49,8 @@
 #include <stdarg.h>
 #include <stdio.h>
 
+#define SERVO_TIMEOUT	30
```
wentasah: Can you add a comment specifying the units of this value?

PR #65, opened 2018-10-18
Improved position controller with multi-turn position & speed

raess1:
We are working on an updated/modified version of original firmware for the VESC (based on FW3.33).

Key Features
1: Multi-turn posistion control (controlling motor position over multiple encoder turns, for example 0-2160Degress)
2: Removed shortest route rotation -+180 degress
3: Handle encoder rollover up or down.
4: First test of implement a cascaded style position, velocity and current control loop. The idea was previously tested on our simulator for cascaded position and speed control (http://grauonline.de/alexwww/motorsim/motorsim.html)


Teslafly:
This probably fixes https://github.com/vedderb/bldc/issues/319

Teslafly:
Looks like the build folder accidentally got added to this pr. That needs to be removed.

All development is now happening in the dev branches. In this case the dev_fw_5_03 branch. update this to be applied to dev_fw_5_03.

akrv:
will this be merged anytime soon?

Teslafly:
> will this be merged anytime soon?

No. It needs significant cleanup and rebasing at the very least. The pull request hasn't been updated since 2018.

There has also been a reluctance signaled to implement "Removed shortest route rotation -+180 degrees". Despite several requests. There is a workaround, but it is sub-optimal. see https://github.com/vedderb/bldc/issues/319

I agree that it would be really nice to be able to command a rotation of 1000 degrees and the motor rotates ~3 times. You don't always have a specific angle division number you know beforehand.

tobiokanobi:
> I agree that it would be really nice to be able to command a rotation of 1000 degrees and the motor rotates ~3 times. You don't always have a specific angle division number you know beforehand.

Yes, that would be great!

PR #69, opened 2019-01-30
Fix issue with FOC and sudden stop

DAddYE:
You can find the rationale here: https://www.electric-skateboard.builders/t/need-a-few-testers-with-focbox-unity-builds-low-speed-braking-fix-needs-testing/80201?u=deodand


vedderb:
I tried this a bit, but it simply made the motor start cogging when loosing tracking instead of going into full brake. The fix kind of misses half of the point of the original if statement anyway. After testing a bit I think I do prefer cogging over locked brakes though, and this is what I came up with that works the smoothest after some testing, including when using sensors:

```
static bool was_full_brake = false;
if (m_control_mode == CONTROL_MODE_CURRENT_BRAKE &&
		fabsf(duty_filtered) < m_conf->l_min_duty * 1.5 &&
		(m_motor_state.i_abs * (was_full_brake ? 1.0 : 1.5)) < fabsf(m_iq_set)) {
	control_duty = true;
	duty_set = 0.0;
	was_full_brake = true;
} else {
	was_full_brake = false;
}
```
It is included in my local version of the firmware, which I will push after finishing my current major update.

DAddYE:
Thanks so much! I'm going to test your code and report back! 
We really appreciated this fix!

vedderb:
Closing this one, as the fix is in the code now.

PR #71, opened 2019-02-18
Powerdesigns dev

nitrousnrg:
* Provide support to HW_VERSION_PALTA
* More watchdog coverage
* Deadtime can be set in nanoseconds
* Flash is not written unless VDD supply is >2.9V
* BOR keeps MCU under reset until VDD reaches 2.9V
* Travis Continuous Integration
* Updated README


vedderb:
Thanks for the pull request. I had a quick look, and saw the following things:

* Timeout should ignore the CAN-thread when CAN_ENABLE is 0.

* Is conf_general_measure_flux_linkage necessary in addition to the new conf_general_measure_flux_linkage_openloop? Can it be removed?

* Is it necessary to include the ST DAC implementation? Using the DAC without DMA or waveform generation is trivial by just writing directly to the registers. Essentially you only need to:

1) Set up the clock:
RCC_APB1PeriphClockCmd(RCC_APB1Periph_DAC, ENABLE);

2) Enable the channels you need:
DAC->CR |= DAC_CR_EN1 | DAC_CR_EN2;

3) Write data to the channels when you want to update the output:
DAC->DHR12R1 = VALUE1;
DAC->DHR12R2 = VALUE2;

* gpdrive should also use conf_general_calculate_deadtime

Can you have a look at that and update the pull request? If I make the changes it might break something for your hardware, and I cannot test it myself.

nitrousnrg:
Today's commits should fix your comments, you can take a look. Its tested on our hardware.
Thanks!

vedderb:
Thanks a lot, I like the new watchdog implementation and the update README. I went though everything and made a few adjustments, and so far it seems to work. I will do a bit more testing later.

PR #72, opened 2019-02-25
AS5047 encoder parity check

flytrex-vadim:
Hi Benjamin,

This PR adds CRC check on the AS5047 encoder frames, and ignores the messages with incorrect CRC.  When over 1% of messages has error, the fault is triggered. Additionally, 0xffff value is dropped regardless of CRC since this value is read from pull-ups when no encoder is connected.

This protects the motor from overheating when encoder wiring is damaged.


nitrousnrg:
Nice feature!
Next time it would be perhaps more accurate to say parity instead of CRC.
Parity is a special case of crc so its not wrong, but makes me think of stronger 8 bit or 16 bit checks rather than 1bit checks.
Thanks

flytrex-vadim:
> Nice feature!
> Next time it would be perhaps more accurate to say parity instead of CRC.
> Parity is a special case of crc so its not wrong, but makes me think of stronger 8 bit or 16 bit checks rather than 1bit checks.
> Thanks

Yes, you're right. I've been struggling with (8-bit) CRC issues so much lately that the term was stuck in my mind :)

vedderb:
Nice addition! Just a few comments:

* It is not good that the fault code is triggered when the motor is not using the encoder. The fault code should only trigger if the motor is in FOC with sensor mode encoder, or in DC mode doing position control or speed control and the encoder mode is AS5047,

* It is not necessary to check for the encoder fault in the high rate interrupt in mc_interface, doing it from the timer thread is better.

* The error rate should be reset when the encoder encoder_init or encoder_deinit is called, as it would be strange if the fault code is still there when switching between sensor port modes.

Can you have a look at that, or is it easier if I do the updates?

flytrex-vadim:
>    It is not good that the fault code is triggered when the motor is not using the encoder. The fault code should only trigger if the motor is in FOC with sensor mode encoder, or in DC mode doing position control or speed control and the encoder mode is AS5047,

Is there a possibility that the encoder readout will be triggered when the motor is not using the encoder ? The motor mode check seem to be done already before calling encoder_init_as5047p_spi() from mc_interface.  And is SPI readout is not called, then the error rate will stay zero and no fault will be triggered.

> It is not necessary to check for the encoder fault in the high rate interrupt in mc_interface, doing it from the timer thread is better.

Do you mean the timer thread in main.c ?  My thinking was that extra one conditional would be negligible load in mc_interface, but the code readability would be much better - all the other checks are done there.

> The error rate should be reset when the encoder encoder_init or encoder_deinit is called, as it would be strange if the fault code is still there when switching between sensor port modes.

Oh, right, will change.

vedderb:
The readout is always called, as it is used from the encoder timer, which always is active as long as the encoder is active. Even when the encoder is not use, I sometimes use the output it provides to see how well the observer is tracking. Therefore the fault would be triggered as soon as the sensor port is set into AS5047 mode.

I was thinking of the timer thread in mc_interface
https://github.com/vedderb/bldc/blob/master/mc_interface.c#L1691
the other checks are done in the high rate interrupt because they should stop the motor as fast as possible. Some other checks and fault code timeout handling is already done in that timer thread, but no fault code generation as that has not been relevant yet. Also, adding the extra checks regarding whether the fault should be generated will also make it a bit slower, so it is good to keep it in the timer.

In general I have been trying to avoid generating fault codes as much as possible, as it most of the time is more important to keep the motor operational if at all possible. Therefore it would be good if this fault code really only is generated while the motor is using the encoder, and there is no other choice. It might even make sense to disable this fault code in the configuration as it probably often is desired to continue running even if the encoder is problematic - that's a different issue though.

vedderb:
I had a look again, the temperature faults are actually generated from the mc_interface timer_thread from the update_override_limits function.

flytrex-vadim:
I moved the check code into the mc_interface timer thread, and added the commutation mode check as suggested.

I'd rather not take on the setting implementation at the moment  (even though I agree with the idea) due to lack of VESC tool development environment setup, and conflicting priorities :)

Also, a bit off-topic: I have a few commits that fix simultaneous operation of serial and USB channels - as of version 3.40 there is a threading bug that sometime sends replies to a wrong channel. Makes sense to make the fix into a pull request ?

vedderb:
Looks good. I forgot to mention it before, but you should also update the help text for the terminal with the new encoder command.

It would be nice if you can make a pull request for the other fix too :-)

vedderb:
Got it merged now. Just noticed that you mixed spaces and tabs for indentation. It would be good if you can stick to using tabs. I use width 4 for all tabs.

Will give it a try now...

flytrex-vadim:
> Just noticed that you mixed spaces and tabs for indentation. It would be good if you can stick to using tabs. I use width 4 for all tabs.

Oups, sorry about that. Auto-detection messed up in one of my IDEs.

flytrex-vadim:
> It would be nice if you can make a pull request for the other fix too :-)

It seems that some of the issues (mixing callbacks) were already fixed in the great redesign of 2019 :)  I'll check is any bugs that I've seen remain and if so make a pull request.

nitrousnrg:
I took a closer look at this, and a sustained 0x0000 response from the sensor is not an error condition. Neither is 0xFFFF.
It would be an error if an odd parity bit would be implemented, in that case 0b000.0000 would have the parity bit as 1. However this sensor uses even parity so 0x0000 is okay.
It may *likely* be a sign of wiring error, but its not necessarely so and we are introducing a false positive. If we want to detect a wiring issue we should use a different register with known values to compare to. For example write the zero position register with a known value and once in a while read it to ensure the interface works well.

flytrex-vadim:
@nitrousnrg : yes, this is by design
Both 0x0000 and 0xfffff are highly likely indicators of wiring failures. And the cost of "false positives" is negligible since the actual readings from encoder include a fair amount of noise. So even if you try to get a precise 0 reading by rotating encoder to such a position, you will get a stream of values randomly moving around 0. So dropping a zero value has no practical consequences - a next value arriving 20us later will be used.

It would have been nice to be able to read other registers, but it is not possible with the current hardware due lack of MOSI signal. Only the 3fff register is accessible.

vedderb:
I actually removed the case that 0 is an error from my local code yesterday, as I triggered the fault when running position control at position 0. I will push the changes later when I get home.

flytrex-vadim:
> triggered the fault when running position control at position 0

A better fix for this might be just raising the threshold of detection from 1%, maybe all the way to 30-40%. I've really encountered ground shorted wring that received zero readings.

PR #77, opened 2019-03-03
Serial thread safety fix

flytrex-vadim:
The easiest way to reproduce the bug addressed in this PR is to send a terminal command that generates a lot of output, such as "threads" with high frequency via UART channel. When USB is connected, some of the replies go to VESC_tool terminal window.

This is due to the callback function being changed from different threads while commands_send_packet is being executed. My fix removes the use of global variable send_func from as many places as possible and replacing it with stack parameter send_func_p and protecting command function from reentrancy (due to static buffer usage)

Note: NRF and CAN functionality not tested due to lack of hardware.

### DIFF
--- a/BLDC_4_ChibiOS.ebp
+++ b/BLDC_4_ChibiOS.ebp
@@ -1,0 +1,968 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
+<EmBitz_project_file>
+	<EmBitzVersion release="1.11" revision="0" />
+	<FileVersion major="1" minor="0" />
+	<Project>
+		<Option title="BLDC_4_ChibiOS" />
+		<Option makefile_is_custom="1" />
+		<Option pch_mode="2" />
+		<Option compiler="armgcc_eb" />
+		<Build>
+			<Target title="upload">
+				<Option output="build\BLDC_4_ChibiOS.elf" />
+				<Option object_output="build\obj\" />
+				<Option type="0" />
+				<Option compiler="armgcc_eb" />
+				<Option projectDeviceOptionsRelation="0" />
+				<Compiler>
+					<Add option="-O2" />
+					<Add option="-g2" />
+					<Add option="-fdata-sections" />
+					<Add option="-ffunction-sections" />
+				</Compiler>
+				<Linker>
+					<Add option="-Wl,--gc-sections" />
+				</Linker>
+			</Target>
+		</Build>
+		<Compiler>
+			<Add option="-Wall" />
+		</Compiler>
+		<Unit filename="appconf\appconf_custom.h" />
+		<Unit filename="appconf\appconf_default.h" />
+		<Unit filename="appconf\appconf_example_ppm.h" />
+		<Unit filename="applications\app.c">
+			<Option compilerVar="CC" />
+		</Unit>
+		<Unit filename="applications\app.h" />
+		<Unit filename="applications\app_adc.c">
+			<Option compilerVar="CC" />
+		</Unit>
+		<Unit filename="applications\app_brake.c">
+			<Option compilerVar="CC" />
+		</Unit>
+		<Unit filename="applications\app_brake.h" />
+		<Unit filename="applications\app_custom.c">
+			<Option compilerVar="CC" />
+		</Unit>
+		<Unit filename="applications\app_nunchuk.c">
+			<Option compilerVar="CC" />
+		</Unit>
+		<Unit filename="applications\app_ppm.c">
+			<Option compilerVar="CC" />
+		</Unit>
+		<Unit filename="applications\app_sten.c">
+			<Option compilerVar="CC" />
+		</Unit>
+		<Unit filename="applications\app_uartcomm.c">
+			<Option compilerVar="CC" />
+		</Unit>
+		<Unit filename="board.c">
+			<Option compilerVar="CC" />
+		</Unit>
+		<Unit filename="board.h" />
+		<Unit filename="buffer.c">
+			<Option compilerVar="CC" />
+		</Unit>
+		<Unit filename="buffer.h" />
+		<Unit filename="chconf.h" />
+		<Unit filename="ChibiOS_3.0.2\demos\STM32\CMSIS-STM32F407-DISCOVERY\chconf.h" />
+		<Unit filename="ChibiOS_3.0.2\demos\STM32\CMSIS-STM32F407-DISCOVERY\halconf.h" />
+		<Unit filename="ChibiOS_3.0.2\demos\STM32\CMSIS-STM32F407-DISCOVERY\main.c">
+			<Option compilerVar="CC" />
+		</Unit>
+		<Unit filename="ChibiOS_3.0.2\demos\STM32\CMSIS-STM32F407-DISCOVERY\mcuconf.h" />
+		<Unit filename="ChibiOS_3.0.2\demos\STM32\HAL-STM32F407-DISCOVERY\halconf.h" />
+		<Unit filename="ChibiOS_3.0.2\demos\STM32\HAL-STM32F407-DISCOVERY\main.c">
+			<Option compilerVar="CC" />
+		</Unit>
+		<Unit filename="ChibiOS_3.0.2\demos\STM32\HAL-STM32F407-DISCOVERY\mcuconf.h" />
+		<Unit filename="ChibiOS_3.0.2\demos\STM32\HAL-STM32F407-DISCOVERY\osalconf.h" />
+		<Unit filename="ChibiOS_3.0.2\demos\STM32\RT-STM32F407-DISCOVERY-G++\chconf.h" />
+		<Unit filename="ChibiOS_3.0.2\demos\STM32\RT-STM32F407-DISCOVERY-G++\halconf.h" />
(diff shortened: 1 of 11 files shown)


rpasichnyk:
I’ve seen this bug in the wild, but instead of USB/UART mixup it was CAN/UART

flytrex-vadim:
> I’ve seen this bug in the wild, but instead of USB/UART mixup it was CAN/UART

Would you consider testing the fix on your setup ?

rpasichnyk:
Unfortunately right now I don’t have a setup to recreate this issue

kalvdans:
Is it a lot of work to remove the static buffer usage? I'd rather not introduce an unnecessary mutex.

flytrex-vadim:
@kalvdans : in this codebase, the pattern seems to be static buffers + mutex all over. Only the commands module was missing the mutex, so adding it was the minimal change.
And I'm not sure using stack for large buffers is a better idea in any case.

vedderb:
I have known about this for a while...

I don't think it is good to lock the commands, as some commands (e.g. detection via terminal and can scan) can take long to complete. There are also some strange interactions between forwarded can commands and their replies, which are not really solved in an elegant way. I want to give some major polish to how this is done soon anyway, but other things have had higher priority so far. Here are some of the things I want to include in the polish:

* Make the process function as reentrant as possible, without locking it. That includes:
  - Use a stack buffer for short replies, and the global buffer for long replies. The only replies that are a bit too long for the stacks are the configurations.
  - Pass a send function on the stack, that may of may not replace the global one. This is the tricky part, due to the way can replies are implemented. I have to give this some thought.
* Make a print function that takes a send function, and pass the send function to the terminal. Wrap that print function with one that uses the global send function (not entirely sure yet). The idea if the print function is to make it usable for debugging without being aware where the last communication came from, for example when making custom applications.
* Look if there are locks needed. Most of this should be handled by mc interface, but currently a detection can be started from the terminal and from the commands at the same time, which shouldn't be possible.

For now it is probably best if I look into this myself and also do some testing, but I will keep this PR open for now so that we can discuss the issue here.

flytrex-vadim:
What do you think about making all the commands short-running, and delegating all the long-running tasks to other thread ? 

Looking at the code, this is already the case with detection - so the lock is *not held* for the duration of the the operation. 
Did some testing now on my version with the lock, and the detection from VESC tool worked fine while continuously replying to UART commands.

Regarding the print function - indeed my first fix was to use a separate callback for it (and yes, I'm using test messages for custom application, both for debug and for command protocol via UART). However this approach did not cover the case when another channel uses another text message at the same time, and I was also getting message buffer corruption.

Generally speaking, both "lock and delegate long running ops" and "use the comm channel thread for long ops with reentrancy" approaches seem feasible.

However my opinion (and admittedly I haven't yet looked at CAN & NRF cases too deeply) is that the first approach is easier to get right, and would be a much smaller change from the current code state.

flytrex-vadim:
Also, it might make sense to treat "maintenance" operation such as CAN scan and detection differently from "running" operation. 

For example, in my use case, some disruption of other channels can be accepted for maintenance since it is usually done via one channel, with a person observing it. When "running" the system on the other hand, robustness is absolutely important. Even small chance of buffer corruption or not getting a reply makes it practically unusable.

Can't speak for other users of course ...

vedderb:
The CAN scan is actually quite common to do, especially when connecting over ble and checking what is going on, so it would be good if it can be done safely. Moving it to the detect thread and not allowing detect functions at the same time might make sense though.

Otherwise I don't really like the idea that one interface can do something that locks the communication for other interfaces. This is especially true for terminal callbacks, which should pose as little restrictions on the user code as possible. Then again, terminal callbacks can be handled by the detect thread as well.

Currently the detect thread will just ignore new commands if it is busy. It might make sense to rename it to long running, or maintenance, and make it such that it ignores new operations if they arrive while it is running.

Then for the other operations, it is probably best to make them reentrant where possible (which actually is true for most of them) and lock it for the rest.

I will work a bit on this tonight. Making it better than now should be quite easy.

flytrex-vadim:
> 
> 
> The CAN scan is actually quite common to do, especially when connecting over ble and checking what is going on, so it would be good if it can be done safely. Moving it to the detect thread and not allowing detect functions at the same time might make sense though.
> 
> Otherwise I don't really like the idea that one interface can do something that locks the communication for other interfaces. This is especially true for terminal callbacks, which should pose as little restrictions on the user code as possible. Then again, terminal callbacks can be handled by the detect thread as well.
> 
> Currently the detect thread will just ignore new commands if it is busy. It might make sense to rename it to long running, or maintenance, and make it such that it ignores new operations if they arrive while it is running.

Yes, completely agree.

> 
> Then for the other operations, it is probably best to make them reentrant where possible (which actually is true for most of them) and lock it for the rest.

In my experience the decision boiled down to: which invariant is easier to keep during code development and maintenance ?  Lock, and keep the all the code paths fast, or not lock, and keep the code reentrant.

Most of the time we chose to do lock and fast, because the  symptoms of violation were much more clear - detecting subtle data corruption is much harder than delays, and explaining the concept of "finish quickly" was **much** easier than explaining reentrancy.

In some case we even added timing code around the very similar mutex, and raised diagnostic warning if it was held for improper amount of time. I actually considered adding it here - storing largest command time and offending command ID. That way it becomes really obvious if the command dispatch code is "correct". Unlike in the "not lock and reenter" case, where maintaining correctness requires very careful review of every code change.

> I will work a bit on this tonight. Making it better than now should be quite easy.

:)

vedderb:
I agree that making it reentrant might not be worth the extra effort when the commands can run relatively fast anyway. In this case I think it is relatively easy though.

I gave it a go, and so far I'm quite happy with the updates. The changes are, among others, the following:

* Renamed the detect thread to blocking thread, moved "maintainance" commands there, and made it ignore other blocking commands if it is running already. Ignoring the other blocking commands makes sense anyway, as almost none of them should run at the same time ever.

* Used a stack buffer where possible, and added locks when using the global buffer. The locks do not stop commands that don't use the global buffer.

* The printf function now uses the last blocking send function, which should prevent the issue you described as the terminal is blocking as well.

* commands_process_packet not takes a send function as an argument (which can be null, for the can "hack" to work)

* Tested everything with the NRFs and USB.

Can you have a look, give it a try and let me know what you think?

flytrex-vadim:
Did some testing on today's master head version on USB+UART. The main thing that bothered me is the classification of the COMM_TERMINAL_CMD as a blocking command. This caused frequent drop of commands when other channel is heavily using them. 
And since my app is using terminal callback text message as primary communication method, that's  a big deal :)

Not sure what's an ideal solution, can see several options:

1. Lock around terminal_process_string (donwside - long running terminal message from one source can delay terminal messages from other sources)
2. Create additional COMM_FAST_TERMINAL_CMD that is not delegated to blocking_thread (downside - breaking protocol changes for 3rd party code)
3. Implement similar blocking thread mechanism in  terminal module, delegating only the long running commands (downside - additional complexity, inability to cleanly classify application_callback as fast or slow)

vedderb:
Ah, that is a bit of a problem :-)

Would it be possible to use the COMM_CUSTOM_APP_DATA command and provide a handler for it using this function?

```c
void commands_set_app_data_handler(void(*func)(unsigned char *data, unsigned int len));
```

Otherwise, is there any other command that would be useful for your app that can be added?

The reason I'm asking is that the terminal is intended for interactive debugging, and possibly blocking for a long time while doing that. It would be a but confusing and inconvenient to make is work for both fast and blocking commands.

flytrex-vadim:
It seems that using COMM_CUSTOM_APP_DATA would require implementing the command twice: once for interactive usage via terminal, and once for the fast and reliable mode. I think this is less desirable - I found the ability to have the same protocol for interactive terminal and inter-module communication a great benefit for development. The extensibility of the terminal protocol for application is major strength of the design, would be sad to loose it.

Would that be a big downside if the long-running terminal commands would delay other terminal commands (instead of getting them ignored as in current version) ? I think it could actually be preferred.

I did implement a couple of extra commands for my use - custom telemetry set (probably made redundant now by COMM_GET_VALUES_SELECTIVE :) and some [dynamic update of non-persistent current limits](https://github.com/Flytrex/bldc/blob/c774e0fa5a8667cf11191401cae524433350dc50/commands.c#L915). So I have some feel of how it is done, but still I think for some app communication needs terminal callback is better suited.

BTW, here's how my app's terminal function looks like:
https://github.com/Flytrex/bldc/blob/c774e0fa5a8667cf11191401cae524433350dc50/applications/app_brake.c#L161

vedderb:
I thought a bit more about it, and it might make sense to simply add another "fast" terminal command that is processed in commands_process_packet as you proposed. If you use it from custom applications, you have to be aware that it will stop the communication (on this channel) until it has finished. Would that work for your case?

flytrex-vadim:
Something along these lines ? Yes, should be adequate, I think. Maybe "synchronous terminal command" is better name than "fast".
```
	case COMM_SYNC_TERMINAL_CMD:
		chMtxLock(&terminal_mtx);
		data[len] = '\0';
		terminal_process_string((char*)data);
		chMtxLock(&terminal_mtx);
		break;
```

vedderb:
Yes, something like that. I didn't have the mutex in mind at first, but it does not hurt. Will make the push this evening, along with some other changes.

vedderb:
I'm done with the implementation now. Took a bit of extra time since I added IMU support as well, but hopefully it works as expected now.

Issue #33, opened 2017-05-12
[Feature Request] Add support for Sine-Cosine analog encoders


tlalexander:
Apologies if this is not the right place for this, but I would like to add support for Sine-Cosine encoders such as the [KMZ60](http://www.nxp.com/products/automotive-products/sensors/angular-sensors/angle-sensor-with-integrated-amplifier:KMZ60) to the VESC. (Files for my encoder are [here](https://github.com/tlalexander/MotorDriver/tree/master/Angle%20Sensor).)

From what I can gather from a quick examination, I should add a new encoder type to encoder.c, give it its own init function, and return a float angle on encoder_read_deg(). Seems easy enough.

Where do I patch this in to the configuration UI? I've built that too. I suppose that's a different project however...

Anyway, I didn't see a good place to discuss this so perhaps this can serve as a discussion board.

I will be happy testing on my own, but feel free to advise if anyone is interested.

gpzhaome:
Hi, 

Not sure I understand your problem correctly. But you can find the encoder configuration (reading from the UI) in conf_general.c . If you want to change the UI, you need to go to bldc-tool project (https://github.com/vedderb/bldc-tool) and change files accordingly.

tlalexander:
I've got the option added to the bldc-tool. That was easy.

![encoder](https://cloud.githubusercontent.com/assets/1310634/25987994/1d84a61a-36ab-11e7-8620-01e689f84163.png)

I will have to look at how to init the analog pins appropriately and write my encoder routine. I'll save that for another time. Bed time right now.

PR #81, opened 2019-04-02
Sin cos encoder support

nitrousnrg:
Here is theoretical support for sine/cosine analog encoders. This was not tested with actual sin/cos encoders attached to a motor, I'll post an update here when we actually get some field testing done.

Having said so, it should work and it was tested using signal generators connected to our boards.
These signals are input into what is commonly used as ACCEL and REGEN inputs, so using these encoders implicitly requires the board to be controlled over CANbus.

SCR03

Resulting in correct angle measurements

sincos phase acquisition

And rpm measurements

sincos rpm acquisition

Hopefully this can be merged now to avoid future merge conflicts, it is only available by asserting:
`#define HW_HAS_SIN_COS_ENCODER` 

When sin/cos signals are not reasonable a fault is asserted. Offset, gain compensation and low pass filtering constant is exposed to be programmed by the user with VESC_TOOL. I have a patch for VESC TOOL to activate sin/cos encoder, but if you prefer we can leave it out for now until we do some formal testing.

Note that encoder.c now includes "mc_interface.h" only to be able to read the analog signals. You may have a way to remove the dependency.

I don't see any reason why `mcpwm_foc_encoder_detect()` wouldn't work, but it needs to be tested.

During testing we will add to `mcpwm_foc_encoder_detect()` the offset and gain compensation to the analog signals so it automatically detects and configures the signals offsets and gains, which are mandatory for correct operation. If sin or cos mean values are too different to the offest would be another interesting fault mode to monitor.


vedderb:
Nice! I will try to merge it in the next days. Was away all last week and really busy in the weekend.

One comment:

ENCODER_SIN_VOLTS
and
ENCODER_COS_VOLTS

could be defined ifndef-wrapped in hw.h, and default to 0 or ADC_VOLTS(ADC_IND_EXT) if ADC_IND_EXT are defined.

It would be good if you can add support to VESC Tool too. confgenerator is generated from VESC Tool, so you only need to use the mcconf editor to generate an XML. VESC Tool will also generate a signature that is added to confgenerator to avoid storing or loading incompatible configurations by accident. Note that you need to restart VESC Tool with the new XML before confgenerator can be generated. I hope this makes sense :-)

nitrousnrg:
Good, let me try your comments.

Also I just found it there is a reported issue about this (#33) When its merged you can automatically close the issue: https://help.github.com/en/articles/closing-issues-using-keywords

nitrousnrg:
I generated the XML and generated confgenerator.c/h from VESC Tool with a new signature. I think I should have bumped vesc tool version. Will create a Pull request on VESC Tool anyways.

image

Issue #75, opened 2019-02-27
Signal injection


nitrousnrg:
Injecting signals into the control loop will help developing new algorithms and improve the coverage and test automation of the code on every commit.

* Being able to simulate a motor within vesc firmware allows developers and researchers to start from a known motor model and evaluate new algorithms by injecting current and voltages into the control loop.
This is an example of an IPM motor model that could be used to approach issue #26
http://cpp.sh/33vcz

* Light VESC users could easily simulate a motor by launching injection from the command line to learn to operate VESC_tool

* Most importantly, signal injection makes it possible to perform Processor-in-the-Loop testing, executing the code in the actual target and exercising all critical functions for a broader and automated code coverage test, including fault cases as well as normal operating conditions.

rpasichnyk:
Being able to simulate a motor would be a dream 😍😍😍

Peemouse:
Definitely a great idea !!!

nitrousnrg:
We have some basic signal injection working based on that reference code. This issue is a placeholder to reference a future merge taking care of this.

PR #82, opened 2019-04-06
Add command to connect a virtual motor with configurable parameters

nitrousnrg:
This commit adds a motor model running within the vesc firmware,and
from the vesc terminal a user or a test script can set the
mechanical load torque, inertia, phase resistance, Lq and Ld phase
inductances (this generic model includes IPM motors), flux linkage
and battery voltage.

Virtual motor parameters set at the command line should match with
vesc configuration, for example phase resistance, inductance and
flux linkage should match and have the correct observer gain.
Observer works with the virtual motor, with some hiccups during
startup
For solid results its better to use sensored mode. If vesc is
configured to use an SPI encoder the virtual motor phase angle
will be injected as an encoder angle readout.

For safety PWM outputs are disabled during simulation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,6 +154,7 @@ CSRC = $(STARTUPSRC) \
        confgenerator.c \
        timer.c \
        i2c_bb.c \
+       virtual_motor.c \
        $(HWSRC) \
        $(APPSRC) \
        $(NRFSRC) \
--- a/conf_general.h
+++ b/conf_general.h
@@ -188,6 +188,8 @@
 // Current ADC to amperes factor
 #define FAC_CURRENT					((V_REG / 4095.0) / (CURRENT_SHUNT_RES * CURRENT_AMP_GAIN))
 
+#define VOLTAGE_TO_ADC_FACTOR	( VIN_R2 / (VIN_R2 + VIN_R1) ) * ( 4096.0 / V_REG )
+
 // Actual voltage on 3.3V net based on internal reference
 //#define V_REG						(1.21 / ((float)ADC_Value[ADC_IND_VREFINT] / 4095.0))
 //#define V_REG						3.3
--- a/hwconf/hw_palta.h
+++ b/hwconf/hw_palta.h
@@ -140,6 +140,9 @@
 #define GET_GATE_DRIVER_SUPPLY_VOLTAGE()	((float)ADC_VOLTS(ADC_IND_VOUT_GATE_DRV) * 11.0)
 #endif
 
+#define ANGLE_TO_DAC_VALUE(angle)	( angle * 512.0 + 0x800 )//angle between -pi to pi
+#define CURRENT_TO_DAC_VALUE(current)	( current * 70.0  + 0x800 )//current
+#define VOLTAGE_TO_DAC_VALUE(voltage)	( voltage * 40.0 + 0x800 )//angle between -pi to pi
 
 // Double samples in beginning and end for positive current measurement.
 // Useful when the shunt sense traces have noise that causes offset.
--- a/mcpwm_foc.c
+++ b/mcpwm_foc.c
@@ -37,6 +37,7 @@
 #include <math.h>
 #include <string.h>
 #include <stdlib.h>
+#include "virtual_motor.h"
 
 // Private types
 typedef struct {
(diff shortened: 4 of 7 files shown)


nitrousnrg:
Commit message should be self explanatory.

One thing I forgot to add is that encoder offset should be set to 0 and encoder ratio to 1.0. In any case I'm attaching a motor configuration xml that we used for testing.
[virtual-motor-config.zip](https://github.com/vedderb/bldc/files/3050533/virtual-motor-config.zip)

Virtual motor is engaged with this terminal command
`connect_virtual_motor [ml][J][Ld][Lq][Rs][lambda][Vbus]`

Where:
ml: mechanical load torque
J: load inertia
Ld: inductance d axis
Lq: inductance q axis
Rs: phase resistance
Lambda: flux linkage
Vbus: battery voltage

For example:
`connect_virtual_motor 0.025 0.001 0.000020 0.000020 0.020 0.020 48.0`

Issue `disconnect_virtual_motor` to disconnect the motor 

Lq and Ld are equal in the SPM motors currently supported by vesc (especially by vesc observer)

Here are some screenshots of the virtual motor running

BEMF
virtual_motor_BEMF

Motor startup injecting encoder signal
virtual_motor_start current 1
virtual_motor_start current 2

Sensorless start
virtual_motor_sensorless start current 1
virtual_motor_sensorless start current 2

Realtime RPM plot
virtual_motor_start

Virtual motor was tested on bare discovery F4 boards and powerdesigns RevD boards.

This is a direct reference to Issue #75 (Signal Injection).

nitrousnrg:
The motor model used here is the same suggested in issue #75 , originally suggested by technologic in endless-sphere forums. The equations model botn SPM and IPM machines.

http://cpp.sh/33vcz

```
	/**
	 * Run electrical model of the machine
	 * @param vd	d- Axis Stator Voltage in V
	 * @param vq	q- Axis Stator Voltage in V
	 */
	inline void run_electro(float vd, float vq)
	{
		// d axis current
		id += (vd + w*Lq*iq - Rs*id)*Ts/Ld;
		// q axis current
		iq += (vq - w*Ld*id - Rs*iq - w*Psi)*Ts/Lq;
	}

	/**
	 * Run mechanical side of the machine
	 * @param ml	externally applied load torque in Nm
	 */
	inline void run_mechanics(float ml)
	{
		// precalculated constants to optimise compiler optimisation
		const float _3by2 = 3.0F/2.0F;
		const float	tsj = Ts/J;

		// electric torque
		me = _3by2 * (Psi * iq + (Lq - Ld) * id * iq);

		// omega
		w += tsj * (me - ml);

		// phi
		phi += w * Ts;
		// for longer runs consider using Double or
		// limit phi with modf function to 2*PI or something else
		// if not this is numerically not stable
	}
```

vedderb:
Looks awesome, and really nice that it plugged in to the rest of the code so cleanly! I will definitely give it a try when I get home.

gowrav:
Linking the issue that has come up with Virtual Motor.,
https://github.com/vedderb/bldc/pull/88#issuecomment-515679255
One more thing., the parameters taken from the real motor has been fed into the virtual motor but the ERPM varies widely.
Since the ERPM is primarily based on Voltage and Flux Linkage what am I missing.
motorparam
The real motor saturates at 7000ERPM but the virtual motor with the below parameter just shoots through 13k ERPM at 18%duty cycle.
**connect_virtual_motor 1.0 0.1 0.000070 0.000070 0.024 0.040 48**

Screenshot from 2019-07-27 17-54-57

nitrousnrg:
Hi @gowrav 
please upload your motor config file xml so I can take a look.
The primary purpose of the simulator is to develop new algorithms, it may not work with arbitrary configurations and we usually use the .zip linked in this thread. I don't recall coding any hall sensor emulator for example.

dothebart:
Hi, @nitrousnrg , this is a very interesting feature. Does one still need the companion switching ICs in a VESC? Or will this also work with a stock Nucleo evaluation board?

nitrousnrg:
Yes it works as long as the hw version doesnt listen to external fault inputs, this branch was developed with a bare discoveryF4 board.

dothebart:
hm, so it maybe could even work with https://beckus.github.io/qemu_stm32/ ?

dothebart:
hey, do you have any hints, how the bldc tool will talk to the board? did you add a usb connector to PA11/PA2? I'm struggling to get it working - any hints would be appreciated.

PR #85, opened 2019-04-10
Add new control type to ADC application

tipsmiller:
Adds a new control configuration for the ADC application. See a corresponding pull request on vesc_tool for the counterpart.


tipsmiller:
See [pull request](https://github.com/vedderb/vesc_tool/pull/31t)

PR #87, opened 2019-04-16
Flash memory CRC integrity check

nitrousnrg:
Hardened flash memory.

This code checks the integrity of the application image on every boot, and keeps the flash locked to prevent unwanted writes. I saw memory corruptions when the hardware was faulting with 1000+Amps flowing a few centimeters away from the control board and it would need to be reflashed to boot again. In some cases the code would run, but the motor would run very bad, and after reflashing it worked perfectly again, which is dangerous, so now it performs the check and does not run corrupted code.
Also some industries require a CRC-based memory integrity check like this.

I thought about adding a crc check to the mc_conf structure but I wasn't sure if vesc_tool code needs to be modified as well.

I avoided computing the CRC at build time because I figured its troublesome to make it build in linux and windows.

This also fixes the linker file which was allowing to build images larger than what the bootloader is capable of flashing,

See commits comments for more detailed info.


vedderb:
This is a good addition, I think I have seen behavior on some VESCs that indicate corrupted flash memory. I have a few suggestions:

1)
I would prefer to avoid using the ST libraries if possible, and there is really no need for hw crc acceleration as it can be calculated quite fast in software. If you need a 32-bit crc you can do something like this:

```
uint32_t crc32c(uint8_t *data, uint32_t len) {
	uint32_t crc = 0xFFFFFFFF;

	for (uint32_t i = 0; i < len;i++) {
		uint32_t byte = data[i];
		crc = crc ^ byte;

		for (int j = 7;j >= 0;j--) {
			uint32_t mask = -(crc & 1);
			crc = (crc >> 1) ^ (0x82F63B78 & mask);
		}
	}

	return ~crc;
}
```

2)
Instead of putting the checksum and the set flag in the app flash memory, you can put it in the first 16kb sector where the vector table is. The rest of that sector is empty anyway.

3)
Instead of relying on someone putting the CRC in flash after compiling, the firmware could write it on the first boot if it is missing based on the flag. As the sector with the vector table only uses the first 500 or so bytes, the app can simply store the table to RAM, erase the first 16kb sector and then write the table + the calculated crc + the flag back.

4)
Another nice addition would be to check the checksum periodically during runtime using a thread, and reset the CPU if it fails. The thread can run at a low rate and split the calculation into chunks of 1kb or so with a sleep in between to not block the other threads and introduce the least possible overhead.

Do you think you can look into these points? I can also have a look the next time I have an opportunity, but currently I'm quite busy making some new hardware versions and remotes.

nitrousnrg:
Thanks for the comments Benjamin,
I'll benchmark hw accel vs software, and move the crc to the first sector. That way I could check the full flash memory instead of only checking the app. Currently the firmware writes the crc in the first boot, but its unable to update that crc because its a large sector to erase so it only happens after a firmware upgrade.

Do you prefer full flash checking or only check the app image?

I really like the continous integrity check idea, will give it a try as well.

ledvinap:
From my experience it's quite easy to compute CRC using srecord tool (http://srecord.sourceforge.net/; both linux and windows supported) and store it in image. It is also possible to include image length, version etc. Just use linker script to allocate info structure at known location (end of normal flash image / end of first page), fill it with magic values (possibly magic value to inform bootloader that CRC shall be ignored for debudding) and use srecord to fill in correct values.

Modifying CRC on first boot can lead to interesting edge cases (bootloop with flash erase, eventually destroying CPU, etc.). It should be considered if build convenience is worth the risk. 
Also including CRC  in image will protect against damaged image file / incomplete flash procedure etc.

Continuous image check is probably useful for motor control. It may detect even intermittent flash read errors (caused by EMI). 
Running out of HW breakpoints may force debugger to overwrite parts of program memory, it should be noted in documentation ... 

@nitrousnrg : CRC speed should not be that important on boot, it may me more noticeable when using background task. 
CRC unit is about 60 times faster on F3 (https://www.st.com/content/ccc/resource/technical/document/application_note/39/89/da/89/9e/d7/49/b1/DM00068118.pdf/files/DM00068118.pdf/jcr:content/translations/en.DM00068118.pdf)
It is possible to use DMA for CRC calculation, further reducing CPU loading (0.72% is quoted in document above, but it seems too good to be true - D-code bus cycles will stall I-code and D-code cache will be trashed).

nitrousnrg:
After looking into this a bit closer, I don't want to use flash sector 0 to store the crc. The only reason for doing so is to be able to change the stored CRC, but the app checksum will never change. It is also more straightforward if the CRC is stored at the end of the block.

If one day we figure out how to build the firmware with crc included, its just a matter of including the flag+crc and the code will never update the crc.

If I used sector 0 to be able to update the crc and include the whole flash image, I would have to rewrite the same flash bits over an over because its not part of the EEPROM emulator.

The mc_conf CRC (not implemented yet) should be stored at the end of structure anyways, because it will make use of the EEPROM emulator for proper wear leveling. If that chunk is corrupted, it is still safe to run the application in limp mode, at least to fix it with vesc tool and avoid having a bricked board.

About st libraries, I will remove them but still use the CRC acceleration for the continuous checking, we don't have that much cpu power to spare when you need high switching frequencies.

vedderb:
I didn't realize that you already store the CRC if it was not found, that is good. The problem is that if the program is smaller than two sectors (less than 256 KB), the last sector will not be erased by openocd and the old CRC will still be there. Then the flag will be set since the last programming, and the check will fail. The new CRC also cannot be written as bits only can be cleared by the write function and a page erase is required to set them.

One way to get around this is to move the whole program 8 bytes forward and put the CRC and flag in the beginning. Then those bytes will always be erased when the program is written. Another way is to put it in the end of sector 0 as the vector table is always written, and thus that page is cleared. I would prefer the second approach as moving the whole program would make this incompatible with the bootloader, and a lot of them are out there. The second approach requires the argument --gap-fill 0xFF to objcopy for generating the bin file, as it otherwise would clear the whole sector and make it impossible to write the CRC and flag.

Regarding boot loops that destroy the flash, I don't think that is a problem if the flag counts if it is !0xFF. An exact comparison might cause a problem though if that flag somehow gets corrupted.

Regarding continuous checking in a thread that sleeps most of the time (not checking all memory at once), that will have minimal impact on performance. The worst thing that can happen is that the check takes long to run, which still is quite ok. Using the HW does not hurt though.

nitrousnrg:
I have a simple code that will make sure the last bytes of the image are always programmed as 0xFF. Thats how I found the linker script could build images larger than it should, I didn't include it in the commit because I didn't see that failure mode, but I will now, thanks for pointing that out.

I have a low priority thread continuously checking flash integrity in the background with the accelerated CRC, crunching 2kB at a time. It has been running non stop since yesterday without problems. After a bit of profiling with the scope I'll push the commits.

nitrousnrg:
Continuous integrity check code was inspired on an app note from ST:
periodic_crc


And here is how the continuous CRC check behaves

With the vesc at low load (20khz fsw) calculating an 8kB CRC chunk requires about 200us.
SCR03

With fsw set at 30kHz, the integrity check task is interrupted more often and it takes ~1ms to compute each chunk.
SCR02

Overall, at 30kHz this is how long it takes it to calculate the CRC
SCR04
Its about 2% duty (includes time spent in higher priority tasks and ISR)

'threads' command in VESC Terminal reports <1% of the time spent in this new thread.

vedderb:
Very good!

I would recommend using a smaller chunk size and running the thread more often, to not block other threads that potentially want to run at a high rate (without forcing preemtive rescheduling). e.g. 1024 and 5 ms. Also, 256 byte stack is probably more than enough for that thread.

crc.c
```diff
@@ -58,3 +59,29 @@ unsigned short crc16(unsigned char *buf, unsigned int len) {
...
+}
+
+/**
+  * @brief  Resets the CRC Data register (DR).
+  * @param  None
+  * @retval None
+  */
+void crc32_reset(void) {
+  /* Reset CRC generator */
+  CRC->CR = CRC_CR_RESET;
```
ledvinap: (rest of CRC_CR is reserved, this should be `CRC->CR |= CRC_CR_RESET;`

flash_helper.c
```diff
@@ -191,3 +216,98 @@ uint8_t* flash_helper_get_sector_address(uint32_t fsector) {
...
+
+/**
+  * @brief  Compute the CRC of the application code to verify its integrity
+  * @retval FAULT_CODE_NONE or FAULT_CODE_FLASH_CORRUPTION
+  */
+uint32_t flash_helper_verify_flash_memory(void) {
+	uint32_t crc;
+	// Look for a flag indicating that the CRC was previously computed.
+	// If it is blank (0xFFFFFFFF), calculate and store the CRC.
+	if( (APP_CRC_WAS_CALCULATED_FLAG_ADDRESS)[0] == APP_CRC_WAS_CALCULATED_FLAG )
```
ledvinap: `crc_info.crc_flag` is already in correct place (but not exported now)

crc.c
nitrousnrg: I copied the function from ST library v1.6.0. It doest hurt if I make the change, thanks

nitrousnrg:
Thanks for merging Benjamin.
I resolved @ledvinap comments because I decided to implement his suggestions and yours as well, they made sense but I didn't get to the pc before the merge.
Heads up, runtime crc check fails with 1kb chunks, trying to figure out why.

nitrousnrg:
Benjamin, this was the final update left out of the PR
https://github.com/paltatech/bldc/commit/215f9697cdd166ceffafff4894f9cd1a7922391d

Fixes a pointer cast and allows it to work with 1kb chunks. I can either create a new PR or you make the changes, its the same to me.

vedderb:
I found them by myself when testing and fixed them slightly differently, should give the same result. I also found a few other issues that I fixed:

* The bootloader would not work as it needs 8 extra bytes for the program length and checksum, so I changed the linker script and sizes.

* APP_CRC_WAS_CALCULATED_FLAG_ADDRESS and where you wrote the CRC was missing the 0x08000000 flash offset. For some reason it worked anyway, but it was probably a coincidence by some bit being masked out. That is fixed now.

* I set APP_CRC_WAS_CALCULATED_FLAG to 0 so that writing it always is successful to avoid destroying the flash in a boot loop if it somehow gets corrupted.

For me it seems to work well now. Can you give it a try?

nitrousnrg:
Awesome, thanks for the feedback. Its working well in our hardware.

It would be interesting to add this thread to the list of watchdog-watched threads. If things go south this low priority thread is the first one that will miss deadlines.

pavel-kirienko:
I seem to be late here, but have you looked at the PX4 Brickproof Bootloader and [Kocherga](https://github.com/Zubax/kocherga)? Kocherga is basically a slightly revised reimplementation of the former optimized for reuse. Kocherga defines a descriptor structure which can be located anywhere in the ROM, its location is marked with a special 8-byte magic sequence. After the firmware binary is generated, it is post-processed with a script which computes its CRC64 and size and populates the respective fields in the descriptor structure. Any dependency on device-specific hardware properties such as flash sector configuration is thus avoided. The script runs on any major general-purpose OS (GNU/Linux, Windows, OSX), so no portability issues should be expected, although it requires Python.

PR #88, opened 2019-04-20
Detect current sensor failures

nitrousnrg:
2 failure modes added:
* On boot, when calculating DC offsets generate a fault if the offset is beyond
HW_MAX_CURRENT_OFFSET. Fault code reports which sensor is having issues (1, 2 or 3).
Most likely cause is a disconnected sensor, if hw has fault-detecting pullups.

* On runtime, in setups with 3 in-line phase current sensors, checks that the sum
of the currents is below MCCONF_MAX_CURRENT_UNBALANCE, with a configurable low pass
filter. If unbalanced current is high, it means a fault to ground, or a disconnected
sensor (this works at 0 Amp if hw has pullups in the sensor input to detect the
failure).

Here's a quick video of this in action. This commit is really important because loose sensors and
wiring damage are a problem in builds like these.
https://youtu.be/nkkPLbIC2JI


gowrav:
Virtual Motor > **connect_virtual_motor 1 0.1 0.000070 0.000070 0.024 0.040 48**
when run produces **CURRENT_UNBALANCE** error., can only be reset by [rebooting].
Screenshot from 2019-07-27 17-27-03
Screenshot from 2019-07-27 17-31-03
Screenshot from 2019-07-27 17-33-09
Screenshot from 2019-07-27 17-33-35
Screenshot from 2019-07-27 17-34-06
Screenshot from 2019-07-27 17-34-15






what could be the issue.
HW: **VESC6+ [60]**
FW: **3.58**

Issue #76, opened 2019-02-27
Field Weakening


nitrousnrg:
We need to provide firmware support for Field Weakening, which allows a motor to spin beyond its base speed at full power.

![image](https://user-images.githubusercontent.com/309472/53501704-67ed9780-3a8b-11e9-95fa-08e88677b662.png)

An implementation example:
![image](https://user-images.githubusercontent.com/309472/53502219-653f7200-3a8c-11e9-82f3-99337f82e9b8.png)

Deep FW operation is not efficient in small BLDC motors, but for IPM motors (issue #73) it is mandatory to go beyond base speed for full utilization of the machine, even if power drops a bit due to Id current losses.

Note this feature is ***not safe*** for most hardware implementations because it requires the powerstage to be able to withstand high motor BEMF voltages and some special fault management has to take place.

jokucale:
Hello does this field  weaking feature work with ebike hub motors

PR #91, opened 2019-05-03
Flux Weakening and MTPA support

### LINKED ISSUE #73: IPM motor support
Most automotive applications use Interior Permanent Magnet motors.

These motors have a permanent-magnet inside the rotor that generates magnetic saliency. This saliency
produces a reluctance torque from the difference between the direct-d axis inductance and the quadrature q-axis inductance. The maximum efficiency of the IPM motors can be achieved by proper selection of the current vector ratio between magnetic torque current and reluctance torque current in the total current.
Hence, a basic FOC control on an IPM machine could be losing as much as 30% of the available torque.

This control scheme is known as Maximum Torque Per Amp (MTPA)

image

And here is some implementation detail with control examples.
http://www.ti.com/lit/an/spracf3/spracf3.pdf

This is a placeholder so we don't forget to implement this feature. Benjamin, feel free to assign this issue to me or to powerdesigns organization.

nitrousnrg:
This is a major feature that will demand some careful review, its tested and crafted to be mergable.

The control scheme has been adopted from [here](http://www.ti.com/lit/an/spracf3/spracf3.pdf)

Here is a video of FW working, injecting a negative Id current to produce torque beyond base speed.

https://youtu.be/02jA0u5V-X0

In speed control mode there are some integrator issues, it takes too long to "forget" the error signal and an anti-windup didn't work very well.

Before trying this you may want to merge #90 as it comes with a fix for proper simulation of this. The simulation was super useful and allowed to test the new control. The real motor behaved just like the simulation.

This also simplifies the virtual motor command, as now it fetches the motor parameters from the stored motor configuration. It wasn't possible before because vesc wasn't aware of Lq and Ld inductances.

To enable this feature, VESC Tool requires this patch:
https://github.com/vedderb/vesc_tool/pull/34
to enable the features
image

If approved this will close #73 and #76 

Have fun!

### DIFF
--- a/confgenerator.c
+++ b/confgenerator.c
@@ -71,6 +71,8 @@ int32_t confgenerator_serialize_mcconf(uint8_t *buffer, const mc_configuration *
 	buffer_append_float32_auto(buffer, conf->foc_pll_kp, &ind);
 	buffer_append_float32_auto(buffer, conf->foc_pll_ki, &ind);
 	buffer_append_float32_auto(buffer, conf->foc_motor_l, &ind);
+	buffer_append_float32_auto(buffer, conf->foc_motor_ld, &ind);
+	buffer_append_float32_auto(buffer, conf->foc_motor_lq, &ind);
 	buffer_append_float32_auto(buffer, conf->foc_motor_r, &ind);
 	buffer_append_float32_auto(buffer, conf->foc_motor_flux_linkage, &ind);
 	buffer_append_float32_auto(buffer, conf->foc_observer_gain, &ind);
@@ -97,6 +99,10 @@ int32_t confgenerator_serialize_mcconf(uint8_t *buffer, const mc_configuration *
 	buffer[ind++] = conf->foc_temp_comp;
 	buffer_append_float32_auto(buffer, conf->foc_temp_comp_base_temp, &ind);
 	buffer_append_float32_auto(buffer, conf->foc_current_filter_const, &ind);
+	buffer[ind++] = conf->foc_mtpa_enable;
+	buffer[ind++] = conf->foc_field_weakening_enable;
+	buffer_append_float32_auto(buffer, conf->foc_field_weakening_kp, &ind);
+	buffer_append_float32_auto(buffer, conf->foc_field_weakening_ki, &ind);
 	buffer_append_int16(buffer, conf->gpd_buffer_notify_left, &ind);
 	buffer_append_int16(buffer, conf->gpd_buffer_interpol, &ind);
 	buffer_append_float32_auto(buffer, conf->gpd_current_filter_const, &ind);
@@ -290,6 +296,8 @@ bool confgenerator_deserialize_mcconf(const uint8_t *buffer, mc_configuration *c
 	conf->foc_pll_kp = buffer_get_float32_auto(buffer, &ind);
 	conf->foc_pll_ki = buffer_get_float32_auto(buffer, &ind);
 	conf->foc_motor_l = buffer_get_float32_auto(buffer, &ind);
+	conf->foc_motor_ld = buffer_get_float32_auto(buffer, &ind);
+	conf->foc_motor_lq = buffer_get_float32_auto(buffer, &ind);
 	conf->foc_motor_r = buffer_get_float32_auto(buffer, &ind);
 	conf->foc_motor_flux_linkage = buffer_get_float32_auto(buffer, &ind);
 	conf->foc_observer_gain = buffer_get_float32_auto(buffer, &ind);
@@ -316,6 +324,10 @@ bool confgenerator_deserialize_mcconf(const uint8_t *buffer, mc_configuration *c
 	conf->foc_temp_comp = buffer[ind++];
 	conf->foc_temp_comp_base_temp = buffer_get_float32_auto(buffer, &ind);
 	conf->foc_current_filter_const = buffer_get_float32_auto(buffer, &ind);
+	conf->foc_mtpa_enable = buffer[ind++];
+	conf->foc_field_weakening_enable = buffer[ind++];
+	conf->foc_field_weakening_kp = buffer_get_float32_auto(buffer, &ind);
+	conf->foc_field_weakening_ki = buffer_get_float32_auto(buffer, &ind);
 	conf->gpd_buffer_notify_left = buffer_get_int16(buffer, &ind);
 	conf->gpd_buffer_interpol = buffer_get_int16(buffer, &ind);
 	conf->gpd_current_filter_const = buffer_get_float32_auto(buffer, &ind);
@@ -505,6 +517,8 @@ void confgenerator_set_defaults_mcconf(mc_configuration *conf) {
 	conf->foc_pll_kp = MCCONF_FOC_PLL_KP;
 	conf->foc_pll_ki = MCCONF_FOC_PLL_KI;
 	conf->foc_motor_l = MCCONF_FOC_MOTOR_L;
+	conf->foc_motor_ld = MCCONF_FOC_MOTOR_LD;
+	conf->foc_motor_lq = MCCONF_FOC_MOTOR_LQ;
 	conf->foc_motor_r = MCCONF_FOC_MOTOR_R;
 	conf->foc_motor_flux_linkage = MCCONF_FOC_MOTOR_FLUX_LINKAGE;
 	conf->foc_observer_gain = MCCONF_FOC_OBSERVER_GAIN;
@@ -531,6 +545,10 @@ void confgenerator_set_defaults_mcconf(mc_configuration *conf) {
 	conf->foc_temp_comp = MCCONF_FOC_TEMP_COMP;
 	conf->foc_temp_comp_base_temp = MCCONF_FOC_TEMP_COMP_BASE_TEMP;
 	conf->foc_current_filter_const = MCCONF_FOC_CURRENT_FILTER_CONST;
+	conf->foc_mtpa_enable = MCCONF_FOC_MTPA_ENABLE;
+	conf->foc_field_weakening_enable = MCCONF_FOC_FIELD_WEAKENING_ENABLE;
+	conf->foc_field_weakening_kp = MCCONF_FOC_FIELD_WEAKENING_KP;
+	conf->foc_field_weakening_ki = MCCONF_FOC_FIELD_WEAKENING_KI;
 	conf->gpd_buffer_notify_left = MCCONF_GPD_BUFFER_NOTIFY_LEFT;
 	conf->gpd_buffer_interpol = MCCONF_GPD_BUFFER_INTERPOL;
 	conf->gpd_current_filter_const = MCCONF_GPD_CURRENT_FILTER_CONST;
--- a/confgenerator.h
+++ b/confgenerator.h
@@ -8,7 +8,7 @@
 #include <stdbool.h>
 
 // Constants
-#define MCCONF_SIGNATURE		503309878
+#define MCCONF_SIGNATURE		1493606584
 #define APPCONF_SIGNATURE		725102789
 
 // Functions
--- a/datatypes.h
+++ b/datatypes.h
@@ -238,6 +238,8 @@ typedef struct {
 	float foc_encoder_cos_gain;
 	float foc_encoder_sincos_filter_constant;
 	float foc_motor_l;
+	float foc_motor_ld;
+	float foc_motor_lq;
 	float foc_motor_r;
 	float foc_motor_flux_linkage;
 	float foc_observer_gain;
@@ -260,6 +262,11 @@ typedef struct {
 	bool foc_temp_comp;
 	float foc_temp_comp_base_temp;
 	float foc_current_filter_const;
+	//MTPA and FW config
+	bool foc_mtpa_enable;
+	bool foc_field_weakening_enable;
+	float foc_field_weakening_kp;
+	float foc_field_weakening_ki;
 	// GPDrive
 	int gpd_buffer_notify_left;
 	int gpd_buffer_interpol;
--- a/mcconf/mcconf_default.h
+++ b/mcconf/mcconf_default.h
@@ -251,6 +251,12 @@
 #ifndef MCCONF_FOC_MOTOR_L
 #define MCCONF_FOC_MOTOR_L				0.000007
 #endif
+#ifndef MCCONF_FOC_MOTOR_LD
+#define MCCONF_FOC_MOTOR_LD				0.00002
+#endif
+#ifndef MCCONF_FOC_MOTOR_LQ
+#define MCCONF_FOC_MOTOR_LQ				0.00002
+#endif
 #ifndef MCCONF_FOC_MOTOR_R
 #define MCCONF_FOC_MOTOR_R				0.015
 #endif
@@ -328,6 +334,18 @@
 #endif
 #ifndef MCCONF_FOC_CURRENT_FILTER_CONST
 #define MCCONF_FOC_CURRENT_FILTER_CONST	0.1		// Filter constant for the filtered currents
+#endif
+#ifndef MCCONF_FOC_MTPA_ENABLE
+#define MCCONF_FOC_MTPA_ENABLE				false	// MTPA Enable
+#endif
+#ifndef MCCONF_FOC_FIELD_WEAKENING_ENABLE
+#define MCCONF_FOC_FIELD_WEAKENING_ENABLE	false	// Field Weakening enable
+#endif
+#ifndef MCCONF_FOC_FIELD_WEAKENING_KP
+#define MCCONF_FOC_FIELD_WEAKENING_KP		0.02	// Field Weakening proportional gain
+#endif
+#ifndef MCCONF_FOC_FIELD_WEAKENING_KI
+#define MCCONF_FOC_FIELD_WEAKENING_KI		0.02	// Field Weakening integral gain
 #endif
 
 // GPD
--- a/mcpwm_foc.c
+++ b/mcpwm_foc.c
@@ -76,17 +76,15 @@ typedef struct {
 } mc_sample_t;
 
 typedef struct {
-	bool enable;
 	float k1;// k1 = 8*SQ(Ld-Lq)
 	float k2;// k2 = 1/(4*(Ld-Lq))
 } mtpa_t;
 
 typedef struct {
-	bool enable;	//enable fw mechanism
-	float vs_int;	//integral voltage sum of PI
-	float kp;		//PI proportional gain
-	float ki;		//PI integral gain
-}flux_weakening_t;
+	float mod_int;	//integral sum of PI
+	float is_sq_max;	//max is current
+	float anti_windup_error;	//last iteration windup error
+}field_weakening_t;
 
 // Private variables
 static volatile mc_configuration *m_conf;
@@ -94,7 +92,7 @@ static volatile mc_state m_state;
 static volatile mc_control_mode m_control_mode;
 static volatile motor_state_t m_motor_state;
 static volatile mtpa_t mtpa;
-static volatile flux_weakening_t flux_weakening;
+static volatile field_weakening_t field_weakening;
 static volatile int m_curr0_sum;
 static volatile int m_curr1_sum;
 static volatile int m_curr_samples;
@@ -150,9 +148,10 @@ static void start_pwm_hw(void);
 static int read_hall(void);
 static float correct_encoder(float obs_angle, float enc_angle, float speed);
 static float correct_hall(float angle, float speed, float dt);
-static void mtpa_run( float Is,float *id, float *iq );
-static void flux_weakening_run(float vs_target, float dt, float *id_fw);
-static void flux_weakening_setup(void);
+static void mtpa_setup(void);
+static void mtpa_run(float *id, float *iq);
+static void field_weakening_setup(void);
+static void field_weakening_run(float dt, float *id, float *iq);
 
 // Threads
 static THD_WORKING_AREA(timer_thread_wa, 2048);
@@ -255,8 +254,9 @@ void mcpwm_foc_init(volatile mc_configuration *configuration) {
 	memset((void*)&m_motor_state, 0, sizeof(motor_state_t));
 	memset((void*)&m_samples, 0, sizeof(mc_sample_t));
 
-	flux_weakening_setup();
-	virtual_motor_init();
+	mtpa_setup();
+	field_weakening_setup();
+	virtual_motor_init(configuration);
 
 #ifdef HW_HAS_3_SHUNTS
 	m_curr2_sum = 0;
@@ -502,6 +502,11 @@ void mcpwm_foc_set_configuration(volatile mc_configuration *configuration) {
 	stop_pwm_hw();
 	uint32_t top = SYSTEM_CORE_CLOCK / (int)m_conf->foc_f_sw;
 	TIMER_UPDATE_SAMP_TOP(MCPWM_FOC_CURRENT_SAMP_OFFSET, top);
+
+	//if any change has been made to the configuration, we will need to update
+	//mtpa and flux weakening constants.
+	mtpa_setup();
+	field_weakening_setup();
 }
 
 mc_state mcpwm_foc_get_state(void) {
@@ -1895,6 +1900,7 @@ void mcpwm_foc_adc_int_handler(void *p, uint32_t flags) {
 		const float duty_abs = fabsf(m_motor_state.duty_now);
 		float id_set_tmp = m_id_set;
 		float iq_set_tmp = m_iq_set;
+
 		m_motor_state.max_duty = m_conf->l_max_duty;
 
 		static float duty_filtered = 0.0;
@@ -2043,29 +2049,10 @@ void mcpwm_foc_adc_int_handler(void *p, uint32_t flags) {
 		}
 
 		// Apply MTPA
-		float id_mtpa = id_set_tmp;
-		float iq_mtpa = iq_set_tmp;
-		float iq_set_by_speed = iq_set_tmp;
+		mtpa_run(&id_set_tmp, &iq_set_tmp);
 
-		mtpa_run(iq_set_by_speed, &id_mtpa, &iq_mtpa);
-
-		if( flux_weakening.enable ){
-			// Apply Flux Weakening
-			float id_fw;
-			flux_weakening_run( 26.0, dt, &id_fw);
-			id_set_tmp = id_fw + id_mtpa;
-
-			//calculate  iq
-			float Idiff = SQ(iq_set_by_speed) - SQ(id_set_tmp);
-			if( Idiff < 0.0 ){
-				Idiff = 0.0;
-			}
-
-			iq_set_tmp = sqrtf(Idiff);
-		}else{
-			iq_set_tmp = iq_mtpa;
-			id_set_tmp = id_mtpa;
-		}
+		// Apply Flux Weakening
+		field_weakening_run(dt, &id_set_tmp, &iq_set_tmp);
 
 		// Apply current limits
 		// TODO: Consider D axis current for the input current as well.
@@ -2497,8 +2484,9 @@ static void control_current(volatile motor_state_t *state_m, float dt) {
 	state_m->vq_int += Ierr_q * (ki * dt);
 
 	//axis decoupling feedforward compensation
-	state_m->vd = state_m->vd - state_m->iq * m_pll_speed * 0.00004;//Lq
-	state_m->vq = state_m->vq + m_pll_speed * (state_m->id * 0.00002 + m_conf->foc_motor_flux_linkage);//Ld
+	state_m->vd = state_m->vd - state_m->iq * m_pll_speed * m_conf->foc_motor_lq;//Lq
+	state_m->vq = state_m->vq + m_pll_speed * (state_m->id * m_conf->foc_motor_ld +
+												m_conf->foc_motor_flux_linkage);//Ld
 
 	// Saturation
 	utils_saturate_vector_2d((float*)&state_m->vd, (float*)&state_m->vq,
@@ -2978,78 +2966,114 @@ static float correct_hall(float angle, float speed, float dt) {
 /**
 * setup mtpa variables
 *
-* @param enable
-* true or false
-*
-* @param Lsd
-* direct Inductance
-*
-* @param Lsq
-* quadrature Inductance
-*
+* this function should be called at initialization
+* and each time a new configuration is set
+* it calculates 2 constants, k1 and k2, needed to speed up run time
 */
-void mtpa_setup(bool enable,float Lsd, float Lsq){
-	float Ldiff = Lsd - Lsq;
+static void mtpa_setup(void){
+	float Ldiff = m_conf->foc_motor_ld - m_conf->foc_motor_lq;
 	mtpa.k1 = 8.0 * SQ(Ldiff);
 	if( Ldiff != 0.0 ){
 		mtpa.k2 = mtpa.k2 = 1.0 / (4.0 * Ldiff);
 	}else{
 		mtpa.k2 = 1.0;
 	}
-	mtpa.enable = false;
 }
 
 /**
 * Implements MTPA algorithm
+* mtpa_run should be called first in order to properly calculate k1 and k2
+*
+* @param *id
+*	pointer to the current at the d axis that should be regulated
+*
+* @param *iq
+* 	pointer to the current at the q axis that should be regulated
+* 	this is the input parameter of the MTPA block
+* 	as control current from the torque or speed loops is generally
+* 	set through m_iq_set
 */
-static void mtpa_run( float Is ,float *id, float *iq){
-	if(mtpa.enable == false){
-		return;
-	}
-	float Is_sq = SQ(Is);
-	*id = (sqrtf(SQ(m_conf->foc_motor_flux_linkage) + mtpa.k1 * Is_sq)
-			- m_conf->foc_motor_flux_linkage ) * mtpa.k2;
+static void mtpa_run( float *id, float *iq){
+	if(m_conf->foc_mtpa_enable){
+		float Is_sq = SQ(*iq);
 
-	float Idiff = Is_sq - SQ(*id);
-	if( Idiff < 0.0 ){
-		Idiff = 0.0;
-	}
+		*id = (sqrtf(SQ(m_conf->foc_motor_flux_linkage) + mtpa.k1 * Is_sq)
+				- m_conf->foc_motor_flux_linkage ) * mtpa.k2;
 
-	*iq = SIGN(Is) * sqrtf(Idiff);
+		float Idiff = Is_sq - SQ(*id);
+		if( Idiff < 0.0 ){
+			Idiff = 0.0;
+		}
+
+		*iq = SIGN(*iq) * sqrtf(Idiff);
+	}
 }
 
 /**
  * initializes FW variables
  */
-static void flux_weakening_setup(void){
-	flux_weakening.enable = false;
-	flux_weakening.vs_int = 0.0;
-	flux_weakening.ki = 0.02;
-	flux_weakening.kp = 0.02;
+static void field_weakening_setup(void){
+	field_weakening.mod_int = 0.0;
+	field_weakening.is_sq_max = SQ(m_conf->l_abs_current_max);
+	field_weakening.anti_windup_error = 0.0;
 }
 
 /**
-* Apply flux weakening
-*
-* @param vs_target
-*	the voltage at which FW will start inject negative d current
+* Apply field weakening
 *
 * @param dt
 *	The time step in seconds
 *
-* @param *id_fw
-*	id current output
+* @param *id
+*	pointer to the current at the d axis that should be regulated
+*
+* @param *iq
+* 	pointer to the current at the q axis that should be regulated
 */
-static void flux_weakening_run(float vs_target, float dt, float *id_fw){
-	if(flux_weakening.enable){
-		float vs = sqrtf(SQ(m_motor_state.vd) + SQ(m_motor_state.vq));
-		float vs_err = vs_target - vs;
-		float output = flux_weakening.vs_int + vs_err * flux_weakening.kp;
-		utils_truncate_number(&output, -1.0, 0);
-		flux_weakening.vs_int += vs_err * (flux_weakening.ki * dt);
-		*id_fw = output * m_conf->lo_current_max;
-	}else{
-		*id_fw = 0;
-		flux_weakening.vs_int = 0;
+static void field_weakening_run(float dt, float *id, float *iq){
+	if(m_conf->foc_field_weakening_enable){
+//		//PI block
+		float Mod = sqrtf(SQ(m_motor_state.mod_d) + SQ(m_motor_state.mod_q));
+		float Mod_error = Mod - 0.8;
+
+		float int_error = Mod_error - field_weakening.anti_windup_error;
+
+		field_weakening.mod_int += int_error * m_conf->foc_field_weakening_ki * dt;
+
+		float output = field_weakening.mod_int + Mod_error * m_conf->foc_field_weakening_kp;
+
+		float windup_error = output;//first we save output previous to truncate
+		utils_truncate_number((float*)&output, 0.0, 1.0);
+		windup_error -= output;//calculate difference generated after truncate
+		field_weakening.anti_windup_error = windup_error * 0.1; //save error
+
+		//calculate id
+		*id += -1.0 * output * m_conf->lo_current_max;
+
+		//in CONTROL_MODE_CURRENT, we set the maximum current is_max to the current set
+		//in other modes, maximum current is the on
+		float max_is_sq;
+		if(m_control_mode == CONTROL_MODE_CURRENT){
+			float m_iq_set_sq = SQ(m_iq_set);
+			if( field_weakening.is_sq_max > m_iq_set_sq ){
+				max_is_sq = m_iq_set_sq;
+			}else{
+				max_is_sq = field_weakening.is_sq_max;
+			}
+		}else{
+			max_is_sq = field_weakening.is_sq_max;
+		}
+
+		//check if we exceed Is limit
+		float id_sq = SQ(*id);
+		float is_sq = id_sq + SQ(*iq);
+		if( is_sq > max_is_sq ){
+			//if Is is higher than limit, iq will be limited
+			float Idiff = max_is_sq - id_sq;
+			if( Idiff < 0.0 ){
+				Idiff = 0.0;
+			}
+			*iq = SIGN(*iq) * sqrtf(Idiff);
+		}
 	}
 }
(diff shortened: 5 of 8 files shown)


vedderb:
This is very nice!

I'm currently looking through the diff, here are some comments so far:

* The virtual motor should probably be re-initialized in mcpwm_foc_set_configuration in case the address of the configuration changes (does not happen currently, but it can potentially become a future bug). Re-registering the same terminal commands is no problem.

* The axis decoupling is nice, but it currently works on random parameters unless they are set manually, and is always active. The best way probably is to remove L and replace it with Ld and Lq. I can look into that.

**Field Weakening**
For the field weakening (FW) I think a completely different strategy might be better. I need to think about it more, but I imagine something like this:
* FW is only active if the q axis modulation is greater than 95% of maximum and the q axis current is lower than the set current. Thus there is no room left for regulation. If that is not the case, the d axis current is gradually ramped back to 0.
* The error is based on the difference between iq and iq_set. We try to always reach the q axis set current.
* The total current has to be truncated to lo_current_max.
* The mechanism for releasing the motor when a current lower than the minimum current is requested has to be changed. If there is d axis current we need keep the modulation on until it is ramped back to 0.
* FW has to respect the speed limits. If it respects lo_current_max that can happen automatically, as lo_current_max is updated by the speed limits in mc_interface.
* There should be an option to limit FW to an RPM where the motor just is below the RPM where it would push current back if all FETs are released (which probably is close to the maximum BLDC RPM). I think this option is _very_ important, and should be enabled by default. It would mean that FW keeps the motor in a safe speed at all times, and still allows to generate torque all the way until that speed is reached. For most motors that should be a significant improvement.

I think this approach will be stable even when requesting low(er) or negative currents at high speed (I hope this makes sense). Even when the set current is negative and the current is more negative and modulation is at maximum, FW will activate and prevent unwanted braking.

To test FW and make sure that it is stable I would recommend the following:
* Use the arrow keys and push the current to a speed that requires FW. When releasing the key, modulation should stay on until the motor has slowed down enough for id to go back to 0, after which modulation should be switched off (motor released).
* Manually push the motor to a speed where it generated back emf higher than the input voltage when the modulation is off. I think that modulation should _not_ switch on in this case, as that might cause strange behavior during faults.
* Test what happens when a fault occurs during FW.

maxicor88:
Thanks Benjamin for your comments, I will try to answer point by point.

About reinitializing virtual motor, it is easy and I will add it in a next commit.

About axis decoupling parameters, I agree with you it can be better to have only Ld and Lq but I did not want to touch the measurement of inductance functions, maybe it's better you touch that. What I have done related to this in Vesc Tool PR is to charge Ld and Lq with the same value as L once you got the apply button.

About Field Weakening comments, I tried to follow a similar strategy I have seen in many papers and in Texas Instruments approach. I think that your concerns can be solved modifying some lines. 

For instance, I am taking as input error signal the absolute modulation, I can change that for q axis modulation only easy. I am taking as limiting current the absolute max current, I can change that for lo_current_max and test.

About your comments of taking care of the id current if the system is stopped, or a low current is set, it was something I did not remember to add and I agree it is very important. I will try to think in a way to control this.

We will test the code further considering your observations. What I have done so far is to always test the code first with the virtual motor attached, and once I was sure it was running well, I tested with out lab motor test bench. What I was very happy about is that many times I have seen the same behavior with the virtual motor and with the real motor, so it is very useful. This way, with the virtual motor, I could change the load while running and see how the speed or current changes.

I am sending this message now, because I am running low battery at the cell phone...then I will add some more comments when I reach the laptop.

Thanks again for considering our PR.

maxicor88:
Hello Benjamin, in last commits I started to add some of the changes you proposed lately.

1. Virtual_motor is reconfigured when a new motor config is set. To do this in a cleaner way, I created a function inside virtual_motor library that grouped all the variables that should be updated when motor configuration changed. This also helped me to clean up the code and delete an auxiliary function in mcpwm_foc.c that now it is unnecessary to have.

2. Now Field Weakening error signal is taken from the difference between the modulation in the q axis and a threshold value. 

3. Total current is limited at lo_current_max.

4. We are still working on your comments of how to control id when a low current command is set. The way we implemented so far is very simple and will be upgraded. We are adding a new motor control mode called CONTROL_MODE_RAMP_DOWN_ID, that will be in charge of this operation, and will need to replace the different calls that stop the pwm signals and stop the motor (as that might be dangerous under field weakening) to a function that activates this new control mode. Once inside that mode, we will try to make the current drop at the same rate as the speed does, if this speed can cause dangerous voltages. 

But we wanted to show you how are we dealing so far.

Thanks

rushijariwala95:
Hi,

Is there a code available for measuring Ld and Lq inductance? Or is there any plan to implement that? 

Also, any idea what would be the impact of Ld, Lq saturation with higher currents?

nitrousnrg:
No plan to implement Ld Lq parameter detection on my end. We focus on big machines which are few and high quality, you can manually measure the inductances once and use the values for all motors of the same model.
Nothing stops you or Benjamin from developing Ld and Lq detection, its likely to gain a % of extra torque in some motors.
You would need to make the motor rotate while you measure the inductances.

Saturation behavior is yet to be seen, the observer struggles with saturation, but in our case these motors usually come with a resolver or encoder.

rushijariwala95:
Thanks Marcos for the quick reply.

Regarding field weakening, I believe the block diagram showed in TI document should work. I have seen it in more papers. MATLAB also follows the same approach in their simulation (pe_pmsm_field_weakening_control). It is similar to the concept shown in Fig 3(c) of the attached paper. 
[Full paper - EPE 2011.pdf](https://github.com/vedderb/bldc/files/3185352/Full.paper.-.EPE.2011.pdf)

rushijariwala95:
Hi Marcos and Benjamin,

I am working on a motor controller from some months now. I have a few queries which I tried searching on a lot of forums on endless sphere or Github discussions. I did not find much success. 

1. How do you size a DC-link capacitor for FOC control? Do you have any documents or is there any discussion that I can go through? I know the rms ratings of the current is the deciding factor. However, I am not sure how to analyze/calculate it. Any help from your side would be great. 

2. Is there any explanation on the sensorless algorithm used in VESC controllers or any documents/discussions I can go through?

There are a lot more queries which I keep for some later time.

You'll have done a great job in making the VESC controller open-source. I will try to contribute as much as I can with my knowledge in the field of power electronics and motor control.

nitrousnrg:
I'm glad to see you on board, but this is not the place for those questions. Please head to the forums for that, this place is created for the sole and specific purpose of discussing the Field Weakening and MTPA integration into the vesc codebase. Anything else is noise for the developers.

gowrav:
Just adding the note to the discussion of safety implementation in Inverter Circuit under Field Weakening Mode.

https://patents.google.com/patent/US7554276B2/en

gowrav:
Field Weakening on Deceleration blew the MOSFET.

https://vesc-project.com/comment/3801#comment-3801

nitrousnrg:
I'm following your case. There is a key missing feature in this branch, as maxi commented, which is the lack of a controlled rampdown while FW is active.
Let me know when the images in your post are fixed so we can take a look

gowrav:
> I'm following your case. There is a key missing feature in this branch, as maxi commented, which is the lack of a controlled rampdown while FW is active.
> Let me know when the images in your post are fixed so we can take a look

The images are now visible...

maxicor88:
> > I'm following your case. There is a key missing feature in this branch, as maxi commented, which is the lack of a controlled rampdown while FW is active.
> > Let me know when the images in your post are fixed so we can take a look
> 
> The images are now visible...

I am trying to replicate your case, I answered your post at vesc forum.

javmarina:
Any update on this?

Teslafly:
Now that HFI is working in fw 4.0, and with it a way to measure ld and lq directly, this may be worth progressing further?

Would field weakening be a better solution than overmodulation in all cases?  Even in motors that are poorly suited to field weakening? Overmodulation would at least be able to match the maximum speed of the bldc mode.

nitrousnrg:
Yes, the time has come. I merged master to this branch locally and will be testing this week starting with a small IPM motor and then a large one.

All I know is field weakening performance depends on motor construction, not all motors have a flat constant power region.

nitrousnrg:
So a small update, I have basic FW working, going 50% beyond base speed with id=-30A, small IPM motor. Its being limited by the phase current limit set at 60A, it can go faster with higher limits.

The code now calculates the base speed ( vbus*flux linkage / sqrt(3) ) and when you command 0amps it will set iq=0 and wait until rpm is below base speed to fully stop pwm.
Previously it had a id rampdown that stopped pwm in a wrong place which I think is why gowrav got some blown fets.

Waveforms look ok with the motor unloaded
image
When I load it it stops working but I think its poor observer tuning, I still don't dare trying this in the bigger motor with resolver to check loaded operation.

Next week I will spend some time tuning sensorless operation and preparing a push to get this remote branch a bit more in sync.

FW will need an extra fault management mode in which when a DC Link overvoltage is detected, SVM is set to v0 so all bottom mosfets are ON and the motor phases get shorted until the overvoltage is gone. This should have priority over an overcurrent fault.

rushijariwala95:
Great work! 

Just a query - when u turn on all the lower FETs during over voltage condition - will it cause the motor to brake suddenly? Or will it just coast down to 0 speed?

nitrousnrg:
> Just a query - when u turn on all the lower FETs during over voltage condition - will it cause the motor to brake suddenly? Or will it just coast down to 0 speed?

It would be a dramatic braking, with possible loss of vehicle control. The energy dumped into the motor windings would be massive and the event should be short, as it only needs to slow down to non-destructive speed. Possible loss of vehicle control is usually better than a certain fire.
A similar thing would happen if pwm stops for some reason during FW, in this case the energy would be dumped into the battery instead of into the motor, and braking power would be smaller.

This motor short event would be very rare, because I don't think vesc will allow FW speeds capable of blowing a powerstage. For example if you have 12s batt with 60V fets, you don't have headroom to safely increase the top speed.

gowrav:
> So a small update, I have basic FW working, going 50% beyond base speed with id=-30A, small IPM motor. Its being limited by the phase current limit set at 60A, it can go faster with higher limits.
> 
> The code now calculates the base speed ( vbus*flux linkage / sqrt(3) ) and when you command 0amps it will set iq=0 and wait until rpm is below base speed to fully stop pwm.
> Previously it had a id rampdown that stopped pwm in a wrong place which I think is why gowrav got some blown fets.
> 
> Waveforms look ok with the motor unloaded
> image
> When I load it it stops working but I think its poor observer tuning, I still don't dare trying this in the bigger motor with resolver to check loaded operation.
> 
> Next week I will spend some time tuning sensorless operation and preparing a push to get this remote branch a bit more in sync.
> 
> FW will need an extra fault management mode in which when a DC Link overvoltage is detected, SVM is set to v0 so all bottom mosfets are ON and the motor phases get shorted until the overvoltage is gone. This should have priority over an overcurrent fault.

🤟🏽Nice to see things coming together., can you share the updated code on a branch for us to fork and test with our motor [unloaded] currently., we are using A200Sv2 from Euan Mutch , so the IRF7769 can handle upto 100v Vds while braking.

My rig is not IPMSM , but I can give my feedback if it helps with a 6kW SPMSM water cooled motor with hall sensors.

nitrousnrg:
I'm working on a push. Yesterday I was setting up a vesc75/300 paired with a turnigy 6364 outrunner to see how it works there.

Base speed was 20kerpm, with id=-10A it reached 22kerpm
image

After some real use we will find out if this is ok, right now the FW controller is running very limited, you can see I commanded 14.5A but iq is only 3.5A. The goal is to follow the commanded iq=14.5A even if it maxes out the phase current limit with a very large id, which may require a bit of algorithm change.

This test was done because I can't get this to run with my small encoderless IPM motor under load. I think Ortega's observer doesn't work with IPM machines so now I'm looking for observers made for high-saliency motors.

nitrousnrg:
I use this attached .xml file in vesc_tool/res/config/4.01 to enable the new fields in the GUI.

[parameters_mcconf.zip](https://github.com/vedderb/bldc/files/4186319/parameters_mcconf.zip)

rushijariwala95:
I feel you cannot pass 14.5A of iq current if you have maxed out your phase current. 

Generally, iq_ref = sqrt(is_ref^2 - id_ref^2) in field weakening zone. 

You might use the other strategy as well - where you keep iq_ref constant and increase id_ref.

The difference between the first and second case being - 
iq_ref is limited to sqrt(is_ref^2 - id_ref^2) in the first case and in the second case iq_ref is limited to sqrt(is_max^2 - id_ref^2). 

I am not sure which method is better. 

Any thoughts on this?

nitrousnrg:
For traction applications I'm pretty sure its better to keep iq_ref constant. Right now I'm focused on getting the IPM to work sensorless but I didn't want to keep the FW merge to myself. FW likely needs some changes to feel right.

gowrav:
@nitrousnrg , here is the configuration I am working with .. 
[mc_conf_6kw_hall_ft.zip](https://github.com/vedderb/bldc/files/4192003/mc_conf_6kw_hall_ft.zip)
Inductance: 45.14 uH, ld_lq_diff: 3.75 uH (8.67 A)

Id seems to be exactly same even when Imax, Kp, Ki values are tuned.
The max ERPM changes from 5300 to 5400 when Hall sensor is used instead of Sensorless but this stands the same without enabled Field Weakening in the tool.

2020-02-12-175601_1680x1050_scrot
2020-02-12-175727_1680x1050_scrot

PS: the connect_virtual_motor 0.025 0.001 48  resets/disconnects the VESC from the VESC_TOOL..

Any clue what needs to be changed... this is SPMSM, and via simulation we have figured at 100Amps Id we can increase the speed by atleast 40%

nitrousnrg:
> PS: the connect_virtual_motor 0.025 0.001 48 resets/disconnects the VESC from the VESC_TOOL..

That's likely a saturated scheduled leading to a watchdog reset because simulation uses extra cpu time. Try decreasing the switching frequency. We should use the measured FOC loop processing time to prevent this issue in the future, many boards got bricked that way.

I'm coding a new rotor angle observer suitable for MTPA that I would like to test first in the simulator so I'll try to get the sim working asap.

gowrav:
> rotor angle observer

No change when tried with 6kHz , 10kHz and 20kHz switching frequency.., keeps resetting the USB communication but I feel your point is valid., which version of GCC are you using ?

gowrav:
Linking resource material for various Flux-Weakening Control Algorithms 
[10102336.pdf](https://github.com/vedderb/bldc/files/4222856/10102336.pdf)
Page 42 to 54 in the attached pdf discusses about the suitable algorithms for 
IPMSM which is Voltage & Current Limited Maximum Torque (VCLMT) 
and 
SPMSM which is Constant Voltage Constant Power (CVCP) and Constant-current-constant-power (CCCP) however since the later needs inverter capability to voltage boost this might not be possible without significant hardware changes.

gowrav:
It Happened.,🤟🏽
Needed to tweak FW Kp and Ki values for it to kick in ...I mean really KICK in.
Here are the observations..
with Hall Sensors the FW is quite unstable and noisy., it even got the VESC disconnected while its running.

The FOC Parameters for my setup
Hall Sensor based SPMSM Motor ~6kw
20khz for PWM to avoid WatchDog Reset and Kp Ki to Kick Start FW
foc_state variables before FW
Without FW
FOC at FW

The Ortega observer is bit better than Hall sensor but still not very consistent as the sensorless operation is relatively less noisy; adding pictures for closure. 

foc_state variables during FW

Ramp down from FW to stand still

With around -50 Amps Is on No Load condition it gave me a speed boost of 12% from 4900ERPM to 5500ERPM.

nitrousnrg:
Thats good to hear. I have one motor with hall sensors and FW doesnt work, I recall it injecting 5A Id while the same config and motor running sensorless did Id=-40A
So there is something funny going on with the rotor angle.
I'm still working in a new observer, fixed the virtual motor disconnection but it looks like the virtual motor feature is broken in the master branch, the virtual motor doesn't spin so I have to take a look.

tdaede:
Do you have an updated set of .xml files for vesc_tool to test this patch? I'm trying this patch and am not seeing any difference with field weakening enabled. (Also, for some reason, FOC realtime D/Q current isn't being logged...)

nitrousnrg:
Today I blew a controller with the current code so its not ready for use, ran ok at light loads and gate drivers blew at higher currents. Had FW disabled and MTPA enabled, but it also had an pcb assembly error so I'm not sure what caused the damage.
In any case its likely this will damage your vesc

This branch is still standing on firmware v4.1, so I drop that xml into vesc_tool/res/config/4.01 
[parameters_mcconf.zip](https://github.com/vedderb/bldc/files/4287272/parameters_mcconf.zip)

Something odd I found is that FW only works in one direction, at least with this motor. Had to swap phase wires to make it work the other way around.

gowrav:
> Today I blew a controller with the current code so its not ready for use, ran ok at light loads and gate drivers blew at higher currents. Had FW disabled and MTPA enabled, but it also had an pcb assembly error so I'm not sure what caused the damage.
> In any case its likely this will damage your vesc
> 
> This branch is still standing on firmware v4.1, so I drop that xml into vesc_tool/res/config/4.01
> [parameters_mcconf.zip](https://github.com/vedderb/bldc/files/4287272/parameters_mcconf.zip)
> 
> Something odd I found is that FW only works in one direction, at least with this motor. Had to swap phase wires to make it work the other way around.

Any update on the new Observer ?., what could be the next course of action ?. simply trying our luck would not do I suppose ...

vedderb:
Can you give the dev_fw_5_00 branch a try? I made some quite significant changes to the observer gain calculation and scaling, and it is working much better for me now. I also added axis decoupling and back emf decoupling to the current controller, which seems to help a bit.

I have also been working on a Qt-based  VESC and motor simulator using your virtual motor code, that I plan to publish soon. In that simulator I played a bit with changing ld-lq, and the observer seems to handle it quite fine with the correct settings.

nitrousnrg:
> Can you give the dev_fw_5_00 branch a try? I made some quite significant changes to the observer gain calculation and scaling, and it is working much better for me now. I also added axis decoupling and back emf decoupling to the current controller, which seems to help a bit.

Oh I wasn't following the dev_fw_5_00 branch, thanks for the heads up. I'll give it a try with an axiom because I keep wrecking bootstrapped hardware. Have you tried this FW code?

In my FW testing I had to disable axis decoupling as it was working poorly.

> I have also been working on a Qt-based VESC and motor simulator using your virtual motor code, that I plan to publish soon

Awesome!

> Any update on the new Observer ?., what could be the next course of action ?. simply trying our luck would not do I suppose ...

While I was implementing a new observer I found a way to better tune Ortega's observer, so I'm using the IPM motor in sensorless mode now and managed to achieve FW speeds. I'm having delays due to hw damage, being a bit sick and these covid-19 restrictions, but I'm focused into this feature.

supercrazysam:
Any update on Field Weakening? Still waiting for this major feature to be integrated into firmware.

hicham-barhoumi:
Any update on this ? it is an excellent feature to have, can't wait

ddosoff:
I am extremely interested in Field Weakening. Ready to sponsor development.

bj97301:
How can we make this happen?  Need more beta testers?  Money?

maxicor88:
Just to inform the guys that are interested in this pull request, we have been working a lot in testing our algorithm. We have had some delays in between because of our daily jobs, but in the meantime we have built a dyno were we can test 2 motors, one acting as a drive, and the other as a load. This helps us a lot in bench testing our boards.
We want to be very careful about this feature, as it can be risky not only for the boards, but for people also ;-) 

Right now we are able to reach FW properly, but we are working on the different scenarios to get out of this region in a proper way. We are really willing to push this feature soon. At least a first revision were others can test.

ddosoff:
> ... we have been working a lot in testing our algorithm.

Please share some payment method, I want donate for this feature. Really need FW on VESC.

nitrousnrg:
> Please share some payment method, I want donate for this feature. Really need FW on VESC

I would encourage to direct donations to the general VESC project:
https://vesc-project.com/donations

Cheers

jokucale:
Hello all 
regarding 
"I would encourage to direct donations to the general VESC project:
https://vesc-project.com/donations. "
I would donate 50 Euros for this Feature and addtional 50 Euro after test at a ebike system. 
My question now shall I add a special tag like " FluxWeaking -MTPA " at my donation, to make it vissible ?

nitrousnrg:
> My question now shall I add a special tag like " FluxWeaking -MTPA " at my donation, to make it vissible ?

Yes, its a good call. Thanks!

ddosoff:
> > Please share some payment method, I want donate for this feature. Really need FW on VESC
> 
> I would encourage to direct donations to the general VESC project:
> https://vesc-project.com/donations
> 
> Cheers

Sent 200.00 EUR to vedder87@gmail.com

Wish you luck with FW!

vedderb:
Thanks! I hear you and will start looking into it now and on Sunday. I exchanged a few emails with Maxi, so I have an idea where to start. If it goes well there will hopefully be something to test by the end of Sunday (no promises, but I will do my best). From then we have to spend some time on extensive testing to make sure that all situations and cases are handled well.

vedderb:
First attempt done! https://github.com/vedderb/bldc/commit/feb1be5fbd6aa9916c1ce0ad918afe70c1b55fec

I also made sure that the modulation is kept running when 0 current is requested until the FW-current drops to 0. That should avoid pushing a lot of current though the body diodes. FW is active only in the current and speed control modes, but when other modes are entered while in FW it will stay active until the FW-current drops to 0. One risk now is that getting fault codes while in FW can cause the FW-current to run in the body diodes, so make sure that the controllers run stable without generation fault codes before activating FW.

I have uploaded a build here to where you can give it a try:
https://vesc-project.com/node/2859

These are the FW-parameters at the moment:
image
It should be quite simple to test it, just set the maximum FW current to a value > 0.

I have tested this on a small motor and pushed it to twice the no-fw speed, and it seems to work well so far, even with 0 ramp time (then the FW-current will be scaled directly to the duty cycle).

Please be very careful when testing and let me know how it works.

bj97301:
> First attempt done! https://github.com/vedderb/bldc/commit/feb1be5fbd6aa9916c1ce0ad918afe70c1b55fec
> 
> 
> 
> I also made sure that the modulation is kept running when 0 current is requested until the FW-current drops to 0. That should avoid pushing a lot of current though the body diodes. FW is active only in the current and speed control modes, but when other modes are entered while in FW it will stay active until the FW-current drops to 0. One risk now is that getting fault codes while in FW can cause the FW-current to run in the body diodes, so make sure that the controllers run stable without generation fault codes before activating FW.
> 
> 
> 
> I have uploaded a build here to where you can give it a try:
> 
> https://vesc-project.com/node/2859
> 
> 
> 
> These are the FW-parameters at the moment:
> 
> image
> 
> It should be quite simple to test it, just set the maximum FW current to a value > 0.
> 
> 
> 
> I have tested this on a small motor and pushed it to twice the no-fw speed, and it seems to work well so far, even with 0 ramp time (then the FW-current will be scaled directly to the duty cycle).
> 
> 
> 
> Please be very careful when testing and let me know how it works.

That was quick!  Nice work. My vesc throw fault codes at the moment otherwise I would give it a try. Let me know if you need me to do some tests after the fault issue has been worked out.

hicham-barhoumi:
Nice 🥇 , i ll guive it a try, i have a free wheeling hub motor so even when i have a fault there is not much quinetic energy of the rotor only(less than 1kg). And a safe 36v voltage setup, good margin to the limit of the controller. Think tomorrow or the day after.

vedderb:
I did some more testing and added an improvement that makes it less likely to destroy MOSFETs when leaving field weakening by leaving the modulation running for at least one second after the fw-current drops to 0. The updated version is now built as well:
https://vesc-project.com/node/2859

I also did some experiments with axis decoupling and noticed that it is more important when using field weakening. Here are some plots:
image
These tests are done with the ramp time set to 0. Cross decoupling tends to make the current controllers a bit unstable and using the filtered currents helps a bit.

hicham-barhoumi:
Is thermal throttling considered in FW ? Because in that region you can't lower the fw current safely.

vedderb:
> Is thermal throttling considered in FW ? Because in that region you can't lower the fw current safely.

Both the D and Q axis current are limited gradually close to the thermal limits. This should slow down the motor gradually and possibly also result in braking current if it is running too fast. As far as I can tell that should be smooth and safe as long as the temperature does not rise way too fast.

Kezii:
I found that with an unloaded motor, with brake/regen current set to 0, the motor tends to stick at max duty when field weakening current is set, any current / duty control stops responding with the motor stuck at max rpm

I guess this is not a big problem since most applications are loaded and with regen current, but this can be kinda scary

vedderb:
> I found that with an unloaded motor, with brake/regen current set to 0, the motor tends to stick at max duty when field weakening current is set, any current / duty control stops responding with the motor stuck at max rpm
> 
> I guess this is not a big problem since most applications are loaded and with regen current, but this can be kinda scary

That should not happen unless your current measurements have some bad offsets and/or the motor inductance is really low. Can you give some details about your setup and configuration?

Kezii:
> That should not happen unless your current measurements have some bad offsets and/or the motor inductance is really low. Can you give some details about your setup and configuration?

I'm testing with a dual fsesc 6.6 and a 120100 35kv freerchobby motor, I did the offsets calibration several times and maybe one time the motor managed to shut down properly, the fsesc current sensing is not very high quality to be fair

Detected motor inductance is 26,46 µH with 17,1 mΩ

It exhibits this behavior even with only 1A of field weakening, I have to brake the motor with my hand to exit the field weakening region and then it starts responding again

the rest of the setup is everything default, with low-ish motor and battery current and no regen (i'm from an ac power supply)

vedderb:
> I'm testing with a dual fsesc 6.6 and a 120100 35kv freerchobby motor, I did the offsets calibration several times and maybe one time the motor managed to shut down properly, the fsesc current sensing is not very high quality to be fair
> 
> Detected motor inductance is 26,46 µH with 17,1 mΩ
> 
> It exhibits this behavior even with only 1A of field weakening, I have to brake the motor with my hand to exit the field weakening region and then it starts responding again
> 
> the rest of the setup is everything default, with low-ish motor and battery current and no regen (i'm from an ac power supply)

Can you try to allow 5A or so regen? If the current measurement is bad it might not allow the modulation to decrease when that violates the regen limit. I have seen that happen on hardware with poor current measurement.

Kezii:
After a long power cycle now it works smoothly, I tested for several minutes without issue, if it happens again I'll increase the regen current, indeed the current measurement is very noisy and it tops at about 5A

Thanks

maxicor88:
I run some tests today with latest code from dev 5_03 branch in Axiom board, and the results were promising.

I did not push it too hard, used a max duty of 40% and start field weakening at 30%, and a field weakening current of 2Amps. The max speed went from around 8000 erpms to 8600 erpms. This was unloaded. 

I will try to run some tests during the week using the dyno.

I will eventually also try to use the algorithm we were working on, and compare results of both.

Great work so far.

nitrousnrg:
> My vesc throw fault codes at the moment otherwise I would give it a try. Let me know if you need me to do some tests after the fault issue has been worked out.

The fault issues can't be worked out, its part of the intrinsic safety risks of running FW.

From the ~25 fault codes we have, I think we could gracefully manage only 3 of them with a coasting instead of pwm shutdown:
FAULT_CODE_OVER_VOLTAGE
FAULT_CODE_OVER_TEMP_FET
FAULT_CODE_OVER_TEMP_MOTOR

The rest of the faults compromise current measurement, angle measurement, or code integrity which renders the firmware unable/unsafe to keep id under control so the safest action is to shut down the PWM, with the consequence of uncontrolled regen current (and breaking) through the body diodes.

bj97301:
Got it. Just trying not to brick it. Sounds like the fault code issue will always be a problem right?￼￼

vedderb:
> Got it. Just trying not to brick it. Sounds like the fault code issue will always be a problem, right?￼￼

Yes, there is no way around that. The fault codes are triggered and switch the entire power stage off as a last effort to save the hardware when there is nothing more that can be done. With field weakening that is still the best that can be done in those situations, but it is much more likely to lead to hardware damage. Therefore you should make sure that your setup and all settings are stable and work well with rapid current changes before activating field weakening.

The things I have with VESCs that I use almost daily (longboards, electric motorcycle, rc car, ebike) never generate fault codes after they are configured properly. I only see them when e.g. one of the motor cables on the longboard unplugs itself or something shorts out, which is quite rare.

I have also been pushing this rig to its very limits for 3 days now while testing the field weakening, and it is still alive:
image
The white brackets are polycarbonate, so they can take over 100C without getting soft. The motors have been running with 50A FW current until thermal throttling many times now while I was provoking them with current commands, and so far everything is alive. One reason for that though is that I used a lab power supply at 20V, and when it tries to push current back the capacitance is quite limited and the diodes in the FETs are fine charging them to 30 or 40V. With a battery on the input it would probably be dead by now, but that is mostly due to my first implementation. In this entire process I got 0 fault codes, so a proper motor configuration, good hardware with decent current measurement and proper connections usually don't generate fault codes. Very low inductance motors are the exception, but I don't think field weakening is too useful on them anyway (if my understanding is correct).

maxicor88:
We tested this code https://github.com/vedderb/bldc/commit/998bc17ba717dccd306f230322ef47e7b4fcb82a in our big dyno setup (Axiom board + Leaf Motors +  Resolver position feedback). 

Without Field Weakening, the code works OK, but once we tested Field Weakening, the motor spins to full speed and keeps it there while letting off the throttle. 

This was with a FW current of 100Amps, FW start at 90%, FW delay 200ms, and Max duty of 95%.

It is worth noting, that we had this same behaviour a week ago with our own FW approach, and we were trying to figure out what was causing it.

Our guess is that we might have some misconfiguration in the resolver, or something that generates an small error in the position feedback, that might cause that when the systems is trying to control 0A Iq current, and 100A Id current, in fact our position error is making 1A Iq and 99A Id (for example), and that is enough Iq to make it go full speed in an unloaded situation.

If that is the case, I guess we should consider that this might be a pretty common situation (a tiny error in position feedback), so having some algorithm that can still lower down the speed should be useful.

We will try to run some more tests, to get more info and get back to you soon.

vedderb:
That sounds very likely. Sensorless probably has a very small position error at full speed, which most likely is why I didn't see that problem. Do you have a way to command a negative current? Then you can see if you slowly can bring it out of that situation and how much negative iq is needed to overcome the offset.

I have also seen that even without field weakening, when the motor has very low friction and modulation is left on with 0A current set, the motor can still stay running because the current offset is enough to overcome the friction. Without field weakening the modulation is switched off completely when 0A is commanded, which always will slow down the motor as the current truly becomes 0A. To test if this is what happens you can disable FW and command a current of 0.1A, which should keep the modulation running. If that also keeps the motor running then current offsets and low friction are likely to be the cause.

One way to work around this would be to add some fraction of the FW current to iq so that it will always tend to slow down the motor when the throttle is released. A new parameter could be a percentage of the FW current that is added to iq (e.g. 10% by default). Then if you have, for example 50A fw current and command 0A iq, iq will be set to -5A (10% of 50A) to ensure that the motor slows down even when unloaded.

Kezii:
> That sounds very likely. Sensorless probably has a very small position error at full speed, which most likely is why I didn't see that problem.

FWIW i got that bug with the AS5047 encoder, I remember that I thought about that too and tried sensorless, and the problem remained

nitrousnrg:
So I tried this code in Luna hardware and I can note 2 things:
* In sensorless mode it works great, it reaches max FW speed and comes back to 0rpm even completely unloaded.

* In hall sensored mode, if the motor is completely unloaded (mid-drive spinning backwards) it will stick to max rpm and I have to reboot the code to stop the motor. In the same test but spinning forwards the drive unit provides a bit more mechanical drag (there is a clutch involved) that is enough to slow it down.

Test was done with a 76V battery but max duty reduced to 35% and FW engagement at 30%. I want to test deeper FW with a 52V battery soon and with more inertia.

This is consistent to previous FW testing I did about a year ago, its pretty easy to mistune the motor and not be able to slow it down.

The code seems to be working, in my case it sounds like I could tune a bit better the hall sensor offset.

With Axiom its misbehaving badly but to me it looks like a wrong resolver offset, will keep you posted. Fun note is that disconnecting the 120V battery at FW speeds increased the input voltage to 330V.

ddosoff:
> First attempt done! [feb1be5](https://github.com/vedderb/bldc/commit/feb1be5fbd6aa9916c1ce0ad918afe70c1b55fec)

Thanks Benjamin!

A question about FW.
In current mode, If overcome motor moment and rotate it backward the motor will act as brake.  Brake moment will become less intensive when speed up in FW mode? Until zero at max speed?

vedderb:
> Thanks Benjamin!
> 
> A question about FW.
> In current mode, If overcome motor moment and rotate it backward the motor will act as brake. Brake moment will become less intensive when speed up in FW mode? Until zero at max speed?

The braking torque should never be limited by FW. As soon as you start braking the modulation will drop, which moves you out of the FW-region the more you brake.

maxicor88:
Good news on Axiom , we confirmed the issue we were having was related to an small difference in the Resolver offset angle. We run several tests, and we were getting offsets from 24 to 31 degrees. We noticed that we some angles, at a large FW current the motor was kept running at high speed, but if we tweak that angle, the motor could successfully slow down. So we end up at around 27 degrees to be the perfect setup.

Here is a video of our last tests hosted by our partner Arlin. 

https://www.youtube.com/watch?v=5DzFZv0SP9U

We still believe that it might be a good feature to have a percentage of braking setting (like 1% of Id or so), as in some applications might be hard to setup the perfect angle.

It is interesting though, that making this tests, it is like a perfect way to realize properly if the Iq you are setting up is actually generating torque or not. It seems like an indirect measurement of the angle, setting Iq as 0, and seeing if it actually generating torque or not.

We will run more tests next to see the behaviour of FW at large power levels.

Cheers

dzid26:
Is the behavior the same for rotation in the other direction? 
I was thinking maybe you are experiencing latency on angle reading or control loop.

vedderb:
Nice test, thanks for the video!

A few comments:
* Field weakening duty start is not an absolute duty, but a percentage of max duty. You can leave it at 90% or so even when you limit the maximum duty.
* The reason that the FW-current decreased for full throttle is probably that the current limit was close the the FW-current. The magnitude of both set currents was limited together, so when commanding a higher current the FW-current would get limited as well. I have now changed that so that the D-axis current has priority over the Q-axis current, so hopefully full throttle should not suddenly bring the FW-current down.

I will add the Q-axis negative current percentage tomorrow.

vedderb:
I have added the negative Q-axis current parameter now. The beta build is updated:
https://vesc-project.com/node/2859#comment-7398

hicham-barhoumi:
I did some basic test on un ebike hub motor vesc 4.12, it work perfectly with up to 7 amps of fw. no issue recovering down from fw.
image

maxicor88:
I wrote the algorithm we have been testing in Axiom boards last year, based on a PI regulator here https://github.com/vedderb/bldc/pull/289.

I tested it successfully in Axiom boards with an small motor. We will run more tests on the big dyno setup and comment.

Something to point out, is that in one of our controllers in the big Dyno, the Motor cannot slow down even at 10% of Q current. We run the encoder detection several times to discard any issues with this angle, but the issue persists. 

Will keep you posted when we run further tests.

ddosoff:
I tested FW on 75300 with QS205 4kw hub.

Without FW it have 16200 ERPM at 62.5V. With 100A FW it have 18800 ERPM with disabled 'Sample in V0 V7'. And 19800 with enabled Sample in V0 V7.

Why Sample in V0 V7 affects max ERPM?

jokucale:
Hello,

I have one question. Were can i found the VESC Tool wich support the FOC Feature? The public VESC Tool support not the FW 5.03 <dev> Version. Can somebody send me a link to a VESC Tool which support FW 5.03 or give me some advice?

Kind regards Jochen Kudicke

Gesendet von Mail für Windows 10

Von: Kirill Kostyuchenko
Gesendet: Donnerstag, 8. Juli 2021 11:19
An: vedderb/bldc
Cc: jokucale; Manual
Betreff: Re: [vedderb/bldc] Field Weakening and MTPA support (#91)

I tested FW on 75300 with QS205 4kw hub.
Without FW it have 16200 ERPM at 62.5V. With 100A FW it have 18800 ERPM with disabled 'Sample in V0 V7'. And 19800 with enabled Sample in V0 V7.
Why Sample in V0 V7 affects max ERPM?

electricrt:
I've been doing extensive road testing with field weakening and MTPA with a small IPM motor.

it seems FW is quite smooth, I'm running 90A of FW.

Was getting brake surge when lifting off the throttle at high speeds but enabling Current Control Decoupling- BEMF reduced brake surge significantly.

maxicor88:
We have been running some more tests in our dyno setup, using Field Weakening and MTPA. 

We found and issue under Field Weakening in which we see duty drops (in fact Duty changing its sign). We tried a couple of solutions to this but could not found a proper way to deal with this.

Here is a video showing our issue. https://www.youtube.com/watch?v=npjrCQJNEbc

You may see that when the duty drops happen, Vq is near 0. 

And I see from this https://github.com/vedderb/bldc/blob/5ad0c889449297f2703c4017044c299ad46105b8/mcpwm_foc.c#L2978

That it might be an issue with noise in Vq that it might be changing its sign.

We tried putting an extra filter on Vq prior to taking its sign, but with no improvements.

I even think this might be causing some drop in performance because it might be affecting the system getting out from field weaking for a short period of time because of this:

https://github.com/vedderb/bldc/blob/5ad0c889449297f2703c4017044c299ad46105b8/mcpwm_foc.c#L3090

We wanted to bring this up to see if we can get any ideas, maybe a better way to calculate the sign of Duty, instead of using Vq (which might get always very close to 0 under Field Weakening.

Thanks for the help.

ddosoff:
I found it's easy to cause ABS_OVER_CURRENT or DRV_FAULT if you use FW, push some strong current and then change direction from fastest speed.

I tested with QS180 8kw mid motor, 22s battery, 100250 Trampa VESC.
Also I tested on custom VESC hardware and got hardware 450A drv faults.

vedderb:
@maxicor88 
I had a look, and I think the problem is that the voltage vector approaches being 90 degrees ahead of the rotor. The duty cycle represents the percentage of voltage modulation and the direction relative to the rotor. If it turns negative it means that a scaled down version of that voltage from standstill would cause the motor to rotate backwards. Maybe the voltage lag from the inductance * current * rpm + the FW current cause the voltage to be more than 90 degrees ahead of a high inductance motor under high load and field weakening.

In most places in the code only the absolute duty cycle value is used. One problem is where the absolute value of the filtered duty is used, as it will start to decrease when these negative pulses appear. One of these is actually my FW implementation, so there might be some problems caused by that. Here is a fix:
https://github.com/vedderb/bldc/commit/4603ff5d4ea0c81487a26ca1b9e8d048cbbfc9e4

ddosoff:
Hello Benjamin,

I spent some hours to find good current KP KI for my 8kw qs mid with 120A FW current. It's not easy to find them good for no FW and FW regions.

I think good solution is different KP1 KI1 for Q axis and KP2, KI2 for D axis. This should solve the issue.

patrolbav:
> Good news on Axiom , we confirmed the issue we were having was related to an small difference in the Resolver offset angle. We run several tests, and we were getting offsets from 24 to 31 degrees. We noticed that with some angles, at a large FW current the motor was kept running at high speed, but if we tweak that angle, the motor could successfully slow down. So we end up at around 27 degrees to be the perfect setup.
> 
> Here is a video of our last tests hosted by our partner Arlin.
> 
> https://www.youtube.com/watch?v=5DzFZv0SP9U
> 
> We still believe that it might be a good feature to have a percentage of braking setting (like 1% of Id or so), as in some applications might be hard to setup the perfect angle.
> 
> It is interesting though, that making this tests, it is like a perfect way to realize properly if the Iq you are setting up is actually generating torque or not. It seems like an indirect measurement of the angle, setting Iq as 0, and seeing if it actually generating torque or not.
> 
> We will run more tests next to see the behaviour of FW at large power levels.
> 
> Cheers

Hi guys!
Keep in mind that motor torque depends not only Q-axis current.
This is valid only for non-salient pole motors in which inductances Lq = Ld.
This is not the case for real engines. And the torque is created not only by the Q-current, but  by D-Current too.
And the greater the difference between Ld and Lq, the higher at this moment.
Here is the moment formula for a salient pole rotor:

M = 1.5 * Zp * (Iq * W + Id * Iq * (Ld-Lq))
where W- flux linkage PM.

Therefore, even with Iq = 0 and FW turned on, there will be an accelerating moment and it can be higher than the friction moment.
The engine will not stop at the same time.

Sorry for my english :-)

vedderb:
Some updates:
We have put some major work into getting all scaling and equations right in the firmware. The main improvement is that the inductance measurement as well as the ld-lq-diff measurement is much more accurate now (they were off by more than a factor of 2 before). The observer also takes motor saliency into account now and tracks a bit better when Ld-Lq != 0 (verified with encoder). The Ld-Lq setting is now separate from MTPA and only affects the observer if MTPA is not used. The MTPA mode is now moved to the FOC->Advanced page.

I have also been working on a motor comparison tool that takes saliency and MTPA into account. It can be used to e.g. calculate how much field weakening current is required to reach a given speed at a given maximum voltage. The tool takes a desired torque and RPM and calculates all motor currents and voltages at that torque output. To deal with the effect of saliency and MTPA on the torque it runs a fixed point iteration loop to calculate the currents such that the desired torque is reached.
image
You can right-click (and drag) in the graph to update all values below it at that point. It also updates the plots and values live when any setting is changed.

Here is a field weakening example with the trampa 160 kv motor with an RPM sweep from 0 to 7000 RPM at 5 Nm load:
image
The current is adjusted so that the output torque is fixed in the test. As you can see, with 40A field weakening current it takes 52.4V bus voltage to reach 7000 RPM at 5 Nm. Without field weakening, the same would take 60V. Also note that the Q axis current is lower with field weakening active, as the D-axis current contributes with some torque because Ld-Lq is set to the measured value of the motor. Note that this tool applies the field weakening current in the entire plot, so it is intended to see what field weakening does to every operating point. I might add a mode some time where the input voltage can be fixes and field weakening is applied in the same way as in the firmware.

To attempt estimating core losses and calculating efficiencies the no-load current has to be set in the Additional Info -> Setup page. Also set the number of motor poles while you are there:
image

If MTPA is active in the config it will be used in the comparison tool as well. Here is a comparison with the trampa motor with vs without MTPA:
image
MTPA only gives a slight efficiency increase on this motor, but more importantly the bus voltage requirement goes down by quite a bit (47.6V to 43.5V) when using MTPA on it.

As usual, you can have a look at the commits in the development branch to see the changes and download a precompiled test version of vesc tool here:
https://vesc-project.com/node/2859

nitrousnrg:
I just gave it a try and its very cool. With the motors I have here it seems to provide reasonable results, but I know my detected motor parameters are not ideal (had to avoid going sensorless for a long time). I'm gearing up for testing the latest code with improved motor detection to test this better

PR #96, opened 2019-06-20
Port Ackmaniac smart reverse

DAddYE:
Hi @vedderb,

I'm planning to backport @Ackmaniac's changes back upstream.

This is the first one, and most requested by the esk8 community.

It's my first time working on this complex codebase, so, please review it carefully, I'm pretty sure I've done a few mistakes. 

If these new settings are okay with you, I'll edit as well the vesc_tool.


vedderb:
Thanks for the PR, and nice that you are getting involved with writing some VESC code! It looks good overall, just a few suggestions:

* Instead of introducing config.max_erpm_for_dir_active, add one or more new ppm_control_type such as PPM_CTRL_TYPE_CURRENT_BRAKE_REV_HYST. It would otherwise be quite un-intuitive if PPM_CTRL_TYPE_CURRENT_NOREV_BRAKE can go reverse too. When you create new enums add them to the end, so that old configurations still are compatible as the enum is stored as an integer.

* Does the adc_config struct really need the updates? The ADC app does not have them.

* confgenerator has to be generated from VESC Tool. The signatures are generated from all variable names, so they must be correct for the configuration to be accepted. Updating confgenerator is actually quite easy using the parameter editors. Just keep in mind that VESC Tool has to be rebuilt with the new XML files before the new confgenerator can be generated. A tutorial for this is probably needed soon...

Also, did you test the code with the updates?

DAddYE:
Hi!

thanks so much for your feedback. 

I'll address the changes you requested and test them properly.

DAddYE:
Hi @vedderb, I've updated the code with your suggestions.

I'll work on the vesc_tool now, and submit a proper pr for it, and the configenerator.

PR #99, opened 2019-08-11
Add self balancing App (EUC)

Mitchlol:
I have added a self balancing app to project. I hope you guys are willing to accept this addition. I'd really like to get the ability to build DIY electric unicycles into more peoples hands. I know its not perfect in every way but I think it is a great starting point for the community.


vedderb:
Very nice! Have you tested if it works well?

One suggestion: instead of adding use_peripheral to imu, you can use imu_init_mpu9x50 or imu_init_icm20948 directly from the app. Then use_peripheral in the app can be an enum with which IMU to use as peripheral, where the internal IMU is the first option.

I'm currently working on some local changes that I will push in a few days, and one of them is adding support for the BMI160 IMU from bosh. I will also add an imu_stop function to the IMU driver so that it is possible to switch from the internal IMU to an external one.

Mitchlol:
"Have you tested if it works well?" 
I have tested and even used it to commute to work. That being said, I can't say with confidence that it works well. The build I'm using has a little oscillation that i cant seem to fully tune out (worse with higher madgwick beta, which is better at maintaining attitude on spins). I have another build with a different motor, that I suspect may work a little better, but i haven't tested yet. I'd say its better than the worst of the commercially available products, but still quite a bit worse than a modern top model from the major brands. 

By trade I'm just a lowly software dev, and all this control system engineering is over my head. I try lots of random things, and most of the time they have no affect. Many long nights were spent trying to figure it out, on various iterations of the project using an arduino and pwm signals. Eventually I finally decided the best bath forward was to get all the processing done on one chip, and to give up on perfect tuning. Instead I just want to get the tools out there so other people can try it, and hopefully get more collective brainpower on the problem. This is also why i tried to put every parameter that can be tuned in the vesc_tool. With the vesc's availability, and the ease of use of the vesc tool, as far as i know, there is no other open source/diy option that even comes close to this setup. However it may be wise to mark the app as beta/experimental to help set users expectations.

I also left off some features id like to add in the future. Some of that is related to the Gyro config.

It seems to me that the gyro functionality is separate from both app & motor configuration. If it were fully up to me and I had infinite dedication. I would implement gyro config as a 3rd setting group with support for configuring which gyro to use (on board/off), which gyro model, onboard gyro settings (scale, low pass filter, etc), sensor fusion algorithm (mahony/madgwick/complementary) and its parameters. However with this initial set of features I tired to be minimally invasive.

But actually it turns out my code has a gyro setup bug anyways, on the vesc6 the init_gyro() in main will overwrite the one setup by my app's startup. Maybe its best for me to wait and see what you do with the gyro changes before I do anything else.

Regardless I'm open to implementing any changes you say are necessary to get this code in. I will make open source EUC a thing even if it kills me :rofl: 

Here is a clip of me riding my build with the VESC. I think it works well, but I also know that is subjective.
https://www.instagram.com/p/B0wqIMUAuZV/

Sorry for the crazy long response. There is a lot in my head when it comes to this project.

supercrazysam:
@Mitchlol Thanks for your work! But may I ask what are the connections from imu to vesc?

Mitchlol:
@vedderb I have done some more testing, and it seems i can significantly improve performance by calibration sequence to detect gyro drift and then eliminate it before feeding the data to AHRS. I believe this would best be done through the vesc_tool, but this is another feature that cant really be done cleanly with the way that the gyro code is currently implemented. I'm willing to do the refactoring I mentioned above, but I don't want to put in all that effort if you wont like the changes. What are your thoughts on the future of the gyro code?

vedderb:
Can you give some details on how that calibration would work, and why it increases the performance? I have spent a lot of time working on IMUs and doing all sorts of state estimation in my other projects, and have a quite good idea of what the tradeoffs are and what causes certain problems. Maybe the the thing that causes the difference you see in performance has a completely different cause.

Mitchlol:
If you set a 6 axis gyro on a stable flat surface, the average values should all be 0 except for 1g in the direction of gravity. However that is not the case. I eyeballed it and hardcoded offsets to correct them on my gyro (-average value when stable), and the result allowed me to use lower values for the madgwick beta which previously would cause it drift off center. The lower beta values gave me much more stability in my balance loop. Resulting very stable EUC can be seen here. https://www.instagram.com/p/B1NqzwdgoL9/

The calibration process would have you set your gyro to level, and then automate the rest.

vedderb:
The gyro values should all be 0 when stationary, it is the accel value that should be one in the Z direction. If you do some sort of calibration on the gyro values you probably want to get them close to 0.

Essentially the filter works by almost entirely integrating the gyro values to get the orientation, and the accel values is only used to compensate for drift over time with low gain. Integrating the gyro to get an orientation is responsive and almost noise free, but the orientation will drift over time due to offsets and numerical errors. The acceleration allows calculating an absolute orientation without integration from the gravity vector, but that is very noisy as vibrations are large in magnitude compared to the 1G of earth. Therefore you want to rely on the gyro as much as possible, and only use the accelerometer to compensate for the drift. The beta value tells the filter how much to use the accelerometer for correcting the heading obtained by integrating the gyro. As you observed, increasing beta will increase the noise, but helps when the gyro values have offsets. For getting the gyro offsets you only need the IMU to be still, the orientation does not matter. Getting the accelerometer offsets requires being on a flat surface, but that is not really necessary in my experience.

The problem with the gyro offsets is that they are very temperature dependent, and just going from indoors to outdoors can have significant impact when the application is sensitive. Therefore I usually just leave them, and use enough beta to compensate any significant drift. For me 0.1 has been enough though, so I don't know why you had to increase it to 0.5. Maybe your IMU has some issues, or needs an extra capacitor if you have it on an external PCB. If you can get away without having to mess with the gyro offsets it makes things much simpler, otherwise you might have to get so far that you put the IMU in a temperature controlled oven (which is quite common on some drone autopilots).

If the drift really is a problem, you could try using the Mahony filter and setting mahonyKi to some small value. mahonyKi will attempt to estimate the gyro drift over time. I have had some problems with oscillations when using it though and you might want to clamp integralFBx, integralFBy and integralFBz to some value (anti-windup).

Another thing you could try is increasing the sampling frequency. Currently it runs at 200 Hz, and running at 500 Hz could help a bit.

Mitchlol:
Thanks for the info. 
Some more context: On my desk at home all the gyro axis's had a constant drift of 1-2 degrees/s. After correcting that, i put the beta way down to 0.02, and that gave me the very stable balancing. Before the correction, any beta below 0.05 would cause the calculated pitch/roll to wander forever. 

I will try increasing the sample rate, and i will also try messing with the temperature of my gyro and see what happens to the values that sounds like a fun experiment.

Most of the commercial self balancing vehicles allow the user to run a gyro calibration sequence through their apps. So I'm trying to steal their ideas :neckbeard:  

Lastly, I ordered 2 VESC6+ from trampa and they should be here monday (They're already in the US), hopefully the onboard gyro will have less noise. I also expect the motor control will work better too. The motor control on my Flipsky6 is trash compared to my Focbox4 so i think the real VESC6 will give improved results.

I'll post again when I've tried everything you suggested.

vedderb:
Very nice, thanks for the effort! Sorry for taking long with my current update, there have been many tricky things with a new upcoming hardware version and the wand firmware. I will try to push my current updates today so that you can start merging and testing, which I think should be quite easy.

Some comments again on the updates:

* When creating new enum members, always add them to the end. This way backwards compatibility won't be broken. The next versions of VESC tool will try to support as many commands as possible in limited mode so that the firmware doesn't necessarily have to be updated for every new version.

* All parameters in VESC Tool should also be present in the mobile version. The graphs don't have to be there, but it should be possible to tweak all parameters on the go. You can uncomment CONFIG += build_mobile in the qmake file to build the qml version. You essentially only have to update the ConfigPageApp.qml file to put the settings there, which should be quite simple.

* The current firmware I'm working on also has support for the BMI160, and I think it is important to support multiple IMUs. Instead of Use I2C peripheral, the setting should be something like **IMU to use**, with the options **Internal (if present)**, **External MPU9x50**, **External BMI160** and **External ICM20948**.  The current firmware also has a method to stop the IMU driver and thread, so that IMUs can be switched on the fly when a new configuration is set.

I also want to let you know that I think what you are doing is really nice, and I want to support it. If you need any VESCs or other hardware I might have just drop me an email and I will try to get it sent to you. And if you feel like visiting me in Sweden to work on the self balancing stuff for a few days and have some vacation you are very welcome :-)

Mitchlol:
Thanks for the positive encouragement!

Re: you comments.

1) I get the idea and why, I like it too. Just not sure what and where exactly in the code. Are you referring to the parameter serialization order from the UI to the Hardware? I only have one ENUM and it is currently last, so I'm OK? Also I should be OK as long as i don't put my ENUM to be serialized before any variables that existing in the previous FW release? Or am i misunderstanding this?

2) Will do. I haven't looked at the mobile code at all which is ironic, as my day job is a mobile dev.

3) I was thinking it would be 2 config fields. One to define the input pins and one for gyro model. Your way will work as well, but worries me if future models ship with different internal gyros. I guess the answer to that, is to have a HW definition for internal gyro type? Both way have pros and cons, just figured id share my thoughts, I can implement it either way.

I'm pretty good on hardware, i actually sold off 2 flipsky vescs to a coworker today. I still have 5 more working vescs ones tho, 2 are going into a mountain board that ill build eventually (sorting out motor mounts is confusing, i want to get a 2nd hand MBS board because theyre cheap and lots of options locally), 2 in my diy self balancing builds, and i also have enough spare parts for a 3rd self balancing build. But I'm curious what the new hardware you mentioned is :thinking: 

Taking a trip to sweden sounds super fun. I think I can actually do that.

vedderb:
With the enums I was thinking of the COMM_GET_DECODED_BALANCE command. Adding it to the end of COMM_PACKET_ID will make it easier to keep it backwards compatible.

I have been working a bit on the IMU configuration myself. Was hoping to finish it in the weekend, but there are a few details left. For the type selection I added

```
typedef enum {
	IMU_TYPE_OFF = 0,
	IMU_TYPE_INTERNAL,
	IMU_TYPE_EXTERNAL_MPU9X50,
	IMU_TYPE_EXTERNAL_ICM20948,
	IMU_TYPE_EXTERNAL_BMI160
} IMU_TYPE;
```

where IMU_TYPE_INTERNAL will use internal imu (if any), which is chosen based on the hardware configuration. The IMU is re-initialized when the configuration is written, so it is possible to switch between IMUs and configurations without rebooting. I also added

```
typedef enum {
	AHRS_MODE_MADGWICK = 0,
	AHRS_MODE_MAHONY
} AHRS_MODE;
```

and this configuration so far

```
typedef struct {
	IMU_TYPE type;
	AHRS_MODE mode;
	int sample_rate_hz;
	float accel_confidence_decay;
	float mahony_kp;
	float mahony_ki;
	float madgwick_beta;
	float rot_roll;
	float rot_pitch;
	float rot_yaw;
	float gyro_offsets[3];
	float gyro_offset_comp_fact[3];
	float gyro_offset_comp_clamp;
} imu_config;
```

I have tested a simple online gyro bias compensation that essentially is a high pass filter. It will not work on an axis that is rotating all the time (yaw in most cases), but roll and pitch will be upright on average for quadcopters and balancing things, so it should be fine there. I noticed however that the BMI160 imu never has more than +- 0.5 deg/s bias even without calibration, and that worked for a beta of 0.01 without problems. The MPU9x IMUs are probably worse, I will test a bit with them too. I also added constant offsets that can be set on the gyro, probably with a calibrate button or with an option at start as you did.

The other added settings are rot_roll, rot_pitch and rot_yaw which will let you rotate the IMU to any orientation in all axes in case the VESC is mounted at an angle (which it is in the scooter that I built, where I want to have wheelie control as it has so much torque). I didn't get the rotation matrix quite right yet, but I'm working on it. Hoping to finish any day now.

November, January or Febryary would work for me for a visit :-) We can take that over email if you are up for it, let me know which dates would work. I can check if I can sponsor you with flight tickets. Would be great to have you involved with some more stuff on the VESC if you are interested, the code you wrote so far looks good and you seem to have grasped the structure of the firmware and VESC Tool :-)

vedderb:
I finally pushed my latest changes. Can you give them a try? You can set gyro_offset_comp X and Y to something like 0.02 - I think that will help a lot.

Mitchlol:
These are a lot of awesome changes, but i also wrote a lot of similar awesome changes, so I'm going to have to fight my way through lots of merge conflicts.

I also got a few bmi160 gyros to try.

I experimented a bit with continuous calibration and it wasn't as simple as it seems. But i still have a few more things i want to try, as well as trying your code.

I also sent you an email to the address listed in the source code :)

vedderb:
Many of the changes were inspired from your code :-)

I find it important to get the details about how to choose IMU right, and how the initialization is done when the settings are written (e.g. stopping threads and starting new threads when changing IMU type). I also wanted to enable using an arbitrary orientation instead of choosing the axes only, as I have two builds where the VESC is mounted at a weird angle. I hope it is not too troublesome to merge your changes, otherwise I can have a look.

Mitchlol:
Already spent some time looking at the changes and conflicts. Ill get it all resolved, but i think the way i want to do it will have some weird looking git history :joy: :joy: :joy:

Mitchlol:
Ok, a little update: got all the changes merged and working, and reworked how I select axes to work with rotatable axes.

Still needs more cleanup, and i have to address your previous comments, and i have at least 2 features i think are critical to add. But i did ride it around my apartment with your changes merged!

vedderb:
Great that it is working again!

What are the two critical features?

Mitchlol:
Ok there are a few featues :joy_cat: the list seems to keep growing, the critical ones were.

1) Dual fault switch support, I was gonna leave this out of v1, but a few people asked me for it already, it is useless on an EUC style setup, but critical for mounting and dismounting a skateboard style setup. Got this feature done tonight.

2) Low voltage tiltback, this is potentially a 2 part feature, and IMO a more important safety concern. 
Part A is just tiltback to let the user know battery is dead.
Part B is scaling the tiltback duty cycle as the battery gets lower. I believe there is a feedback loop of sorts when the voltage is low, it drops lower pretty easy, so if you gas it too hard, the voltage will drop triggering a tiltback, however executing the tiltback requires more current which drops the voltage even lower, and when you combine this with the low voltage current limiting setting in the motor control it gets even less current, and pretty much the tiltback fails and and you fall off and die :joy_cat::joy_cat::joy_cat::joy_cat:. I'm pretty sure this actually happened to me, and although i was able to run it out without a scratch, I'd like to avoid this happening to anyone else. Implementing this will be weird tho, it would be cool to base it on the voltage settings in the motor config, but passing that data around gets pretty ugly.

Other things i want to do but will probably skip:
1) Add time delay to faults or at least the duty cycle cutoff, for example i suspect the over duty cycle shut off is likely to happen if you jump at full speed as the wheel generally ramps up mid air. Adding a 1-2 sec time delay before the cut off should solve this. As it is now if you set a safe tiltback, you should have a decent buffer space where the wheel can ramp a bit before cutoff.
2)  Nose angling, just a fun bonus feature, but the ability to have the nose angle up or down and have that only happen at speed (maybe > 2% duty) and based on your direction.

Mitchlol:
Ok!

Dual switch: DONE!
Low voltage tiltback: DONE! (only the basic version, but for v1 its enough to prevent injuries)

Enum move: DONE!
Mobile app: DONE? (i can build the app and see the settings but i cant connect to my vesc to test read/write (cant read firmware version, never tested this vesc with bluetooth before))
Gyro Peripheral Setting: DONE! (by you)

Ship it?

Mitchlol:
I am currently using the drift estimation, but only on the 2 axis that aren't self balancing. With it setup like this, it is the best it has been so far. I would have guessed that estimation on the balancing axis would work best since it is always stable, but actually the constant teeny tiny correction motions or maybe the motion of driving seem to throw it off (actually if driving is messing with rotation, I wonder if remounting my gyro rotated 90 degrees on that axis will have any effect).

Not sure how to explain this best, or if you're familiar with it. But if you put the beta too low and your gyro is not perfectly calibrated, your orientation will drift endlessly, you've probably seen this. However if your gyro is better calibrated and or your beta is a higher, there can still be a sort of in between state, where it will be nice and steady at first but the more you shake it, the worse it will be, and if you shake it enough it can start to drift randomly for a while till it settles down. When riding an euc this mainly manifests as the nose/tail angle drifting up and down as you turn. With gyro estimation on the balancing axis, I get a lot of this, much more than a startup self calibrate set it and forget it. However with estimation only on the other 2 axies it works even better. At this point someone who isn't me and hasn't spent months nit picking wouldn't notice it. 

I also haven't tried using mahony's at all yet, so there is more to try.

I also am curious to try some logic that will detect periods of calm/no movement and use those to do quick calibrations. But it works pretty dang good right now, so I'm not too concerned. I can feel little bits of nose drift on turns on even my most stable commercial wheels sometimes, and older commercial wheels were terrible.

vedderb:
I thought about it, and it makes sense why what you describe happens. When you are turning you are also leaning in the direction you are turning, meaning that the yaw rate will be coupled to the pitch rate proportional to the sine of the roll angle. Therefore the assumption I made with the calibration, namely that roll and pitch are 0 on average, is not true for pitch while turning. That is why the pitch offset will slowly drift when you turn. The issue with having high beta and no gyro offset is a different thing though.

Assuming:
Roll is the axis that goes sideways (not the balancing one) with index _r_ind_
Pitch is the balancing axis with index _p_ind_
Yaw is the turning axis with index _y_ind_

Can you try the following:

split this line to three lines, one for each gyro axis

```
utils_step_towards(&m_gyro_offset[i], m_gyro[i], m_settings.gyro_offset_comp_fact[i] * dt);
```

and do the following to the pitch axis

```
// Calculate the coupling (maybe needs a flipped sign)
float p_coupling = sin(ahrs_get_roll(&m_att)) * m_gyro[y_ind];

utils_step_towards(&m_gyro_offset[p_ind], m_gyro[p_ind] - p_coupling, m_settings.gyro_offset_comp_fact[p_ind] * dt)
```

don't forget to flip the sign as the comment suggests if it gets worse.

with a bit of luck that should decouple the yaw and pitch axes when you are leaning and turning at the same time. I don't know if this works or if the coupling is more complex than that, but I have nothing to test it on at the moment. I did buy and old EUC from a friend who had one though, and it will hopefully arrive tomorrow. Only problem is that I probably am the problem if it doesn't work, as I have no idea how to ride it :-)

Mitchlol:
Ok, a few things..

1) I made the requested changes
2) I found what i thiiiiink is a bug in the gyro code, but i don't really understand timer.c/h because it looks to be native CPU code (I don't even know)? But anyways in imu.c the first 3 lines of imu_read_callback() seems to be always setting last_time to 0 so dt is always 0 which will throw off the ahrs?
3) Aside from balance stability issues, which at this point is working pretty great, there is one other issue. When changing directions under high load/ at high speed the motor will stutter for a tiny bit. Changing directions is more of an advanced maneuver and many/most riders can't even do it. I however do it quite a lot :joy_cat:. Anyways i recorded some footage of this happening, it feels worse than it looks. If i change directions really slow and carefully, i can avoid the twitch, but on a realistic direction change it pretty annoying. I don't really think it is a PID issue, but I am open to being wrong, I can also provide graphs or whatever of this happening. Video: https://www.instagram.com/p/B2TU6PnAd0I/
4) I looked at your recommended comp changes, and I think I get how to implement them, its a little extra confusing because i have some mounting orientation changes, and I'm not balancing on pitch/roll, I also work 2 jobs on Wednesdays, and the sun was in my eyes, and etc etc... Ill try to test it soon, I may hide the laptop from myself till the weekend for some forced relaxing time.

vedderb:
I have merged now, and will give it a try as I got my EUC :-)

I was wondering, is there really a need to choose a gyro orientation? You set the IMU rotation in the IMU app so that the coordinate system always is aligned with the vehicle. Then you can use the pitch axis for balancing regardless of how the VESC is mounted, which I think is more natural.

Mitchlol:
So I thought that would be the case initially, but then i tried doing the rotation required for me and it didn't seem to work. Ill try to mess with mine again and see if i can get it such that pitch is balance. Once i rotated far enough that the axes names got crossed, my brain started to hurt.

Also there is one axis that wont rotate a full 360, just to 180 and back which is weird :thinking: 

Thanks for the merge! I will throw a party :confetti_ball:

vedderb:
Rotations are always tricky, I still don't have a full intuition for what the math does to the physical thing in each operation, especially not quaternions.

The idea with the rotations is that you first adjust roll and pitch so that the board is level, then you adjust yaw so that the roll and pitch axes point in the correct directions. Keep in mind that they can point backwards too, which is not too obvious in the graph; then you you have to add + or - 180 degrees to the yaw. I will add a 3D visualization to the plot so that it becomes easier to see what is going on - just haven't decided if I will embed QML or use a regular QOpenGlWidget.

gihad:
There are quite a few DIY Onewheel like projects coming up. Do you guys think this addition will help with those as well?

PR #113, opened 2019-12-07
Formatting style and tachometer interface

ddosoff:
I did not found anything about code formatting style. After some research I found you use Microsoft style. Just added as default option for vscode into project folder.

I am writing winching application and do some control loop to rewind motor back to tachometer zero position. I need to slightly modify zero position sometimes. That is why I added:

`int mc_interface_set_tachometer_value(int steps);`

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,1 @@
+BasedOnStyle: Microsoft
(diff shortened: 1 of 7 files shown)


ddosoff:
I found mistake in the .clang-format

PR #115, opened 2019-12-07
Set tachometer value interface

ddosoff:
Hello Benjamin,

I use VESC for winching application - https://www.youtube.com/watch?v=KoNegc4SzxY

I'd like to move my zero point with terminal command in my custom app. That's why I added interface for setting tachometer value.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,1 +1,0 @@
-BasedOnStyle: Microsoft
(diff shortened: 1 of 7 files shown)


ddosoff:
Oh sorry, still fighting with git.

PR #116, opened 2019-12-07
Set tachometer value interface

ddosoff:
Hello Benjamin,

I use VESC for winching application - https://www.youtube.com/watch?v=KoNegc4SzxY

I'd like to move my zero point with terminal command in my custom app. That's why I added interface for setting tachometer value.

### DIFF
--- a/applications/app_skypuff.c
+++ b/applications/app_skypuff.c
@@ -1,0 +1,582 @@
+/*
+	Copyright 2019 Kirill Kostiuchenko	kisel2626@gmail.com
+
+	This file is part of the VESC firmware.
+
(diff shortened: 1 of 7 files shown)


vedderb:
Interesting application, I have to give it a try :-)
Also nice with a solid example of a custom application. Thanks for the PR!

ddosoff:
Benjamin, sorry I am newbie on github.

Now I understand that I have to create different branches for set
tachometer interface and app_skypuff, because app_skypuff is not finished
yet.

vedderb:
As long as app_skypuff isn't in conf_general.h it won't be used, so having it there is an example of work in progress is not a problem.

PR #120, opened 2019-12-26
KTY83/122 and priority to custom commands

ddosoff:
I have QS-Motor 4000w with KTY83/122 motor temp sensor.

It is impossible to calculate PTC sensor temperature accurately with coefficient of PTC thermistor only.

I used method described here https://vasilisks.wordpress.com/2017/12/14/getting-temperature-from-ntc-kty83-kty84-on-mcu/#more-645 to approximate datasheet temperature curve into small polynom.

Then I used integer only calculations (without FPU) to calculate temperature from ADC value.

I created polynom for KTY83/122 sensor here: https://docs.google.com/spreadsheets/d/1iJA66biczfaXRNClSsrVF9RJuSAKoDG-bnRZFMOcuwU/edit?usp=sharing

I checked, this method works well for my motor.

This method is better for PTC thermistors.

----------------------------------------------------------------

Custom applications could rely on high rate of terminal commands between UI and MCU.

There is CPU intensive copying https://github.com/vedderb/bldc/blob/master/terminal.c#L78
and many strcmp in the terminal.c.

I moved execution custom command before these steps.


vedderb:
This changes the 1k@100 PTC to a 1.01k@25 PTC, which is not consistent with the config option. Also, it does not allow hardware to use pullups other than 10k. The STM32F4 has an FPU, so adding the bitshifts and integer calculations most likely makes it slower and less readable.

If accuracy of the KTY83 away from 100 degC is important I would suggest making the polynomial fit based in the measured resistance and not the ADC value (for example with NTC_RES_MOTOR). Also, I would prefer using floats.

ddosoff:
Yep, using resistance as polynom parameter is better. Will change my PR.

Thanks.

ddosoff:
I changed polynom to resistance as input parameter and incremented firmware version.

PR #128, opened 2020-01-24
Modifying encoder spi_transfer to sample on falling edge

TobinHall:
I was having issues with encoder errors. After looking into the AS5047x manual, the data is meant to be sampled on the falling edge of CLK. This is not consistent with the current implementation which samples one spi_delay after the rising edge.

When moving the sampling to after the falling edge, I have found the encoder to be much more reliable.
It also has the advantage of compressing all of the delay into one place. I have increased the delay to bring the duty cycle of the clock closer to 50%. 
This correspond to approximately 48% duty cycle at 3.8MHz (compared to 77% at 4.0MHz previously).

I have not tested this with the AD2S1205 which also uses this function, though the datasheet for this device also specifies falling edge sampling.

### DIFF
--- a/encoder.c
+++ b/encoder.c
@@ -606,6 +606,7 @@ static void spi_transfer(uint16_t *in_buf, const uint16_t *out_buf, int length) 
 
 			palSetPad(SPI_SW_SCK_GPIO, SPI_SW_SCK_PIN);
 			spi_delay();
+			palClearPad(SPI_SW_SCK_GPIO, SPI_SW_SCK_PIN);
 
 			int samples = 0;
 			samples += palReadPad(SPI_SW_MISO_GPIO, SPI_SW_MISO_PIN);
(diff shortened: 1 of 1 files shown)


nitrousnrg:
I think the code was written that way so it works with data clocked out on rising edge and also be compatible with falling edge data signals.

From the user perspective, they don't know about clock phase or polarity, they just choose the encoder type from a droplist, so there is no real need to keep it flexible. If we need to support different phase we can write custom low level SPI code for each new encoder protocol.

Besides making it more reliable, it can make vesc faster because the SPI bitbanging makes heavy CPU use. Latest firmware version can't run at 30kHz control loop freq without the OS crashing on my boards, so make sure the overall SPI transaction is not longer than before to avoid impacting the max FOC frequency.

Also for example the AD2S1205 could be run at much faster CLK  speeds as the physical trace lengths are minimal comparing to the AS5047 signals that have to reach all the way into the motor, so flexible SPI config in end would improve the FOC control loop frequency on my builds if it can run at 20MHz SPI clk.

I'll stick around and test the code for AD2S1205 if you guys make any change.

TobinHall:
> Besides making it more reliable, it can make vesc faster because the SPI bitbanging makes heavy CPU use. Latest firmware version can't run at 30kHz control loop freq without the OS crashing on my boards, so make sure the overall SPI transaction is not longer than before to avoid impacting the max FOC frequency.

We can bring the frequency back up, my bench testing seems to work (not exceed the 1% error rate) at 5.2MHz.
Is there a reason why hardware SPI isn't used? The AS5047_USE_HW_SPI_PINS still does bitbanging. 



>From the user perspective, they don't know about clock phase or polarity, they just choose the encoder type from a droplist, so there is no real need to keep it flexible. If we need to support different phase we can write custom low level SPI code for each new encoder protocol.

I agree. Shouldn't falling edge be the default given the two encoders that are currently implemented?

nitrousnrg:
Bitbanging is used for flexibility on the pinout.
I really wanted to use the SPI peripheral but the couple of times I tried I couldn't make it work. With ChibiOS when you enable the spi peripheral it sets up a DMA abstraction layer which I recall broke the USB, and I couldn't debug that further.

TobinHall:
> Bitbanging is used for flexibility on the pinout.
> I really wanted to use the SPI peripheral but the couple of times I tried I couldn't make it work. With ChibiOS when you enable the spi peripheral it sets up a DMA abstraction layer which I recall broke the USB, and I couldn't debug that further.

We could look into using hardware SPI outside of ChibiOS. I might have a look if I get some free time.

ZachOB:
I got hardware SPI working in [this PR](https://github.com/vedderb/bldc/pull/511) using ChibiOS and DMA. I was only able to test it using the 75_300 hardware but it should at least be a good example to get hardware SPI working.

PR #138, opened 2020-02-23
Balance app updates!

Mitchlol:
-Can bus support.

-Wheelie board support! Any ol skateboard board can now balance! With yaw stabilization, and using pitch to steer! Can also center mount a channel truck for a super easy center wheeled build!

-Switch updates. Uses adc so users can configure for their sensors, also allows for a single switch when one of the pins have failed (Cant say why, but I've had 2 or 3 users with a broken ADC pin).

-Adjustable holding current. (helps with hill starts)

-Fault delay. (now you can have some air time without a safety cutoff, but keep safety cutoff enabled)

-Directional nose angle adjustment. (can set higher angle for hills, or down angle for extra carving)


vedderb:
Nice! Do you have a video where someone has tried this on a skateboard?

Mitchlol:
> Nice! Do you have a video where someone has tried this on a skateboard?

https://www.instagram.com/p/B84skBUABnl/
https://www.instagram.com/p/B9D6rt2g9Lk/

This last video is NOT running this code, it is running a vesc with an external balance controller made by @blezalex who inspired many of these changes but it illustrates the general idea of what it would look like on a kick tail skateboard setup.
https://www.instagram.com/p/B8pXOBrAWAh/

vedderb:
That is really cool! Also impressive that HFI works for self balancing applications, that is the ultimate challenge for sensorless operation. I used to tell people that in most cases you don't need sensors, unless you build something that is self balancing :-)

Mitchlol:
While it does work, i wouldn't recommend HFI as a first resort. Every few seconds one of the wheels will glitch for some cycles. With 2 wheels that results in a yaw motion. It can be very scary if you're not expecting it, and an experienced rider. Im used to riding science experiments tho, and it only really happens at very low speeds. Interestingly tho, when i did first switch to HFI, aside from the random glitching, my observation at the time was that performance was overall noticeably smoother and the balancing felt more locked in.

In general I'm surprised that belt drive dual motor with a channel truck can ride at all, the lack of hall sensors is the icing on the cake! I don't think the channel truck is very happy, i have to keep tightening it, the belts get loose over time too, and when you jump the wheels spool way up and try to destroy everything when you land. 

I'd love to see what a gear drive & encoder setup feels like. I may print up some parts i can attach to my motor and motor mounts for encoders.

Regardless this is my best performing setup to date, mostly because the form factor gives me good quality parts, with a nice top speed and range. Everything is off the shelf, and the build is a breeze!

I have a few more tweaks to the code I'm still figuring out. I think a max yaw current would be nice so you can break traction if you want. The turning converts lean to degrees per time, while normal vehicles are degrees per distance, this helps with stability at speed, but it helps a little too much. I'm also working on a smooth feeling algorithm that allows the setpoint to follow the angle to some degree with delay but also bias back to center. I have something that actually works, while its good enough for a v1, I want to spend some more time playing with it.

Mitchlol:
Perhaps i can configure HFI better, i watched the video but i don't really understand much that about motors.

PR #143, opened 2020-03-21
More bbshd support

nitrousnrg:
This branch supports minor bbshd hw features and an observer monitor:

Disabled by default, the monitor compares the encoder angle with the estimated observer angle when using an encoder.
When using hall sensors, it compares the interpolated hall sensor angle with the observer angle.

The comparison is done only beyond sensorless rpm, when the observer is supposed to work reliably. It prevents reaching very large angle errors that ultimately result in an overcurrent condition.

Fault is triggered with a hardcoded 20° angle difference but the threshold could be exposed to the user if we find its an useful feature.

### DIFF
--- a/hwconf/hw_axiom.h
+++ b/hwconf/hw_axiom.h
@@ -35,6 +35,7 @@
 #define HW_HAS_3_SHUNTS
 #define HW_HAS_PHASE_SHUNTS
 #define HW_HAS_SIN_COS_ENCODER
+#define HW_HAS_GATE_DRIVER_SUPPLY_MONITOR
 
 // Macros
 #define ENABLE_GATE()			palSetPad(GPIOC, 14)
--- a/hwconf/hw_luna_bbshd.c
+++ b/hwconf/hw_luna_bbshd.c
@@ -28,6 +28,8 @@
 // Variables
 static volatile bool i2c_running = false;
 
+void hw_luna_bbshd_setup_dac(void);
+
 // I2C configuration
 static const I2CConfig i2cfg = {
 		OPMODE_I2C,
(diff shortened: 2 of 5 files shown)


mcpwm_foc.c
```diff
@@ -3305,70 +3324,67 @@ static float correct_encoder(float obs_angle, float enc_angle, float speed, floa
...
-
-		// Only override the observer if the hall sensor value is valid.
-		if (ang_hall_int < 201) {
-			static float ang_hall = 0.0;
-			float ang_hall_now = (((float)ang_hall_int / 200.0) * 360.0) * M_PI / 180.0;
+	int ang_hall_int = m_conf->foc_hall_table[read_hall()];
 
-			if (ang_hall_int_prev < 0) {
-				// Previous angle not valid
-				ang_hall_int_prev = ang_hall_int;
```
nitrousnrg: The hall logic is kept pretty much the same, the big heads up is that I deleted this line from the original code
`ang_hall_int_prev = ang_hall_int; 
`
Looks like it prevents the code to reach cases when ang_hall_int_prev is -1 or -2.

vedderb:
Can you base this on the dev branch instead, and maybe separate the bbsh features from the observer monitor?

I'm not sure how useful the observer monitor is, so I want to test it a bit separately when I have time.

nitrousnrg:
> Can you base this on the dev branch instead, and maybe separate the bbsh features from the observer monitor?

No problem, this branch now its only including bbshd code and pointed at the dev _fw_5_00 branch

About the observer monitor, I'll give it a try when I have time, it has a bunch of conflicts with the dev branch.

Issue #153, opened 2020-04-12
ADC Application Potential Issue


SJ-Innovation:
When using a potentiometer as an ADC application input source, there is no "sanity checking" on the input values. This poses a potentially significant safety issue in the case that a wire becomes disconnected from the potentiometer. 

For example, if a ground wire becomes disconnected, the ADC input will immediately shoot up to the V+ rail, and stay there. The ADC application will map and truncate this to full power. Effectively the motor will have full power applied regardless of the real throttle setting. This is bad and I found out first hand... 

In this case it would make sense to add some sanity checks based on the ADC voltage min and ADC voltage max settings. If the ADC voltage strays out of the prescribed range, a hardware error is raised and the motor is put into emergency (timeout) braking mode.

I'll start a PR to do this.

Jfriesen222:
Often times there is noise present from the adc throttle input wires running close to the motor wires in e-scooters etc. Be mindful that any kind of thing that cuts power when out of range will also cut power on a noisey throttle input and might cause new issues. Maybe some type of counter when it is out of range continuously for a certain duration would be ok. 

My best advice would be to not use ADC throttle whenever possible as it is so vulnerable to corruption.

PR #154, opened 2020-04-12
Add ADC range sanity checking

SJ-Innovation:
Addresses #153 

Adds sanity checking to the ADC range just after a read, if the voltage is out of range, timeout brake current is applied and the rest of the ADC app is bypassed. Only once both sensors are within safe ranges can the motor be driven as per the throttle and brake settings.

### DIFF
--- a/applications/app_adc.c
+++ b/applications/app_adc.c
@@ -84,6 +84,15 @@ float app_adc_get_voltage2(void) {
 	return read_voltage2;
 }
 
+static bool adc_sanity_check(float check, float min, float max) {
+    // If we are outside of the given min max range.
+    if ((check < min) || (check > max)) {
+        //  Use timeout braking current instead of max brake.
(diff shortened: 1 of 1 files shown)


vedderb:
Have you tested this? If you are on full throttle and get a tiny bit of noise, wouldn't it trigger the brake randomly then? For it to work you probably have to have some margin and a counter to only trigger when the throttle has been outside of the range for long enough. Also, if you use a potentiometer instead of a hall sensor throttle, you essentially always use the full range. Further, calling return from the ADC thread without properly updating the app state is not a good idea either.

I think a better solution is to have a pull down on the throttle close to the PCB. Unfortunately the STM32F4 does not allow to use the built in pull down pins while the pins are in ADC mode, but I thought about altering between ADC and GPIO with pull down to discharge the input capacitance in case of a connection break and read low values then, but I haven't tested it yet and don't know how well that will work.

PR #155, opened 2020-04-14
Dev fw 5 00

Jfriesen222:
Should provide a slightly more accurate estimate of the remaining battery, I fit a fifth order polynomial to a discharge graph of a samsung 30q cell. I fit the graph with 3.2V (2.57 Ah) as the 0% to match with the existing method. 

Technically the watt hours left should be integrated across this curve instead of using the average remaining voltage but maybe the combination of both provides a good midway estimate between the actual curve and a line since all cells are slightly different.

image

image

### DIFF
--- a/conf_general.h
+++ b/conf_general.h
@@ -24,7 +24,7 @@
 #define FW_VERSION_MAJOR			5
 #define FW_VERSION_MINOR			00
 // Set to 0 for building a release and iterate during beta test builds
-#define FW_TEST_VERSION_NUMBER		3
+#define FW_TEST_VERSION_NUMBER		4
 
 #include "datatypes.h"
 
--- a/conf_general.h
+++ b/conf_general.h
@@ -72,8 +72,8 @@
 // Mark3 version of HW60 with power switch and separate NRF UART.
 //#define HW60_IS_MK3
 
-#define HW_SOURCE "hw_60.c"
-#define HW_HEADER "hw_60.h"
+//#define HW_SOURCE "hw_60.c"
+//#define HW_HEADER "hw_60.h"
 
 //#define HW_SOURCE "hw_r2.c"
 //#define HW_HEADER "hw_r2.h"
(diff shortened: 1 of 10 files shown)


Peemouse:
Really nice !
How about storing the wh_left in the EEPROM ?
This could be achieved without EEPROM stress at least on HW versions with powerswitch, couldn't it ?

Jfriesen222:
Right now this is still just a more accurate mapping of a voltage reading so no memory needed. If we did get that working then we could consider doing coulomb counting which would be really nice.

Peemouse:
Oh sure, this PR is still really great !
I'm using a DieBieMS on a board which use coulomn counting and I have to admit that the SoC is amazingly accurate !
Adding it to the VESC would be an awesome feature (it can be debated though as this is more related to the battery management).

ddosoff:
Bat SOC curve have to respect battery chemistry type..

Jfriesen222:
This already does rely on a switch for 4.2 v lithium batteries, overlaying LiPo and lithium ion they aren't terribly different.  It wouldn't be terribly hard to switch different chemistries but at certain point I think you just count coulombs instead. This can be a good way to get an absolute reference though before you start counting.

vedderb:
Is this mapping for watt hours, or for amp hours? It should be for amp hours for the rest of the equations to work. However, if the voltage is non-linear, the assumption I make that the average voltage left will be the average of the voltage now and minimum voltage won't be true any more, which means that the average voltage left also needs a new calculation to get the remaining capacity correct (as I display it in wh and not ah, since that makes more sense). Even if that is done though, the largest source of error for the range estimation is that the Ah amount that people enter only is true if the battery is drained to below 3V, which won't happen with the 3.2V limit. This makes me feel that there are several loose ends here still, and I almost prefer the old liner mapping as that is less confusing, especially for FW5.0.

In the future we can try to improve the battery SOC estimation, but for making it as good as possible I think a routine where you leave the motor running overnight and let VESC Tool look at the coulomb counting (ah) as well as wh counting, and map it to the voltage to get an individual graph for each battery. A report could be presented about the battery pack after that, and this over night procedure could be repeated in the future to get an accurate estimation of how the battery has degraded over time. This is too much for FW5 though, so for now I think I prefer the old mapping method. Maybe I missed something with this PR though, and you can still convince me to include it :-)

PR #158, opened 2020-04-19
IMU Calibration Thingy

Mitchlol:
Properly zeroing out an IMU is too hard to calculate for an average guy like me due to the sequential nature of rotation matrices. In addition accel/gyro offsets need to be recalculated for every change to the orientation. So I created a tool to do it for me (and everyone else too!)

Things to note:

- Its ugly/in a weird place in the UI
- Does this command need to be rewritten as an aync thing?
- This doesn't do any accel offset calculation, I left all the fields there for the future, and a comment explaining why I didn't implement it.


vedderb:
Nice, thanks! Just one note: When adding enums to COMM_PACKET_ID, always add them to the end. This way compatibility with old versions won't be broken.

vedderb:
I see now that the datatypes in VESC Tool weren't updated, maybe it was confusing because of that. Never mind :-)

Mitchlol:
Yeah, i added it there first.

vedderb:
Yes, about the async thing, this command probably should be async as it takes some time to complete. There you can do e.g. RT data streaming at the same time and see what is going on. I will make it async commit that.

Mitchlol:
Sweet something else i also thought of is the mobile app? Not sure how this will show up there.

Personally I manually calibrated my imu over a week ago and it still seems pretty dialed. But it would be nice to be able to do it on the go.

vedderb:
It should be quite easy to add it to the mobile app. Have a look at ConfigPageApp.qml and PpmMap.qml. You would probably make something like PpmMap.qml and add it to ConfigPageApp.qml as a dialog that is opened from the menu. Uncomment CONFIG += build_mobile in vesc_tool.pro to build the Qml gui. If you are on linux you can connect over BLE just as you do on the phone and test it out.

Mitchlol:
I am on linux, ill give it a go.

I can wire up two of my cheap testing vescs so one has bluetooth and one has imu, and then can bus over, so i don't need clear the settings on my running vehicle lol!

vedderb:
You could just make a backup of the settings by saving the xml files

vedderb:
Btw, it would be nice to have some tutorial on how to set up the balance app. There are quite a lot of settings. Would you be interested in writing a documentation page here? https://vesc-project.com/documentation

Mitchlol:
Sure, I recorded that first video a long time ago, and have been meaning to do an update for a while. But I was actually holding off till I had a better handle on the IMU stuff, as that is something that confuses the most users. 

So yes, its what I was going to do next anyways. My plan was to make a series of short videos for each section, but I can do a text write up as well, perhaps combine the two. It would be great to have something on the official site, balance users are starting to pop up faster than i can keep track of them :joy: 

^^I'm concerned about how the xml will hold up to a newer beta build, i guess i could re-flash the old build and load the xml when I'm done. Maybe i will, because i think i need to make some cables for the other setup, i just don't like messing with working things as a side effect. Either way ill figure something out.

vedderb:
A combination would probably be best. You don't need to write a lot, just listing all videos is also better than nothing. Do you have a user name on vesc-project.com? I can give your user permission to create and edit documentation pages.

Mitchlol:
NuRxG

PR #159, opened 2020-04-23
Adjustment to smart reverse.

Jfriesen222:
If already rolling backward switch back into duty cycle mode rather than braking.  In my opinion this feels more intuitive and can still use forward throttle to "brake" from reverse speed. 

If going in reverse above max duty then it will ramp down at the same acceleration rate as it ramps up in smart reverse.

[Idea from forum via ducktaper](https://forum.esk8.news/t/vesc-fw-5-0-beta-testers-wanted/23342/325?u=deodand)

### DIFF
--- a/applications/app_nunchuk.c
+++ b/applications/app_nunchuk.c
@@ -413,10 +413,16 @@ static THD_FUNCTION(output_thread, arg) {
 				duty_control = true;
 			}
 
+			if((!was_duty_control) && (rpm_lowest < -500) && (out_val < -0.1)){
+				duty_control = true;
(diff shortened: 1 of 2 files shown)


vedderb:
Not sure I like this. Imagine jumping and turning the board 180 degrees around, and then slightly pressing reverse - it will throw you off the board as you are going fast suddenly set a really low duty cycle. What I would do instead of checking the ERPM is checking the duty cycle. If it is more than 10% of the smart reverse duty and less than the maximum go to duty mode, otherwise go to break mode as before.

Issue #160, opened 2020-04-25
Apply/read profile without the need of reading MCCONF


Peemouse:
Hi Ben,

This is less an issue than exchanging thoughts.

I'm trying to play with profiles apply/read on small MCU such as Arduino.
Those MCU are very limited in term of memory and reading the whole MCCONF is quite difficult.

However, if I want to apply or read the current profile applied, I do need some values of MCCONF packet.

- **Applying a profile**

COMM_SET_MCCONF_TEMP requires some values such as duty_min or max and others that can be known only if MCCONF is retrieved.

1. What about a "sanity" check that would skip those values if the deserialization of them shows 0.0 ?
For instance :
```
float32 read_min_duty = buffer_get_float32_auto(data, &ind);
if (read_min_duty > 0.0) { mcconf.l_min_duty = read_min_duty }
```
That way I can just send a profile with the values to change and send 0.0 for those I don't want to change ?

This isn't elegant (like could be a mask) but at least it keeps backwards compatibility.

- **Reading the current profile**

This would require an extra command I guess for retrieving needed values.
Something like :
```
case COMM_GET_MCCONF_TEMP : {
  mcconf = *mc_interface_get_configuration();
  int32_t ind = 0;

  buffer_append_float32_auto(buffer, conf->l_current_max_scale, &ind);
  buffer_append_float32_auto(buffer, conf->l_current_min_scale, &ind);
  buffer_append_float32_auto(buffer, conf->l_min_erpm, &ind);
  buffer_append_float32_auto(buffer, conf->l_max_erpm, &ind);
  buffer_append_float32_auto(buffer, conf->l_min_duty, &ind);
  buffer_append_float32_auto(buffer, conf->l_max_duty, &ind);
  buffer_append_float32_auto(buffer, conf->l_watt_min, &ind);
  buffer_append_float32_auto(buffer, conf->l_watt_max, &ind);
  buffer_append_float32_auto(buffer, conf->l_in_current_min, &ind);
  buffer_append_float32_auto(buffer, conf->l_in_current_max, &ind);
  //needed for speed calculation
  buffer[ind++] = (uint8_t)conf->si_motor_poles;
  buffer_append_float32_auto(buffer, conf->si_gear_ratio, &ind);
  buffer_append_float32_auto(buffer, conf->si_wheel_diameter, &ind);

} break;
```
  
[...]

PR #163, opened 2020-05-01
Add COMM_GET_MCCONF_TEMP command

Peemouse:
Related to issue #160


vedderb:
Looks nice, thanks!

vedderb:
Just one comment for next time: it is better to not include the binary files in the PR. Updating them too often makes the history grow in the repository, so I would like to minimize that. Running the build script to see if everything is ok after making changes is good practice though.

Peemouse:
TBH. I did hesitate on that. Good to mention it.
Could you add this to the CONTEIBUTING file ?
BTW the code is tested on HW 4.12
I always make run my code on a real VESC to make sue everything is ok.
Thanks for merging.

PR #164, opened 2020-05-09
Fix fet temperature in UAVCAN mode is not consistent with VESC

jxltom:
In vesc, the fet temperature as ```T FET``` is using ```mc_interface_temp_fet_filtered * 1e1``` directly but uavcan is using ```mc_interface_temp_fet_filtered + 273.15```

It looks like ```273.15``` is additional, this PR fixes this. Let me know if I'm wrong, thanks.


pavel-kirienko:
The original code is correct and does not need fixing. It is a conversion to kelvin.

jxltom:
So VESC is using °C and UAVCAN is using °F?

pavel-kirienko:
UAVCAN uses SI everywhere. The SI unit for temperature is kelvin.

jxltom:
Got it, thanks for the clarification!

PR #166, opened 2020-05-10
Add ramp input to the speed PID loop.

maxicor88:
This feature works together with a pull request in Vesc Tool that adds 2 control inputs in the config file.

The goal of this ramp is to prevent large steps commands being directly fed into the PID loop. 

In our tests this ramp was extremely useful to improve the startup of the motor in Speed Control Mode. We could be able for better tune the PID, because before this, we needed to apply very small proportional and integral gains to the loop to get a smooth startup, but with worse results in stable mode.

We believe this is also useful for safety reasons, preventing that for any reason a wrong speed command is entered, causing the PID loop to inmediately try to address that command.

We added a control to disable this in Vesc Tool, so user can decide to use the ramp, or continue with previous code.

Hope you find this useful, it was day and night for us in our Dyno tests with large loads, but we also tested this with a 4.10 hw and it worked better also.

This works together with a new config file of Vesc Tool, that adds 2 variables to the Speed Pid page.

### DIFF
--- a/confgenerator.c
+++ b/confgenerator.c
@@ -118,6 +118,8 @@ int32_t confgenerator_serialize_mcconf(uint8_t *buffer, const mc_configuration *
 	buffer_append_float32_auto(buffer, conf->s_pid_kd_filter, &ind);
 	buffer_append_float32_auto(buffer, conf->s_pid_min_erpm, &ind);
 	buffer[ind++] = conf->s_pid_allow_braking;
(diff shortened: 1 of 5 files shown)


vedderb:
I think ERPM per second is better than ERPM per MS, as that is what I used in the apps. Also, instead of having an activation flag, a negative value could mean that the no ramping is used. Then the default value can be -1. Can you update the PR?

maxicor88:
I tried to implement the changes you propose in this last commit.

PR #167, opened 2020-05-11
CI

varnerac:
Adds continuous integration via GitHub Actions to build firmware for different versions. The `build_all` results are uploaded as artifacts. Right now I have it run on every push. Based on the release strategy, you could change it to PRs, etc. I am hoping this makes it easier for folks to test the 5.x branch without having to install an ARM build chain locally.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,27 @@
+name: Build BLDC
+on:
+  push
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v1.0.0
+      - name: install ARM embedded
+        run: |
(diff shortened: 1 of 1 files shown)


varnerac:
As a side note, I can also have it automatic upload release binaries to the GitHub release page after you tag a release. That way you could have `5.0.0-rc1`, `5.0.0-rc2`, etc.

Teslafly:
This should be re-targeted to the dev_fw_5_03 branch.

@vedderb, is there any reason you don't want CI on vesc-fw/vesc-tool? There are open pull requests on both to add this.

Also, as there have been some issues with newer versions on the arm toolchain, this doesn't look like it currently locks that down.   See https://github.com/vedderb/bldc/issues/210.

I would love automatic release candidates though. Especially for vesc tool.

rusefillc:
@vedderb what are your thoughts on this PR?

PR #171, opened 2020-05-16
Axiom: adds input current sensor capability.

maxicor88:
use HW_HAS_INPUT_CURRENT_SENSOR in hw_axiom.h to enable it, it is commented by default

We added the possibility of using an external DC current sensor for measuring the input current, instead of calculating it indirectly. Our goal is to achieve better global system efficiency estimation.

### DIFF
--- a/mcpwm_foc.c
+++ b/mcpwm_foc.c
@@ -193,6 +193,7 @@ static float correct_encoder(float obs_angle, float enc_angle, float speed, floa
 static float correct_hall(float angle, float dt, volatile motor_all_state_t *motor);
 static void terminal_plot_hfi(int argc, const char **argv);
 static void timer_update(volatile motor_all_state_t *motor, float dt);
+static void input_current_offset_measurement( void );
 static void hfi_update(volatile motor_all_state_t *motor);
 
 // Threads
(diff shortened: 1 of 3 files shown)


mcpwm_foc.c
```diff
@@ -2570,6 +2571,9 @@ void mcpwm_foc_adc_int_handler(void *p, uint32_t flags) {
 		motor_now->m_motor_state.iq = 0.0;
 		motor_now->m_motor_state.id_filter = 0.0;
 		motor_now->m_motor_state.iq_filter = 0.0;
+	#ifdef HW_HAS_INPUT_CURRENT_SENSOR
+		GET_INPUT_CURRENT_OFFSET();
```
vedderb, line 2575: Should this really be done here? Just when the motor is released there might still be some current flowing through the body diodes due to the inductance, and that will offset this measurement. I think it is better to only sample the offset at boot.

maxicor88: Hello Benjamin, Thanks for considering our pull request. 
That code only executes if a trigger was generated.  That trigger is generated inside this function **input_current_offset_measurement()**. 
The code as it is now, waits for a second after boot, and then takes some samples, and average them to get a better offset measurement. We did this because we got quite bad results if we took the very first samples of the ADC right from the beginning, so we used a delay of 1 second. It may be also possible to trigger offset measurements later, by calling the MEASURE_INPUT_CURRENT_OFFSET() macro, but now we only found it necessary at this point.

PR #177, opened 2020-06-04
Add open source HW project, app for diving scooter, drv8301 set_gate_current

CTSchorsch:
Hi Benjamin,

if you want, please feel free to add my open source hw design found on https://github.com/CTSChorsch/gsvesc in your firmware repo. It is based on your 75V design, but, however, the design has some special shape to fit in my diving scooter tube. If it is too special for you, just close this pull request. In addition i add a custom app to read out three hall sensors, one for trigger, two for a simple incremental encoder (using interrupt)for rpm adjust. I add a set_gate_current function to the drv8301.*, maybe it is usefull for some designs using the drv. i don't like this device anymore and use  single gate drivers.

I'm new to git, i expect some discrepancy to your rules, but i tried to comply with your requirements

And once again, if the hardware design is to special and at the moment with only one user, kill this request :)

anyway,thanks for this project and your good job
regards from germany
Georg


Mitchlol:
Are you planning to produce & sell this controller?

CTSchorsch:
Hi,

i did a run @jlcpcb with 5pc of a previous version. They were  some bugs, but for me it is ok. So, at the Moment there are no plans to produce more. 

I fixed the bugs in schematic and layout. You can make some on your own.  But please check the schematic. Maybe some bugs are left :). No guarantee from my site.  Most Parts are available at jlc and they did a great assemble Job. Only Connectors and BLE have to solder by yourself. 

Georg

PR #178, opened 2020-06-06
Replace call to qsort with insertionsort

Gamer43:
Small change, probably negligible impact. Should save a few clock cycles :).

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -812,15 +812,44 @@ static int uint16_cmp_func (const void *a, const void *b) {
 	return (*(uint16_t*)a - *(uint16_t*)b);
 }
 
+//possibly replace with a four pole butterworth filter from https://www-users.cs.york.ac.uk/~fisher/mkfilter/trad.html
+//perhaps also choose filter lengths that are powers of 2 to optimize modulo to bitwise AND.
+//this can be forced using enumerations.
 uint16_t utils_median_filter_uint16_run(uint16_t *buffer,
-		unsigned int *buffer_index, unsigned int filter_len, uint16_t sample) {
-	buffer[*buffer_index++] = sample;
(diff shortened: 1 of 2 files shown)


kalvdans:
While you are on to optimizing it, the median can be found in O(n), no need to sort the array.

Gamer43:
> 
> 
> While you are on to optimizing it, the median can be found in O(n), no need to sort the array.

While such algorithms (such as quickselect) have a linear asymptotic running time, they are recursive and have unacceptable overhead computation. As such they are unsuitable for this application. Perhaps if the need were to find the median in one thousand or one million elements, then an alternative algorithm would be more suitable. But, in this application, the number of elements does not exceed 100.

kalvdans:
The call overhead depends on the compiler. Measure, don't assume :)

Gamer43:
> The call overhead depends on the compiler. Tail calls gets converted to a loop. Measure, don't assume :)

You would have to implement that algorithm manually to ensure it uses tail calls and optimizes into an iterative loop. I am also referring to the overhead coefficients that are not present in the asymptotic notation. There is a reason why Insertion Sort beats every other sorting algorithm with arrays smaller than 40 elements. 
If median of medians or quickselect were actually viable algorithms, then we'd be using them for pivot selection in quicksort, except we don't because the overhead terms causes the average and even best case running times to be significantly worse.
Not only that, this is an embedded system that requires minimal resource usage overhead in every respect. Insertion Sort is the lowest overhead implementation of this.

Teslafly:
Is this still relevant?

Gamer43:
> Is this still relevant?

Codebase appears to have diverged quite significantly since opening, and I think any possible benefit would be minimal. 

I will be closing this pull request as it is no longer relevant.

PR #183, opened 2020-06-13
Add option to disable HFI when stopped in duty cycle and speed modes

lgeek:
This introduces the HFI "Min ERPM with no input" option. It dynamically
disables HFI in the duty cycle or speed modes when 0 duty cycle /
speed is commanded and the ERPM of the motor is lower than the set
value.

A value of 0 will behave the same as before, while a value of typically
a few hundred ERPM will cause HFI to turn off when the motor is stopped.

The idea is to keep HFI enabled as long as the motor is moving even if
the input is 0, to allow it to respond immediately to a different input,
for example to accelerate an inertial load which was braking down but
still in motion.

On the other hand, just enabling and disabling HFI at 0 speed can
induce some movement, so the threshold can't be 0 ERPM - turning off
HFI would make the motor move slightly, which would turn on HFI, which
would move the motor again and so on. The minimum ERPM threshold which
works as intended depends on the HFI configuration and the motor.
Value around 600 seem to be a good starting point for outrunners.


Teslafly:
probably want to remove the fw version change, but i see no inherent problem with this?

PR #185, opened 2020-06-19
zesc v6c and v7h

discoverzesc:
Hello! This is ZESC. I would like to add my hardware v6c and v7h in to it. Thanks. 
I also modified mcconf_default file. But I don't know what is the proper way to merge this file.


TechAUmNu:
You should really clean up your config files, you still have `#ifdef HW60_IS_MK3` all over the place.

PR #186, opened 2020-06-20
Update virtual motor code to work with latest Vesc v5.

maxicor88:
We implemented some updates to the original virtual motor library to make it more practical to work with.

The "connect_virtual_motor" terminal command only requires 3 arguments now, load torque, load inertia and battery voltage. 

connect_virtual_motor [load_torque in Nm] [load_inertia in Nm /s^2] [battery voltage in Volts]

image

The rest of the motor characteristics, are taken from already existant variables in VESC motor configuration.

We solved also some other issues with encoder sensor, fault gate driver undervoltage (in Axiom boards) and we improved hardware dependency defines in virtual motor.c

Something interesting to point out is that this code can also be run from a stm32f4 discovery board, or any board with same MCU. So you can play with VESC code, without having a VESC board yet. For instance, just testing this pull request I was able to find a tiny bug in speed ramp code (I will make another pull request with that).

The next snapshots are taken running from a STM32F407G-DISC1 and using latest vesc tool 2.07, showing some oscillations in motor startup.

image

image

image

I hope you find this useful, some guys were asking this update in the forum, but I could not find time to do it previously.

### DIFF
--- a/mcpwm_foc.c
+++ b/mcpwm_foc.c
@@ -529,7 +529,7 @@ void mcpwm_foc_init(volatile mc_configuration *conf_m1, volatile mc_configuratio
 	update_hfi_samples(m_motor_2.m_conf->foc_hfi_samples, &m_motor_2);
 #endif
 
-	virtual_motor_init();
+	virtual_motor_init(conf_m1);
 
 	TIM_DeInit(TIM1);
 	TIM_DeInit(TIM2);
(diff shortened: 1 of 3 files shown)


vedderb:
Nice, thanks!

I actually used the virtual motor code to make a stand alone tool in Qt that acts like a VESC connected over TCP. It is great for debugging things and experimenting with e.g. the observer (which has the same implementation in that tool as on the VESC). I also used this tool to experiment with and updating the flux linkage measurement code, and getting the axis decoupling right. When I have time to clean it up I will publish that tool too.

maxicor88:
Thanks Benjamin, 

One thing we could use this virtual motor for, is to add some experiment new tab in the GUI, that allows to input load changes on the fly, so you can test how your PID settings or any other settings are working against a load sudden change or gradual change.

I tested already that if you send the connect virtual motor command many times, with different values of load torque or inertia or voltage, you can test how this influences the control behaviour.

Another idea as an improvement is to add the Voltage Input series resistance, so we can model how an increment in battery current generates a voltage drop at the input. This could be useful for testing modulation limited control algorithm such as... field weakening.

PR #188, opened 2020-06-29
Added BMI160 SPI support

EsbenFR:
Added basic SPI support for BMI160

TODO:
- Add support for shared SPI between different devices (mainly DRV SPI)
- Add support for external BMI160 SPI (using comm pins)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,6 +155,7 @@ CSRC = $(STARTUPSRC) \
        confgenerator.c \
        timer.c \
        i2c_bb.c \
+       spi_bb.c \
        virtual_motor.c \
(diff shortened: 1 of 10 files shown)


vedderb:
Nice, thanks!

I moved the PR to the dev branch, and there are some conflicts now. Also, it looks like some auto formatting tool has modified imu.c in many places, and it would be nice to revert that to only reflect the changes in behavior from the PR if possible. I inspect all pull requests carefully, and only seeing the differences that change the behavior makes it easier. For example, the rotation matrix is styled quite differently now, and it is hard to spot if the code changed or if the lines just have been broken up.

PR #189, opened 2020-07-01
App PPM rework

Peemouse:
- **CHANGED** : Master keeps the slave(s) running even if a local fault occurs
- **CHANGED** : Master now sends relative value instead (%) instead of absolute value (A) (great for multiple VESC with different currents settings)
- **FIXED** : Timeout current brake feature not working
- **FIXED** : brake command sent over CANbus even if Multiple VESC over CAN disabled
- **FIXED** : delay before switching slave off in case of RX timeout
- **ADDED** : Traction control auto-disabling in case of fault on master, re-enable only if no effect on motors will happen (avoid weird behaviour exiting fault state on master).

### DIFF
--- a/applications/app_ppm.c
+++ b/applications/app_ppm.c
@@ -405,7 +405,7 @@ static THD_FUNCTION(ppm_thread, arg) {
 		}
 
 		if ((send_current || send_duty) && config.multi_esc) {
-			float current_filtered = mc_interface_get_tot_current_directional_filtered();
+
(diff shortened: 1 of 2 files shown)


vedderb:
Is there something missing in this PR? Only two lines are changed.

Peemouse:
Yes, it misses a lot... :confused:
I'll find out what happens and PR the right app_ppm.c file.
Sorry for that.

PR #190, opened 2020-07-02
Bmi160 spi support

EsbenFR:
Added basic SPI support for BMI160

TODO:

    Add support for shared SPI between different devices (mainly DRV SPI)
    Add support for external BMI160 SPI (using comm pins)


FelixvdDonk:
How do you enable this in VESC tool? I have wired my BMI160 on the COMM port according to the SPI pin assignments but I don't get a connection

PR #192, opened 2020-07-03
App PPM rework

Peemouse:
- **CHANGED** : Master keeps the slave(s) running even if a local fault occurs
- **CHANGED** : Master now sends relative value instead (%) instead of absolute value (A) (great for multiple VESC with different currents settings)
- **FIXED** : Timeout current brake feature not working
- **FIXED** : brake command sent over CANbus even if Multiple VESC over CAN disabled
- **FIXED** : delay before switching slave off in case of RX timeout
- **ADDED** : Traction control auto-disabling in case of fault on master, re-enable only if no effect on motors will happen (avoid weird behaviour exiting fault state on master).

### DIFF
--- a/applications/app_ppm.c
+++ b/applications/app_ppm.c
@@ -405,7 +405,7 @@ static THD_FUNCTION(ppm_thread, arg) {
 		}
 
 		if ((send_current || send_duty) && config.multi_esc) {
-			float current_filtered = mc_interface_get_tot_current_directional_filtered();
+
 			float duty = mc_interface_get_duty_cycle_now();
 
 			for (int i = 0;i < CAN_STATUS_MSGS_TO_STORE;i++) {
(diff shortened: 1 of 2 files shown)


Peemouse:
Please find the correct `app_ppm.c` in this file.
I don't know what happens, I still struggle with my local git setup.

You'll find quite few changes that are only comments addition. It helpt me to understand the way you designed the ppm app.
I thought that it may help other future contributors so I didn't delete them.
I hope it's ok.

vedderb:
Thanks, merged! I only made sure that it builds and a quick test on my bench, so testing is needed. The test build here are updated now:
https://vesc-project.com/node/1724
Would be great if you can do some testing and let me know if things work properly.

Peemouse:
I tested it on bench (dual motors) for all the control types.
Here a quick edit during my tests (not finished yet at this stage) : https://youtu.be/yc2fn4PBN6c

PR #196, opened 2020-07-21
Dev fw 5 02

Jfriesen222:
Mostly just some stormcore stuff and a slight modification to how motor temp faults are triggered.


mc_interface.c
```diff
@@ -1921,7 +1922,7 @@ static void update_override_limits(volatile motor_if_state_t *motor, volatile mc
 	float lo_max_mot = l_current_max_tmp;
 	if (motor->m_temp_motor < conf->l_temp_motor_start) {
 		// Keep values
-	} else if (motor->m_temp_motor > conf->l_temp_motor_end) {
+	} else if (motor->m_temp_motor > (conf->l_temp_motor_end + 10.0)) { // only fault if 10 degrees over max do give some cushion for noisey sensors
```
Peemouse, line 1925: Personnaly, I'm absolutely not fan of this workaround.
In case that the sensing is working perfectly fine, then the fault will trigger 10 (real) °C above the threshold !
It would have been better to use either a delay or a counter to reject spikes, not an hard coded offset.
Something like : 
```
if (motor->m_temp_motor < conf->l_temp_motor_end) {
 counter++;
}
else {
counter = 0;
}
if (counter > 20) {
 fault = true;
}
```
What do you think about that ?

nitrousnrg: Good catch, thanks. Temperature signals can be heavily filtered on both hardware and software. Software can even be set to only trip after 30 seconds above a temp threshold (Overload). However, a hardcoded 10°C offset hidden in the code is not acceptable.

Jfriesen222: Yeah, thanks for the discussion guys. I also think maybe there should just be a different form of fault that doesn't cause a timeout. Ultimately just reporting an overtemp fault without exacerbating the issue caused by a momentary spike with additional delayed time of no current in motor.

vedderb: In most cases the overtemp fault is not needed, as the motor current will be reduced to 0 at that point. An enum setting with motor overtemp behavior might be the best approach. The options can be:

0: Do nothing
1: Reduce current
2: Generate fault
3: Reduce current and generate fault

The default value can be 1: reduce current, which reduces the current but won't generate a fault. For e.g. the axiom hardware the default can be 3 instead, if that is important. That can be set in the hwconf file.

I can implement it next week, when I will become much more active with coding again. I had vacation for a couple of weeks and spent most time fixing stuff on my house.

Jeff, maybe you can remove the temperature behavior change from the PR and only include the stormcore stuff.

Jfriesen222: I think that sounds awesome. I removed the change.

mc_interface.c
```diff
@@ -1922,7 +1922,7 @@ static void update_override_limits(volatile motor_if_state_t *motor, volatile mc
 	float lo_max_mot = l_current_max_tmp;
 	if (motor->m_temp_motor < conf->l_temp_motor_start) {
 		// Keep values
-	} else if (motor->m_temp_motor > (conf->l_temp_motor_end + 10.0)) { // only fault if 10 degrees over max do give some cushion for noisey sensors
+	} else if (motor->m_temp_motor > conf->l_temp_motor_end) { // only fault if 10 degrees over max do give some cushion for noisey sensors
```
Peemouse, line 1925: Sorry for being kind of fussy, but could you also remove the comment for less confusion ?

Jfriesen222: not fussy at all thanks for the catch

PR #215, opened 2020-09-20
Positional Control Using PPM

wtjstanley:
This was a quick feature to allow for -180 to 180 degree control from the ppm port on the vesc. It uses the standard PID positional control that already exists.


vedderb:
Can you base this on the dev branch? The PPM app is reworked a bit since the last master release. I can add the enum to the config in vesc tool.

wtjstanley:
Sure thing.

wtjstanley:
Switched the branch but looks like I accidentally committed the builds I made.

PR #216, opened 2020-09-26
Firmware odometer & verify MC/APP config via CRC

trosenband:
Store odometer in emulated EEPROM (update at shutdown).
   * New "dist" terminal command shows odometer and trip distances.
   * Odometer value is sent to UI via COMM_GET_VALUES_SETUP.
   * UI can set the odometer via COMM_SET_ODOMETER.
   * Unfortunately, the bootloader clears the odometer to 0 when FW is reprogrammed.

FLASH will be written a little more, so verify integrity of MC and APP config with CRC.
   * Currently the FW code has CRC checks, but configs are also important!
   * This helps maintain FLASH integrity when the config pages are occasionally
      re-written due to odometer updates.
   * If the config CRC checks fail, store the fault and fall back to default config.
   * New 'crc' terminal command displays CRC values.

[Pull request to display the odometer in VESC Tool.](https://github.com/vedderb/vesc_tool/pull/93#issue-493626958)

### DIFF
--- a/applications/app.c
+++ b/applications/app.c
@@ -25,6 +25,7 @@
 #include "rfhelp.h"
 #include "comm_can.h"
 #include "imu.h"
+#include "crc.h"
 
 // Private variables
 static app_configuration appconf;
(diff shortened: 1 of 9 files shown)


vedderb:
Nice! I will have a closer look and try to merge it as soon as I finish what I'm currently working on. Just one comment for now: when adding new fault codes, it is best to add them at the end of the enum. That will maintain backwards compatibility.

vedderb:
Also, just noticed that you based the pr on the master branch. It would be good if you can base it on the dev branch instead.

trosenband:
Thanks for the advice.  Also, my config CRC checks are a bit cumbersome compared to the existing code-check that uses [post-complemented CRC](https://stackoverflow.com/questions/58393307/verification-of-a-crc-checksum-against-zero).  I can resend the PR if needed (but haven't yet figured out the post-complementing trick).

trosenband:
Superseded by new PR.

PR #219, opened 2020-10-06
Lower default fault stop time

b264:
The high default of 500ms can easily knock folks off their electric longboards --- and after lots of experience, I feel like sub-100ms values work better. It could be more risky for some hardware but I feel like it's safer for most humans. Given the choice between body or ESC I will choose body every time.

I'd like to propose changing this default.

This updates the default Fault Stop Time in the Motor > General > Advanced tab, for hardware that hasn't provided its own value.

In combination with https://github.com/vedderb/vesc_tool/pull/95

### DIFF
--- a/mcconf/mcconf_default.h
+++ b/mcconf/mcconf_default.h
@@ -385,7 +385,7 @@
 
 // Misc
 #ifndef MCCONF_M_FAULT_STOP_TIME
-#define MCCONF_M_FAULT_STOP_TIME		500	// Ignore commands for this duration in msec when faults occur
+#define MCCONF_M_FAULT_STOP_TIME		85	// Ignore commands for this duration in msec when faults occur
 #endif
 #ifndef MCCONF_M_RAMP_STEP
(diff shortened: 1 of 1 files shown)


Mitchlol:
`Given the choice between body or ESC I will choose body every time.`
That's not actually a possible choice, its either body or body and ESC. 

Granted for most situations its just body or no body, as the ESC wont go. However unless you have safestart disabled, reducing the fault time shouldn't have any effect on a user with a remote getting tossed off a board.

b264:
Hitting small pebbles during hard braking seems to cause exactly that.

It's the difference between feeling a sudden power/brake loss, or getting knocked off balance and falling off.

mhweiner:
Can we get some more eyes on this? @Mitchlol I'd like to get some clarification on what you said. I understand that Safe Mode won't allow the startup of the motor unless there is a commanded duty cycle of 0 for "several cycles after boot." However, does the vesc always restart after a fault, or does it gracefully recover? If the latter is true in some cases, then the Fault Stop Time would be in play here.

PR #220, opened 2020-10-06
Lower default duty cycle current limit start

b264:
This needs to default less than Maximum Duty Cycle which defaults to 95%.

This updates the default Duty Cycle Current Limit Start in the Motor > General > Advanced tab, for hardware that hasn't provided its own value.

In combination with https://github.com/vedderb/vesc_tool/pull/94

### DIFF
--- a/mcconf/mcconf_default.h
+++ b/mcconf/mcconf_default.h
@@ -114,7 +114,7 @@
 #define MCCONF_L_CURRENT_MIN_SCALE		1.0	// Minimum current scale
 #endif
 #ifndef MCCONF_L_DUTY_START
-#define MCCONF_L_DUTY_START				1.0 // Start limiting current at this duty cycle
+#define MCCONF_L_DUTY_START				0.85 // Start limiting current at this duty cycle
 #endif
 
 // Speed PID parameters


Mitchlol:
This change seems dangerous for balance app users.

b264:
As-is it's dangerous for non-balance esk8 users.

This is also merely a "default default" as each hardware can specify its own default, like Stormcore does:

https://github.com/vedderb/bldc/blob/f560f27871878cda3474cf29e547ed8f52e5fb0e/hwconf/hw_stormcore_100s.h#L245

As-is this causes lots of oscillations and jerks right at top speed, the most dangerous time to have that.

Is there a different change that can be better for more scenarios?

venom121212:
I crashed when my motors started freaking out, wobbling out of sync. This was before anyone recommended this setting be lowered.

Setting lowered to suggested default of 85% eliminated issue immediately.

vedderb:
This is indeed a problem for the balance app, as the tiltback settings won't work as expected with this setting. Also, it will limit the top speed on all setups.

Currently I don't have any setup that shows oscillations, but I will try to reproduce them and see if I can fix the problem.

If this is an inherent problem due to e.g. to high impedance at the power source, we can look into adapting the tiltback settings to take this value into account.

b264:
The problem typically doesn't present itself until the vehicle is under load carrying a person.

On the bench it's hardly apparent.

vedderb:
Just a not that I haven't forgotten about this, and will definitely get it fixed for the next release. I made some experiments and have a few ideas for modifications to the current controller that can help, so that the top speed can be retained with reduced risk for oscillations.

If anyone can who can reproduce the problem can help me testing in the next weeks it would be great.

mhweiner:
@vedderb can we at least make a note/warning in the Vesc Tool somewhere to alert esk8 folk to lower this setting in the meantime as a stopgap measure? I think that would solve at least 50% of the problem. If that's not the most appropriate place, could you come up with another more suitable place for such a warning?

vedderb:
I spent some time on this, and I think I managed to reproduce the problem on the bench. The problem was what I expected that it was. Here is an attempt at fixing it:
https://github.com/vedderb/bldc/commit/66b4d8f56d829b7f2800820a720546d70f5ff491

Here is a build with the latest version
https://vesc-project.com/node/1724

Filcas:
@vedderb do you think this can have impact on the multi-rotor motors being locked at full speed while in duty-cycle control mode? I've been trying change every single setting on vesc 6 and not able to get rid of this. I will give this build a try but I would like to hear some input from you, as it might be another issue that we could not understand. Let us know what you think. Thanks!

b264:
93% of the 71 esk8 builders [I polled](https://forum.esk8.news/t/vesc-hd-a-credit-card-sized-twin-motor-controller/43178/56?u=b264) thought this change was a good thing.

PR #224, opened 2020-10-14
Add hall sensor interpolation erpm and sample count

Mitchlol:
I worked with @blezalex on this. Simplified it down to two changes that give a big impact. For me increasing the interpolation ERPM had the biggest impact (game changer), @blezalex is into the increased sample count. Regardless they're now both configurable.

I must admit I'm not super confident with the level of testing on this. I don't have any dual motor hardware to test, and on my setup the sampling doesn't seem to have any impact, but its simply a for loop and some math, and it looks right to me. The interpolation change is easy to test as you can hear the square wave. But this is the dev branch after all :rofl: 

I left the default values where they were, so this change should be non-impacting for existing users.


vedderb:
I just pushed my pile of code and commits from the past months, and was already half way done implementing this. Instead of merging I made this commit, which should make it equivalent:
https://github.com/vedderb/bldc/commit/af7cd36cb861b7867b2094b7fa6f925efbe226b3

Can you give it a try?

Mitchlol:
@vedderb 's changes cover these, so I'm closing this.

PR #225, opened 2020-10-20
Fixed two issues relating to libcanard and updated library

sprenger120:
After updating libcanard (50c95ac) and enabling UAVCAN in appconfig the canard thread locked up due to setNodeId being called in every iteration (068f5ad). This happens now because the update put in the assertions again that were commented out. Was there a specific reason for this? A local fork runs the library with assertions without problems.

Also after looking through the code appconfig is loaded after the can thread is started 1321299 which needs config to be ready for nodeId. This might not be an issue because its unlikely that the main thread is preempted exactly before the loading is finished.

### DIFF
--- a/libcanard/canard.c
+++ b/libcanard/canard.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2017 UAVCAN Team
+ * Copyright (c) 2016-2019 UAVCAN Team
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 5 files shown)


vedderb:
It was long ago that I worked with uavcan, but I think I set the ID in every iteration so that changing the appconfig would take effect without a reboot. Setting the ID was a light operation too, so there was not really a problem.

I would prefer changing libcanard so that the ID can be changed any time. So far I have tried to make all configuration changes take effect instantly as that is quite nice.

sprenger120:
Thanks for the clarification this makes sense. I've updated the PR so setting Id now always works.

bsilver8192:
I independently found the initialization order issue and fixed it the same way in #625, found this looking through open PRs afterwards, for reference.

Issue #231, opened 2020-11-11
UNITY does not respond with second motor controller_id


wunderk1nd-e:
If I've understood correctly the call to `COMM_GET_VALUES` (and probably others) should return `mc_interface` data pertaining to the second motor.
These calls are currently returning with the primary `controller_id` making them, as far as I can tell, indistinguishable from the primary calls.

If the above understanding is correct then I think the `controller_id` in the context of the second thread should be the result of `utils_second_motor_id()`

wunderk1nd-e:
Example here: https://github.com/vedderb/bldc/pull/232

PR #232, opened 2020-11-11
[COMM_GET_VALUES] Unity should return second motor controller ID if that is the selected thread

wunderk1nd-e:
This addresses: https://github.com/vedderb/bldc/issues/231


vedderb:
Correct, well spotted!

PR #235, opened 2020-11-20
Feature app io default hw

cabralmachado:
Added io app to reconfigure UART pins as output digital pins. 
Control of pins is done using VESC terminal messages.
App is always started by default, but pins are only configured after a call to the io app.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,8 @@ build
 .project
 .cproject
 .settings/language.settings.xml
+bldc.config
+bldc.creator
+bldc.creator.user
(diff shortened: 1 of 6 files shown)


Mitchlol:
What is/are the use case(s) for this app?

cabralmachado:
Hi Mitch,

We use it to actuate an H Bridge in one of our robots, but it's just a
generic controllable digital output.

It is not supposed to be selectable from the app menu in vesc_tool but
always available if needed.

Kind regards,
Diogo

PR #237, opened 2020-11-29
hw_60.h flips drv8301 spi pin definitions

Teslafly:
I have been wondering why the commands: drv8301_read_reg x and drv8301_print_faults have appeared to not work. only returning 0xffff and all faults set:
```| FETLC_OC | FETHC_OC | FETLB_OC | FETHB_OC | FETLA_OC | FETHA_OC | OTW | OTSD | PVDD_UV | GVDD_UV | GVDD_OV |```

It turns out that the defines for the spi pin definitions on the hw 6.1/2 and 6.3/4 have been swapped. 

Figuring this out was made much harder by the lack of schematics for the hw6.3/4 hw versions. As they have not been posted to https://vesc-project.com/node/311

I had to ohm them out on the two different hardware versions to confirm my suspicions, having just gotten my hands on a v6.4 board.

Not sure how this wasn't caught before now. I would think that part of trampa's production process would be testing communication with the drv and verifying no drv faults.

This pull request gives expected results on both my 6.1 hw, and trampa made 6.4 hardware:
```
## on 6.4 hw
$ drv8301_print_faults
No DRV8301 faults

$ drv8301_read_reg 0
Reg 0x00: 01111000 00000000 (0x7800)

$ drv8301_read_reg 1
Reg 0x01: 00001000 00000001 (0x0801)
```


vedderb:
Well spotted, thanks! Fixed manually in dev branch, as the HW file has changed a bit since.

I haven't looked at the DRV8301 code in a long time, so I completely missed this. The fault reporting was still working, as the fault pin does not require SPI to work. Only the DRV faults would be displayed incorrectly in the message.

About the schematics of the MK3 to MK5, it is on my huge todo list to dig them up and put them on the site. There are no major differences really, mainly that the IMU is replaced with the BMI160 as the MPU9250 isn't produced anymore.

Teslafly:
The main other thing I have wanted to see the schematics for is how the power switch circuitry is implemented. I have come up with my own implementation on a higher voltage design, but it would be nice to know how you did it. Now that I have a 6.4 hw I can ohm things out, but would rather not. 

It has been 3(+?) hardware versions and  since the last schematic was published. That's not great.
what is great is how this project has grown though. It has been interesting to watch and see the patches.

PR #238, opened 2020-11-30
MT6816 SPI encoder support

nitrousnrg:
This encoder needs x2 16bit SPI transfers so the hardware SPI peripheral was used to run SCLK at 10.5MHz and reduce the ISR duration.

This IC needs MOSI data, which works poorly in the bitbanged code. I also tried DMA accesses but overall its slower so I think this is the best way to release CPU time so its better invested doing FOC.


nitrousnrg:
Forgot to mention, on VESCTool end this PR needs:

- Add SENSOR_PORT_MODE_MT6816_SPI to the sensor port enum list in mcconf xml. This will require updating the confgenerator hash.

- Update datatypes.h Add the FAULT_CODE_ENCODER_NO_MAGNET text string

I think its easiest for you to make those changes yourself instead of managing a vesc tool PR. If I'm wrong let me know and I can make the PR!

vedderb:
Looks good! I will add the enums to VESC Tool.

vedderb:
Can you just send me a help text for the sensor port mode?

bild

nitrousnrg:
Thanks!

Here is the text, feel free to modify it:

MT6816 Encoder
A magnetic encoder using a high speed SPI communication. Provides absolute position from start. It has to be connected to a hardware-based SPI peripheral because software bit-banging is too slow.

Teslafly:
@nitrousnrg  
So this looks like it is using 4 wire spi instead of 3 wire spi in duplex mode? can you confirm?

a link to documentation on connecting this would be helpful.

If 3 wire isn't implemented, I may take a wack at it. need it for the TLE5012B encoder anyway.

PR #243, opened 2020-12-07
Pedal assist support (PAS)

nitrousnrg:
This commit enables cadence-based pedal assist for ebikes using
quadrature-style pedal speed sensors.

There are 2 operation modes:
* PAS only: Motor current is based only on pedal feedback
* ADC + PAS: The code will use both ADC and pedal feedback and use
the strongest command to provide seamless overlap when the user is
pedaling and requests extra torque with the throttle.

For this to work it needs the PAS hall sensor pins configured in the hw_xxx.h,
this is an example:
```
// Pedal Assist pins
#define HW_PAS1_PORT			GPIOB
#define HW_PAS1_PIN				5
#define HW_PAS2_PORT			GPIOB
#define HW_PAS2_PIN				4
```

I'm preparing a VESC Tool Pull Request that unlocks the GUI side of this.


vedderb:
Looks good!
I made the PAS pins default to the UART pins if they are undefined. Do you think there is any issue with that?

nitrousnrg:
Glad its merged, I was about to post some datalogs to prove its has been tested.

> Looks good!
> I made the PAS pins default to the UART pins if they are undefined. Do you think there is any issue with that?

I don't see any real problem with that. One potential problem is that usually a system with PAS most likely will have a display, and most displays use a serial communication so I wonder if people would struggle to have both PAS and display support if PAS takes the UART pins by default. Anyway, I'd say go for it, 2 lines of code are the least of their problems, ebike displays suck because they require some custom and undocumented wiring.

There is another patch coming that provides support to some ebike displays but I'm leaving it for last as it provides very basic support, and will leave it for right before the product goes on sale :)

Mitchlol:
Sick!

kylerlaird:
I'm building ebikes for my family using the Sempu T9 pedal torque BB.  It has analog torque and digital cadence outputs.  I was planning to just use the torque output to command current (with braking, preferably).  I can do this with the standard setup, using the torque sensor as the throttle input, right?  This leaves the cadence sensor unused.  (When I was still thinking of using an Arduino and another ESC, I considered using cadence for braking.  Brake if torque and cadence are both zero.)

Is there any advantage in using PAS mode over standard throttle mode?  (Note that I'm new to VESC so I might be missing some basic concepts.)

nitrousnrg:
you might need to use the quadrature sensor (cadence) anyways to make sure the user is not pedaling backwards...

kylerlaird:
What would be the issue to resolve if the user pedals backward?  The torque will be zero, so it should slow/stop.

nitrousnrg:
> What would be the issue to resolve if the user pedals backward? The torque will be zero, so it should slow/stop.

Oh, I've been told that the torque sensor could produce an output when the user was pedaling backwards, but its not necessarily your case

kylerlaird:
This one seemed to only register a change while torqued forward when I static tested it.  Besides, these bikes freewheel so it's hard for me to imagine getting enough torque to register when backpedaling.

Sidenote...the quadrature signal is handled internally on the T9 and a simple (single) square wave is output for cadence only for pedaling forward.  I'm a little disappointed because I was hoping to use backpedaling to activate regen braking (like a coaster brake).  Now I'm hoping to regen when torque and cadence are zero - like a very forgiving fixie.

afzalmam:
I am trying to understand the PAS code. In the current code, pas_event_handler() is polled instead of being invoked from ISR. As it is the case, if update_rate_hz is low enough, wouldn't it cause pedal_rpm calculation & thus torque control go wrong ?, sorry if I am wrong, based on code it is what I felt.

Consider the default values for PAS variables,
pedal_rpm_start = 10;
pedal_rpm_end = 120;
update_rate_hz = 500;
magnets = 24;

sleep_time = 2ms (1/500), i.e. thread would get invoked every 2 ms
min_pedal_period = 0.17s [ 1/ ((120/60) * 24 * 3) ]

If both PAS signals are high when pas_event_handler() is invoked, new_state = 3. As both signals can be high during the next invocation of pas_event_handler() (as it would get invoked every 2ms), again new_state = 3.

So per pas_event_handler(),
timestamp - old_timestamp = 2ms;
period = 2ms * 24 = 48ms;
period_filtered = period = 48ms = 0.048s;

Since period_filtered(0.048) < min_pedal_period(0.17), pedal_rpm would not be calculated at this point of time, this case is okay.

Now suppose update_rate_hz = 100, thread would get invoked every 10ms, i.e.,
timestamp - old_timestamp = 10ms;
period = 10ms * 24 = 240ms;
period_filtered = period = 240ms = 0.24s;

In this case, period_filtered(0.24) < min_pedal_period(0.17) is no longer true & hence pedal_rpm would get updated to a value that corresponds to the thread frequency & not to QEP signal frequency. If that is the case the torque would not correspond to the cadence.

Based on the above, update_rate_hz should be a high enough value for the current PAS code to work, right?

nitrousnrg:
Yeah I used pretty high polling frequencies (500Hz), we could set a minimum update rate in vesc tool, or just hardcode it too 500Hz.

Originally I intended to use interrupts, which is why pas_event_handler() was named that way. Interrupts were okay, but made the code diff a bit less clean for no real reason as it worked just fine without them. If someone else wanted to use different pins the change in the isr files would be a somewhat dirty so I settled for the polling approach.

afzalmam:
Thanks for the clarification

Ecyclist:
Is PAS build into VESC Tool?
Con you do YouTube video how to set up PAS?
Thank you

RobertXuChun:
The PAS app is great. It works when in PAS alone mode. However, I could never make it work in ADC + PAS mode. Only the ADC works when in the combined mode. I think it's due to the pas_event_handler() being polled instead of being invoked from ISR. I could be wrong. But could anyone test this and hopefully solve the problem? Thank you.

nitrousnrg:
When I coded this I only covered the use case where the ADC also has a brake signal:
```
                case ADC_CTRL_TYPE_CURRENT_REV_BUTTON_BRAKE_CENTER:
		case ADC_CTRL_TYPE_CURRENT_NOREV_BRAKE_CENTER:
		case ADC_CTRL_TYPE_CURRENT_NOREV_BRAKE_BUTTON:
		case ADC_CTRL_TYPE_CURRENT_NOREV_BRAKE_ADC:
		case ADC_CTRL_TYPE_CURRENT_REV_BUTTON_BRAKE_ADC:

```
In particular I use ADC_CTRL_TYPE_CURRENT_NOREV_BRAKE_ADC.

My idea was to force an immediate motor release when the user applies brakes (many ebikes have brake sensors) because cadance PAS has like 1 second of ramp up / ramp down times and I don't want the motor to keep pushing when you are braking.

My guess is that you are using ADC_CTRL_TYPE_CURRENT which doesn't read the PAS state. Not sure if we should include this mode. Should we?

RobertXuChun:
Yeah, you made my day! Thank you so much for your quick and valuable information. I didn't read your code carefully. ;) I did use ADC_CTRL_TYPE_CURRENT and when it didn't work I changed to a duty cycle mode and it still didn't work. Now I understand. I think you should include this mode or provide some tips on using PAS+ADC. 

By the way, I adapted your code to make my particular cadence sensor work. It works but not so predictable.. Sometimes the motor simply continue run even I stop pedaling. Could you check the code below and see if it could be improved? Thank you so much for your time.

void pas_event_handler(void) {
#ifdef HW_PAS1_PORT
	//const int8_t QEM[] = {0,-1,1,2,1,0,2,-1,-1,2,0,1,2,1,-1,0}; // Quadrature Encoder Matrix
	//float direction_qem;
	uint8_t new_state;
	static uint8_t old_state = 0;
	static uint8_t count = 0;
	static float old_timestamp = 0;
	static float inactivity_time = 0;
	static float period_filtered = 0;

	//uint8_t PAS1_level = palReadPad(HW_PAS1_PORT, HW_PAS1_PIN);
	uint8_t PAS2_level = palReadPad(HW_PAS2_PORT, HW_PAS2_PIN);

	new_state = PAS2_level;
	if (new_state != old_state)
	  count++;

	//direction_qem = (float) QEM[old_state * 4 + new_state];
	old_state = new_state;

	const float timestamp = (float)chVTGetSystemTimeX() / (float)CH_CFG_ST_FREQUENCY;

	// sensors are poorly placed, so use only one rising edge as reference
	if(count > 1) {
		float period = (timestamp - old_timestamp) * (float)config.magnets;
		old_timestamp = timestamp;

		UTILS_LP_FAST(period_filtered, period, 1.0);

		if(period_filtered < min_pedal_period) { //can't be that short, abort
			return;
		}
		pedal_rpm = 60.0 / period_filtered;
		pedal_rpm *= direction_conf;
		inactivity_time = 0.0;
		count = 0;
	}
	else {
		inactivity_time += 1.0 / (float)config.update_rate_hz;

		//if no pedal activity, set RPM as zero
		if(inactivity_time > max_pulse_period) {
			pedal_rpm = 0.0;
		}
	}
#endif
}

nitrousnrg:
Have you tried longer ramp times? I configured VESC Tool to use 0.3s by default, but then we found that sometimes with shocks or vibrations the pas would be engaged and the motor spin. So we made the up/down ramp to be 1.2sec and the shock issue was removed.
I wonder if that could be your case as well.

Definetively the help page could use some improvement, will try to make it happen on my next vesc tool PR

RobertXuChun:
That's very likely. Great tip again. I will try it in another day. I could never figure this out without your advice. Looking forward to your next PR.

albertomercurio:
I really like it! i want to test it but i have a 3-wires PAS sensor, with GND, VCC and OUT. Is there any way to make it work?

RobertXuChun:
My PAS sensor has 3 wires like yours. You have to adapt the app code to make it work with your sensor, like my code snippet above. Or nitrousnrg may release a new version which incorporates the 3-wire sensor. 

To nitrousnrg: My pas sensor is indeed very unreliable. It has 2 hall sensors and 12 magnets, but only 3 wires. Even when I crank the shaft with my hand at home, it will be engaged quite frequently. I can tell this because I connected the signal wire to the tx pin. The led will light when magnet passes.

afzalmam:
@RobertXuChun Me too have 3 wire sensor. PAS code was adapted to take care of single signal instead of QEP, but I did not face the issues you mentioned about motor running with no pedaling. I then made the changes like you made (your code looked better than mine), still no issue was observed. Note that I am testing without connecting the sensor to pedal, i.e. tested by rotating rotary part of sensor by hand (rather finger). Sensor was part of a bicycle conversion kit, checking signal in oscilloscope, it seems to be a 12 magnet one.

livello:
@nitrousnrg 
photo_2021-08-13_09-22-32
photo_2021-08-13_09-25-17
[app_motor_config_livello.tar.gz](https://github.com/vedderb/bldc/files/6980339/app_motor_config_livello.tar.gz)

Unreliable motor stop after stop pedaling. Every time when stop pedaling motor stops very slowly. I think this is wrong way.  Motor must stop immediately after stop pedaling.

livello:
I guess I need to disable uart. I use default hw_config, no modification of source code.

adnion:
@afzalmam @RobertXuChun @albertomercurio Could you please help me to build vesc tool for 3-wire PAS? How can I change PINs so I can use a display on the UART port?

livello:
@afzalmam  did you see the first message? There is an instruction how to change pins. You will face problems with PAS what I mentioned. But if you plan to ride fire or paved roads it will not stop you...

afzalmam:
@adnion Code snippet by RobertXuChun above in the thread works for me

afzalmam:
@nitrousnrg i have gearless e-cycle and the location here has a lot of high gradient slopes.  It was difficult to ride uphill with Cadence control. To make ride comfortable in my situation, a new control type - Constant Torque has been, more details in the pull request - https://github.com/vedderb/bldc/pull/346, let me know if you have any comments

livello:
@007trains look at my issue:
https://github.com/vedderb/bldc/issues/331

I've adopted some fixes that solved some problems:
- motor continue run forever after stop pedal rotation in some positions
- I changed  pins for pas sensor to free default uart pins

rupebac:
so, where can I buy a good PAS that works out of the box with Vesc? How do I wire them? VCC to 5V I assume, GND, and one to RX and the other to TX  in the uart port? is there any documentation? thanks!

livello:
> so, where can I buy a good PAS that works out of the box with Vesc? How do I wire them? VCC to 5V I assume, GND, and one to RX and the other to TX in the uart port? is there any documentation? thanks!

Out of the box is not about vesc. You may need pull up and protection resistors for reliable work. At least you should use more recent app_pas.c version like in my patch.

dronisuav:
> app_pas.c

@nitrousnrg 
Using UART is not convenient. Is it possible to use ADC3 and PPM? ADC3 is not used, PPM is also not used.

geofrancis:
uart is not avalible when using bluetooth or a screen, rc input would be ideal.

livello:
@dronisuav I am using pas with fsesc 6.6 on PA5, PA6 pins. 
hwconf/hw.h is the place to set pins:
https://github.com/livello/bldc/commit/4f92b12ac3962f2b800efafb7b8b52e396c4f282#

dronisuav:
> @dronisuav I am using pas with fsesc 6.6 on PA5, PA6 pins. hwconf/hw.h is the place to set pins: [livello@4f92b12](https://github.com/livello/bldc/commit/4f92b12ac3962f2b800efafb7b8b52e396c4f282)

PA5 and PA6 are ADC1 and ADC2 right ? With this you loose Throttle/brake function. Correct me if I am wrong.

livello:
@dronisuav according to vesc 6 schematic on which fsesc 6.6 is based:
https://vesc-project.com/sites/default/files/Benjamin%20Posts/VESC_6_plus.pdf
screenshot_2022-04-18_12_05_01
MISO SCK pins are used in my case.

Teslafly:
You could use the swd pins too. (having options is probably a good idea). If you are using the swd pins for debugging,  you have the know how to change the pinout if needed.

livello:
rx, tx for default pins is not the best for out of the box configuration. 
PAS works fine since 5_03 release with this commit.
https://github.com/vedderb/bldc/commit/627c12b1ca3400dd50e72c3707d6acac08176b76

rupebac:
@livello so latest master should work fine with 3 wire Pas? Would you please give us some instructions on how to wire it? Thanks so much in advance :)

dronisuav:
@livello @nitrousnrg 

are possible to work only with cadence sensor ? without Torque. Should work on PAS or ADC and PAS ?

where is possible to read cadence or torque inputs/status. Because even when I connect it how to check right connection ? How 
 to check? I see only ADC1 and ADC2 values. ADC3 and PAS are not visible in Vesc Tool, or maybe not ? But where, and how ?

livello:
@rupebac 3-wire is a single ended encoder? VCC, GND, Vout?  At now app_pas support only quadrature encoder. 2 channels needed for detecting direction. Otherwise you would not stop your bike when any pedal movement will cause motor to go.

@dronisuav you can run vesc_tool 4.0 select pass application and look closer to settings. 
You can open app_pas.c file and look into code. 

The best way to check your connections is oscilloscope. A,B with 10k Om to VCC or GND, depends on sensor type. A,B with 200 Om to MCU for limit current. 

I've succesfully used panasonic optical cadense sensor with 17 poles  and self-made hall-effect sensor with 8 pairs of magnets.

saad27-1:
Many thanks for posting this,

I currently have PAS only mode working fine, but PAS + ADC results in only ADC working with no PAS

I believe this is since I am using on ADC1 a throttle not brake sensors (will be wiring brake sensors to ADC2)

How would I get ADC_CTRL_TYPE_CURRENT to read the PAS state?

livello:
@saad27-1 could you shot a video which gives us full information about your problem include motor and app config xml files?
Can you test bug motor runs forever after stop pedaling in some pedal positions? I faced with it again on latest release. I'll back on my revision of app_pas.c

livello:
@saad27-1 I'm not sure. Problem may exists in your ADC configuration. I have no experience with it. You can try different modes Control Type and Mapping.
At first look there is code
https://github.com/vedderb/bldc/blob/5f94f2d0c97dc6a114b0738efb91735406743e56/applications/app_adc.c#L313
https://github.com/vedderb/bldc/blob/5f94f2d0c97dc6a114b0738efb91735406743e56/applications/app_pas.c#L291
https://github.com/vedderb/bldc/blob/5f94f2d0c97dc6a114b0738efb91735406743e56/applications/app_pas.c#L114
must work. I' also want to use throttle with pas like you. Hope I'll do it soon.

geofrancis:
has anyone got any links to a pas sensor that is supported?

kylerlaird:
It's not "supported" as far as I know, but I use these.
    https://www.aliexpress.com/item/33030138390.html

rupebac:
@livello alright, then a PAS compatible with CA v3 should work right? Like this one:

https://fasterbikes.eu/media/pdf/e1/aa/ce/PAS_24PolevKnowBLLe6nLGqQrxM4uiZbQ4i.pdf

(notice the input from 5-10v).

Can you please confirm me before I pull the trigger and order it :)

Thanks a lot for your work and support!

livello:
@rupebac I don not use CA and do not know much about it. 
Your PAS_24P should work but you may need voltage dividers that restrict signal level to 3.3V.
Without recompilation, the pass will work, but it will only be possible to configure the vesc via USB not bluetooth.
Moreover, there was a glitch with the constant operation of the motor. How much it interferes with driving, I can not say.

Try it, tell me.

rupebac:
oh yes, you are right, stm32 does not support more than 3.6V on ADC inputs... alright, then something like this on both PAS signals should do fine I guess:

schema

"Moreover, there was a glitch with the constant operation of the motor. How much it interferes with driving, I can not say."

you mean, you fixed this yourself in #331 but it still didn't make it to master? Alright, will try it out, with stock 5.3 first and let you know. I anyway need to setup connectors and everything, it just arrived, so it will take me a while

thanks dude :)

livello:
@rupebac how is your progress?
I found out that adc_ext, adc_ext2 pins are using for acceleration and brakes.  No luck with rotation adc_ext and adc_15 in code.
Now I'm trying to use swdclk and swdio pins for pas. 
`#define HW_PAS1_PORT			GPIOA
#define HW_PAS1_PIN				13
#define HW_PAS2_PORT			GPIOA
#define HW_PAS2_PIN				14`

Also no luck. 9 may In the evening I'm going to use oscilloscope for watching the problem.

Teslafly:
No. The canbus pins go to the can tranciever and are not accessible as gpio's. However the swd pins can be.

livello:
@007trains I'm glad to see you here! @Teslafly is right. Bad idea. 
Go with swd pins as I described early. Worked good for me on test mode. I'm stuck with case for my fsesc 6.7. Hope will ride soon... 
for #331 use my version of https://github.com/livello/bldc/blob/IS-331pr/applications/app_pas.c

007trains:
@Teslafly @livello Thanks, I've switched to SWD and its working,
Next I'll test out your version of the PAS app and see how it goes
Cheers

007trains:
@livello your version of the PAS app is working great, runs on SWDIO+SWCLK pins without issue and starts/stops correctly
Modified adc.c to include UART

	case APP_PAS:
		app_pas_start(true);
            hw_stop_i2c();
		app_uartcomm_start(UART_PORT_COMM_HEADER);
		break;

	case APP_ADC_PAS:
		app_adc_start(false);
		app_pas_start(false);
            hw_stop_i2c();
		app_uartcomm_start(UART_PORT_COMM_HEADER);
		break;

Which also seems to work, will tidy up my wiring start field testing it

P.S. Setting the ADC control type to "Current" seems to break PAS somehow,
Other ADC control modes seem to work fine though

jon2com:
I tried the PAS mentioned:
https://fasterbikes.eu/media/pdf/e1/aa/ce/PAS_24PolevKnowBLLe6nLGqQrxM4uiZbQ4i.pdf
with a Flipsky 75100:
https://flipsky.net/collections/electronic-products/products/flipsky-75100-foc-75v-100a-single-esc-base-on-vesc-for-electric-skateboard-scooter-ebike-speed-controller
Running native 5.3 but cannot get it to work. connected it to the comm port and used 5v, gnd and rx/tx for the signals.

Tried both "pas" mode and "adc and pas" but nothin happens when pedaling

any  ideas

livello:
@jon2com 
0) better one time to see, than hundred times to hear. Put screenshots of your config or even export your app settings file. 
control type candence:
image

1) check voltage level on s1, s2. Is there 3.3V? If it's higher then use voltage divider.
2) 75100 has two different uarts. Use third:
https://cdn.shopify.com/s/files/1/0011/4039/1996/files/7_2179ea3d-2b8a-48de-8e04-261caff70566_1024x1024.jpg?v=1637046090

3) check pulses with prober at constant voltage sense mode.  Or even better to check it with oscilloscope.

jon2com:
Thanks for your reply. I got it to work by changing control type to cadence. I had it set to constant torque. After tweeking some parameters like rpm start and stop the bike runs very smooth.
My bike runs a 1000w hubmotor and pas is my only throttle input. I also use ADC2 for regen brakes which works nice

rupebac:
> @jon2com 0) better one time to see, than hundred times to hear. Put screenshots of your config or even export your app settings file. control type candence: image
> 
> 1. check voltage level on s1, s2. Is there 3.3V? If it's higher then use voltage divider.
> 2. 75100 has two different uarts. Use third:
>    image
> 3. check pulses with prober at constant voltage sense mode.  Or even better to check it with oscilloscope.

Guys, I accidentally fried my vesc while I was connecting the oscilloscope to debug my problem. I just ordered another one, will fix the old one in the future.

Anyway, what could be the reason of Pas not working? Connected an iic converter 5v to 3.3v from pas signals to RX/Tx in Comm port (converter works fine, I tested it apart). But motor wasn't moving at all. It is set on cadence, and firmware 5.3...

How do you analyze the signals coming to RX/tx? I am afraid to connect it to oscilloscope again (probe was on X10).

jon2com:
I connected the PAS to the COMM port on the 75100 I did not use 5vto3.3v converter either, and I am running stock 5.3. Works pretty well I think

rupebac:
that is really strange...  According to [this](https://vesc-project.com/sites/default/files/Benjamin%20Posts/vesc_75_300.pdf) COMM port is connected directly to the scm32, and this I think only supports 5V in digital mode.... which is I think why it is working for you. Look [here](https://forum.esk8.news/t/dont-wire-adc-to-5v/40871/33):

But then why it didn't work for me? maybe signals order matter?

I have compiled latest master, which seems to already have all the changes from @livello, and I changed PAS port to: SWDIO and SWCLK... Will try this next, as this should also free up the COMM port for another throttle.

Which headers you guys using to compile firmware for Flipsky 75100? the V2 from [here](https://forum.esk8.news/t/how-to-update-firmware-on-the-flipsky-75100-foc-esc/61819)? or the 75_300_R2 from the public repo?

rupebac:
This is the output of [this](https://fasterbikes.eu/de/zubehoer/cycle-analyst/192/teilbarer-kurbelsensor-pas-fuer-ca-v3?c=52) PAS:

osci1

does it look alright?

livello:
@rupebac "how I place the sensor" this is the key!
You can go study "quadrature encoder"

You should see something like this: 
изображение

rupebac:
yes, issue found... this sensor has to be placed vertically... only the sensor area of the body of the sensor should be near the magnet ring... now it looks alright. Will configure everything and test it with motor. Thanks!

rupebac:
for me, DIO and CLK pins (connector just above the CAN one in the 75100) is NOT working, at least in 5.3... 

I copied pasted the app_pas.c from livello. And configured GPIOA/13 and GPIOA/14 on "hw_75_100.h" (from jaykup) as PAS pins.

Am i missing something? checked with oscilloscope, directly on port pins, and looks good (channel A on DIO, channel B on CLK).

Here is my repo:

https://github.com/rupebac/bldc/blob/release_5_03

Any idea ? thanks!

livello:
> JST

You can try UART2 pins.
image
SWDIO, SWDCLK pins are not connected to plugs on fsesc 75100.



> // UART Peripheral
> #define HW_UART_DEV				SD3
> #define HW_UART_GPIO_AF			GPIO_AF_USART3
> #define HW_UART_TX_PORT			GPIOB
> #define HW_UART_TX_PIN			10
> #define HW_UART_RX_PORT			GPIOB
> #define HW_UART_RX_PIN			11
> 
> // Permanent UART Peripheral (for NRF51) (jaykup) for UART2 port
> #define HW_UART_P_BAUD			115200
> #define HW_UART_P_DEV			SD4
> #define HW_UART_P_GPIO_AF		GPIO_AF_UART4
> #define HW_UART_P_TX_PORT		GPIOC
> #define HW_UART_P_TX_PIN		10
> #define HW_UART_P_RX_PORT		GPIOC
> #define HW_UART_P_RX_PIN		11

from here: 
https://forum.esk8.news/t/how-to-update-firmware-on-the-flipsky-75100-foc-esc/61819

rupebac:
There is as internal JST connector that you can read ono the PCB: DIO and CLK, but I am not 100% sure those are connected to the same pins that you guys are trying.

UART2 I am using it with the bluetooth module.

I think the issue is that I have to stop the UART in order to use RX and TX as input pins, right?

I think I have to change the enabled arg to false in this line:

image

Or use ADC1 and ADC2 pins (but I would prefer to have them free for a possible throttle or brake sensor).

Will continue tomorrow.

jon2com:
> 
> I think the issue is that I have to stop the UART in order to use RX and TX as input pins, right?
> 
 I am using the UART port for a display(or connect the bluetooth module when needed). The PAS sensor connected to the COMM port works anyway

SoftwareNood-2022:
Hi guys, I have a VESC 6. I'm wondering if this code can work also with it....
...regardless all, among other things I am a researcher in the energy sector, the only code I used is in the math field not control boards, I think many people have the same problem. Some weeks ago I had an idea, why not use the already supported AS5048A  to track perfectly the pedal rotation in the VESC ? There is a smart guy who programmed an Arduino to solve a similar problem, the video is this:
https://www.youtube.com/watch?v=hhSib8TQQps
he got a huge resolution, in theory the VESC has the hall sensor input, if you use a sensorless motor this slot is available to perform this task, moreover you colud program it for reverse braking as you are backward pedaling for instance...
Following his path you might have a perfect resolution PAS not a raw 12 or 24 pulse system ! What do you think about it ?
Sorry if I am intruding in your discussion, but I'd like to share with you this thought.
Thanks.

rupebac:
issue fixed.... I feel embarrassed to explain what the problem was :S I was pressing on VESC Tool on saving motor changes, instead of the other button for applying applications changes... 

Anyway.... 5.03 stock works perfectly now, with TX/RX (stock), as well as with ADC. I am using a converter to 3.3v.

stock PAS implementation in 5.03 runs fine, only I noticed that sometimes it won't stop and you need a new pulse from PAS to make it stop. But I think it is not a big deal.

@livello your implementation I played with it a bit, look here:

https://www.youtube.com/watch?v=T4I_XHKodhM

the startup is not as smooth as with the stock one, but it could be cos there is not enough current. Also, when it is going full, stop pedalling, and start again, the ramp up seems very aggressive and you can see how the whole support that I made shakes (last third of the video or so). All these things could just need a bit of adjustment. Will see.

TreadMTB:
Hi guys, Please bare with me being a beginner. I would like to connect the VESC to my Tongsheng TSDZ2, the phase wires and hall connectors are simple enough. How should I connect the torque sensor to the VESC pins for the stock firmware? The TSDZ2 has a torque sensor with only 2 wires, black and red. Where should I try to connect those? If I want to do throttle and PAS would I choose PAS + uart? or just select PAS and constant torque? I will eventually want to try livello's version after I figure out how to get this torque sensor working, so i could use the other SWDIO, SWDCLK pins for PAS instead and have comms for throttle and uart for bluetooth. TIA for the help.

jon2com:
> stock PAS implementation in 5.03 runs fine, only I noticed that sometimes it won't stop and you need a new pulse from PAS to make it stop. But I think it is not a big deal.

I have noticed the exact same behaviour

livello:
> How should I connect the torque sensor to the VESC pins for the stock firmware? The TSDZ2 has a torque sensor with only 2 wires, black and red. Where should I try to connect those? If I want to do throttle and PAS would I choose PAS + uart? 

The pas sensor of TSDZ is two hall based with 24 magnets. It has nice resolution. 
https://opensourceebikefirmware.bitbucket.io/development_tsdz2/About_Tongsheng_TSDZ2_mid_drive_motors--Torque_sensor_and_PAS.html


But torque sensor is very complicated. Also, there is no place to hide vesc controller inside motor. 

So why you need vesc on TSDZ2? For smooth ride? For more watts? 

The best and only way of using torque sensor on TSDZ is to keep stock controller and make him send data from torque sensor to vesc. But it is complicated. You must keep calibration feature...

TSDZ2 has his own open source firmware with a lot of functions.

livello:
> @livello your implementation I played with it a bit, look here:

Thank you very much for feed back!
Unfortunately I'm on my own legs since my shaft is broken. I try to repair and to print another one.

image
image

geofrancis:
does anyone have a link to a compatible quadrature PAS ?

rupebac:
> does anyone have a link to a compatible quadrature PAS ?

any one for Cycle Analyst V3 should work (there are some links in this thread, check my comments), just power up the PAS with 5V, and trust @jon2con on that 5v is fine on RX/TX lines, or use a 5v - 3.3v converter to step down voltage from the 2 signals of the pas.

rupebac:
> > @livello your implementation I played with it a bit, look here:
> 
> @rupebac Thank you very much for feed back! Unfortunately I'm on my own legs since my shaft is broken. I try to repair and to print another one.
> 
> image image

wow, how did it happen?

TreadMTB:
> > How should I connect the torque sensor to the VESC pins for the stock firmware? The TSDZ2 has a torque sensor with only 2 wires, black and red. Where should I try to connect those? If I want to do throttle and PAS would I choose PAS + uart?
> 
> @TreadMTB The pas sensor of TSDZ is two hall based with 24 magnets. It has nice resolution. https://opensourceebikefirmware.bitbucket.io/development_tsdz2/About_Tongsheng_TSDZ2_mid_drive_motors--Torque_sensor_and_PAS.html
> 
> But torque sensor is very complicated. Also, there is no place to hide vesc controller inside motor.
> 
> So why you need vesc on TSDZ2? For smooth ride? For more watts?
> 
> The best and only way of using torque sensor on TSDZ is to keep stock controller and make him send data from torque sensor to vesc. But it is complicated. You must keep calibration feature...
> 
> TSDZ2 has his own open source firmware with a lot of functions.

I was planning to put the controller external for lower heat in the system, I would love the ability to run 72v and throttle. VESC is easily set up with throttle, but I wouldn't want to give up the torque sensing PAS. My goal is to get a torque sensing system working with a mid drive set up for my ebike. TSDZ2 happens to be a system I already own, that I was hoping to test on. For me to test more thorougly I would need to order a torque sensing bottom bracket and hope I can get it to work. I don't have much experience  with the VESC. Doesn't seem to be too many people using it this way, but seeing as there is a constant torque setting under the PAS section of VESC Tool. Is that the mode I would need to use with a torque sensor?

TreadMTB:
> > does anyone have a link to a compatible quadrature PAS ?
> 
> any one for Cycle Analyst V3 should work (there are some links in this thread, check my comments), just power up the PAS with 5V, and trust @jon2con on that 5v is fine on RX/TX lines, or use a 5v - 3.3v converter to step down voltage from the 2 signals of the pas.

I think if you buy some of the original ones from china(be sure to read the specs, most but not all run off of 5v input), they are cheaper priced and also not modified to the higher voltage/price/longer cable by grin tech. Then you can also avoid having to buy that voltage converter in the first place. TLDR don't buy from grin unless you like paying more and having to buy something extra to wire in just to get it to working.

rupebac:
> > > does anyone have a link to a compatible quadrature PAS ?
> > 
> > 
> > any one for Cycle Analyst V3 should work (there are some links in this thread, check my comments), just power up the PAS with 5V, and trust @jon2con on that 5v is fine on RX/TX lines, or use a 5v - 3.3v converter to step down voltage from the 2 signals of the pas.
> 
> I think you should buy some of the original ones from china(be sure to read the specs, most but not all run off of 5v input), they are cheaper priced and also not modified to the higher voltage/price/longer cable by grin tech. Then you can also avoid having to buy that voltage converter in the first place. My statements apply to their torque sensors, I'm not sure about if they have modified the input voltage on their cadence sensors(although I'm sure they put a good mark up on them). TLDR don't buy torque sensors from grin unless you like paying more and having to buy something extra to wire in just to get it back to 5v input.

I haven't yet found a quadrature one from china, if you do please put the link. Grin uses these "Kingmeter". I was able to find one in Germany for 25€.

livello:
@TreadMTB
> How can I keep the TSDZ2 controller and have it send it's torque/cadence data to the vesc?
You can make cables longer. Two wires from torque sensor and four wires from hall cadense sensor. 
After that you will need to send data from tsdz2 controller to vesc. That could be obtained with nrf52840 chip ble-uart module of TSDZ2. You should add can bus module, you will need to modify tsdz2 nrf52840 code, and you will need create vesc can bus app for retrieve cadence and torque data... 

Lot of work. Lot of tests. Weeks, months, depends on your skill. 

If you have a lot of time and no money... Go, try. 

But much beteer is to buy bafang ultra with vesc-based controller or keep stock one...

chuyskywalker:
> It's not "supported" as far as I know, but I use these.
> https://www.aliexpress.com/item/33030138390.html

Sorry to ping an old thread -- are you just using the torque output signal directly to, say, ADC1?

thebrain76:
I have a Mini Vesc 4.20 with broken mini usb connector.
I only have access via the SWD PIN and an STLink adapter. The firmware flashing works great.
After flashing I can operate the Vescsoftware via Bluetooth over UART port.
A friend programmed me the signal line for a quadrature PAS on ADC and ADC2. This has also worked already.
Problem: when I am connected to controller on a device via Bluetooth, and I select PAS as control sensor, there is no selection PAS+UART. 
If I select only PAS, the communication is abruptly terminated. Logically, the BT adapter communicates via UART.

Actually it should run from the connections, if I select ADC and UART, only the software does not start the support for PAS although the connections fit. 

When the Miniusbbuchse still went, I could make the engine run via laptop with the newly assigned PIN ADC/ADC2 and the selection PAS.

thebrain76:
Can I get a connection from the Vesc tool without mini usb and without Bluetooth(UART), for example about the SWD Port?

007trains:
> 

By default the PAS app uses the UART pins for the quadrature sensor, this is the reason there is no PAS+UART app
To get around this you need to reassign the HW PAS pins (your friend has already done this)
And modify adc.c with the code to enable UART on the PAS app:

`case APP_PAS:
	app_pas_start(true);
	app_uartcomm_start(UART_PORT_COMM_HEADER);
	break;`

thebrain76:
Ok. I am not a professional. Please check my thinking .

In the first step my friend adjusted the hwconf/hw.h. The signal of the QuadraturPAS is now processed by the PIN ADC +ADC2. 

Then I would have to select the App Application "PAS" in the Vesctool. So that the connection via BT (UART) still exists another adjustment is necessary.

To get around this you need to reassign the HW PAS pins (your friend has already done this)
And modify adc.c with the code to enable UART on the PAS app:

case APP_PAS:
	app_pas_start(true);
        hw_stop_i2c();
	app_uartcomm_start(UART_PORT_COMM_HEADER);
	break;

After that the PAS works on ADC + ADC2 and the controller is still reachable via BT.

Remind: I need the BT connection because my mini USB socket is broken.

Thanks for your help

thebrain76:
changes for Quadrat PAS in hwconfig/hw.h:

[
59A31763-E3A7-44FA-8F81-A31020745F43
](url)

007trains:
> @007trains : thanks.
> 
> Ok. I am not a professional. Please check my thinking .
> 
> In the first step my friend adjusted the hwconf/hw.h. The signal of the QuadraturPAS is now processed by the PIN ADC +ADC2.
> 
> Then I would have to select the App Application "PAS" in the Vesctool. So that the connection via BT (UART) still exists another adjustment is necessary.
> 
> To get around this you need to reassign the HW PAS pins (your friend has already done this) And modify adc.c with the code to enable UART on the PAS app:
> 
> case APP_PAS: app_pas_start(true); hw_stop_i2c(); app_uartcomm_start(UART_PORT_COMM_HEADER); break;
> 
> After that the PAS works on ADC + ADC2 and the controller is still reachable via BT.
> 
> Remind: I need the BT connection because my mini USB socket is broken.
> 
> Thanks for your help

Yes your correct in your thinking, in order to get Bluetooth working the next step is to add the line:

`app_uartcomm_start(UART_PORT_COMM_HEADER);`

into adc.c under `case APP_PAS:`

That will start the UART when the PAS app is selected so you can use Bluetooth to connect to VESC while the PAS app is selected

ReFeXe:
3 wire pas
I did it!
It works correctly, detects the rotation of the pedals in the opposite direction, provides a smooth start
https://github.com/ReFeXe/bldclcd3/blob/release_6_02/applications/app_pas.c

ReFeXe:
mistakenly filled in the wrong file. fixed

livello:
> pas_one_magnet

what does it mean? 1 pulse per 1 round? Or 1 hall sensor in pas sensor?
Did you face with problem of too big cadence, more than max value? 
And pas_adc still does not work.

ReFeXe:
> > pas_one_magnet
> 
> what does it mean? 1 pulse per 1 round? Or 1 hall sensor in pas sensor? Did you face with problem of too big cadence, more than max value? And pas_adc still does not work.


this variable in my code, changes the operation of the controller to read a 3-wire sensor with 1 hall

Unfortunately, my bike is not ready yet, so I conduct experiments on the table.

rupebac:
guys, I tried at the beginning of the year, the stock fw 6.0 version, and it still had a very laggy stop (aka, it takes 1-2 secs to cut off power to the motor after stopping pedalling). Does anyone know if this has been fixed now in the master branch? is it worth for me to try again ?

leachim1:
@rupebac 
> issue fixed.... I feel embarrassed to explain what the problem was :S I was pressing on VESC Tool on saving motor changes, instead of the other button for applying applications changes...
> 
> Anyway.... 5.03 stock works perfectly now, with TX/RX (stock), as well as with ADC. I am using a converter to 3.3v.
> 
> stock PAS implementation in 5.03 runs fine, only I noticed that sometimes it won't stop and you need a new pulse from PAS to make it stop. But I think it is not a big deal.
> 
> @livello your implementation I played with it a bit, look here:
> 
> https://www.youtube.com/watch?v=F1SG49BhHj4
> 
> the startup is not as smooth as with the stock one, but it could be cos there is not enough current. Also, when it is going full, stop pedalling, and start again, the ramp up seems very aggressive and you can see how the whole support that I made shakes (last third of the video or so). All these things could just need a bit of adjustment. Will see.

I am really confused because i had all up and running with FW 5.02 and just updated to 5.03 no changes in the HW cabeling. And it just stops working. All settings the same.

If i downgrade to FW 5.02 and setup the same config all works as usual. I use a Quadrature sensor with 12 magnets.
Using a Flipsky 75100 75V 100A Single ESC Based.
Is there any difference in the Hardware Cabeling from 5.02 to 5.03?

Sandroforms:
> 3 wire pas I did it! It works correctly, detects the rotation of the pedals in the opposite direction, provides a smooth start https://github.com/ReFeXe/bldclcd3/blob/release_6_02/applications/app_pas.c
> 
> to work, just set a variable "pas_one_magnet" true

Could you tell me which input did you connect the 3 wires of the pas sensor to the vesc?

Tukk1S:
> > @007trains : thanks.
> > Ok. I am not a professional. Please check my thinking .
> > In the first step my friend adjusted the hwconf/hw.h. The signal of the QuadraturPAS is now processed by the PIN ADC +ADC2.
> > Then I would have to select the App Application "PAS" in the Vesctool. So that the connection via BT (UART) still exists another adjustment is necessary.
> > To get around this you need to reassign the HW PAS pins (your friend has already done this) And modify adc.c with the code to enable UART on the PAS app:
> > case APP_PAS: app_pas_start(true); hw_stop_i2c(); app_uartcomm_start(UART_PORT_COMM_HEADER); break;
> > After that the PAS works on ADC + ADC2 and the controller is still reachable via BT.
> > Remind: I need the BT connection because my mini USB socket is broken.
> > Thanks for your help
> 
> Yes your correct in your thinking, in order to get Bluetooth working the next step is to add the line:
> 
> `app_uartcomm_start(UART_PORT_COMM_HEADER);`
> 
> into adc.c under `case APP_PAS:`
> 
> That will start the UART when the PAS app is selected so you can use Bluetooth to connect to VESC while the PAS app is selected


 I cant find case APP_PAS, so im bit confused where to add 
`app_uartcomm_start(UART_PORT_COMM_HEADER);
could someone give more detailed instructions? 
Thank you! 

-TukkiS

Tukk1S:
> > > @007trains : thanks.
> > > Ok. I am not a professional. Please check my thinking .
> > > In the first step my friend adjusted the hwconf/hw.h. The signal of the QuadraturPAS is now processed by the PIN ADC +ADC2.
> > > Then I would have to select the App Application "PAS" in the Vesctool. So that the connection via BT (UART) still exists another adjustment is necessary.
> > > To get around this you need to reassign the HW PAS pins (your friend has already done this) And modify adc.c with the code to enable UART on the PAS app:
> > > case APP_PAS: app_pas_start(true); hw_stop_i2c(); app_uartcomm_start(UART_PORT_COMM_HEADER); break;
> > > After that the PAS works on ADC + ADC2 and the controller is still reachable via BT.
> > > Remind: I need the BT connection because my mini USB socket is broken.
> > > Thanks for your help
> > 
> > 
> > Yes your correct in your thinking, in order to get Bluetooth working the next step is to add the line:
> > `app_uartcomm_start(UART_PORT_COMM_HEADER);`
> > into adc.c under `case APP_PAS:`
> > That will start the UART when the PAS app is selected so you can use Bluetooth to connect to VESC while the PAS app is selected
> 
> I cant find case APP_PAS, so im bit confused where to add `app_uartcomm_start(UART_PORT_COMM_HEADER); could someone give more detailed instructions? Thank you!
> 
> -TukkiS

Found it! 
it's not in adc.c it is in app.c

PR #254, opened 2021-01-21
Match Focer hardware names to firmware directories

Mitchlol:
Trying to get the firmwares into the tool, not sure if anything else is wrong.

### DIFF
--- a/hwconf/hw_Cheap_FOCer_2.h
+++ b/hwconf/hw_Cheap_FOCer_2.h
@@ -20,7 +20,7 @@
 #ifndef HW_Cheap_FOCer_2
 #define HW_Cheap_FOCer_2
 
-#define HW_NAME					"Cheap FOCer 2"
+#define HW_NAME					"Cheap_FOCer_2"
 
 // HW properties
 #define HW_HAS_DRV8301
--- a/hwconf/hw_Little_FOCer.h
+++ b/hwconf/hw_Little_FOCer.h
@@ -22,7 +22,7 @@
 
 #include "drv8323s.h"
 
-#define HW_NAME                 "Little FOCer"
+#define HW_NAME                 "Little_FOCer"
 
 // HW properties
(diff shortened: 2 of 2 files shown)


hwconf/hw_Cheap_FOCer_2.h
```diff
@@ -20,7 +20,7 @@
 #ifndef HW_Cheap_FOCer_2
 #define HW_Cheap_FOCer_2
 
-#define HW_NAME					"Cheap FOCer 2"
+#define HW_NAME					"Cheap_FOCer_2"
```
b264, line 23: It'd be awesome to add `_v10` here or something so as not to mix it up with the popular version 0.9

hwconf/hw_Little_FOCer.h
```diff
@@ -22,7 +22,7 @@
 
 #include "drv8323s.h"
 
-#define HW_NAME                 "Little FOCer"
+#define HW_NAME                 "Little_FOCer"
```
b264, line 25: It'd also be great to add _v10 here or something so as not to mix it up with the popular version 0.9

Mitchlol: There is no v0.9 for the little focer.

hwconf/hw_Cheap_FOCer_2.h
Mitchlol: 0.9 was beta, this is the first release and will be the standard hardware going forward. It might be a little confusing for some people now, but this is the cleanest path forward.

FWIW this firmware wont overwrite firmware on a beta unit using the tool. You'd have to build it from scratch and flash with an st-linker. Additionally there is a migration guide in the main repo for making old hardware compatible with this firmware.

Lastly flashing this firmware on a beta unit just lowers your current values by half but works fine, not a huge risk or danger.

PR #255, opened 2021-01-29
Added feedforward term to PID control

jac93:
Uses kv term measured experimentally


vedderb:
That looks a bit arbitrary to me, and is only implemented for BLDC. There is no config option for it either. Closing this for now. Please open another PR if you want to revisit this.

PR #271, opened 2021-03-27
- Add driver for SX1278 LoRa boards, eg. HopeRF RFM95W (driver by Woj…

CTSchorsch:
…ciech Domski, https://github.com/wdomski/SX1278)

- Add a simple bool isInitialized() to commands.c, to check if commands are initialzied
- Add hwconf for GESC, hobby open source ESC (https://github.com/CTSchorsch/gesc) including a simple double pulse test function
- Add app_dpv.c to read out a simple I2C magnetic field sensor and two hall sensors to control a underwater scooter


vedderb:
This is a good start, thanks for the PR! Once my RFM95-modules arrive I will start testing a bit too. It would be nice to make it possible to connect them to the comm-port too.

Here are a few comments about the code. I will give some detailed comments on the files too.

* Try to keep the indentation correct and consistent. Use tabs and not spaces.
* conf_general.h now always includes the custom app and the version number is decreased. By default no custom app should be selected. Best would be to only add your hardware files to conf_general.h and leave the rest unchanged.
* Run the rebuild_all script in the build_all directory to ensure that the changes are compatible with all hardware versions. If that script fails the problems need to be fixed to make everything build.
* Instead of modifying app.c to start the uartcomm app after the custom app, you should start the uartcomm app from the custom app itself if you need to have it running. There are several custom apps that use the same pins, so uartcomm should not be running by default.
* It is probably better to put the SX1278-files in their own directory with a mk-file, as with the imu code.

commands.c
```diff
@@ -1424,18 +1431,6 @@ void commands_apply_mcconf_hw_limits(mc_configuration *mcconf) {
     }
 
 #ifndef DISABLE_HW_LIMITS
-
```
vedderb, line 1427: This comment should still be there.

hwconf/rfm95w.c
```diff
@@ -0,0 +1,123 @@
...
+
+static THD_FUNCTION(packet_process_thread, arg) {
+	(void)arg;
+
+    unsigned char buffer[128];
+	chRegSetThreadName("rfm95w proc");
+
+    buffer[0]=COMM_GET_VALUES;
+
+    while(!commands_is_initialized() )
```
vedderb, line 106: Use a sleep here while waiting. Otherwise this thread will take all CPU resources and block threads with lower priority.

hwconf/hw_gesc.c
```diff
@@ -0,0 +1,492 @@
...
+    TIM_Cmd(TIM4, ENABLE);
+    TIM_Cmd(TIM4, DISABLE);
+    TIM1->ARR = (breaktime+pulse2)*utick;
+    TIM1->CCR1 = breaktime*utick;
+    while ( TIM1->CNT != 0 );
+    TIM_Cmd(TIM4, ENABLE);
+
+    chThdSleepMilliseconds(1);
+    TIM_CtrlPWMOutputs(TIM1, DISABLE);
+    mc_configuration *mcconf = mempools_alloc_mcconf();
```
vedderb, line 469: You need to free the configuration when you are done, otherwise each call to this terminal command will make the memory pool run out of configurations. That is what is called a memory leak.

hwconf/rfm95w.c
CTSchorsch: argl, yes, i'm wondering why sometimes after a few tests, the systems stops.

hwconf/hw_gesc.c
CTSchorsch: argl, yes, you are right. i'm wondering why sometimes the system stops after a few tests

commands.c
CTSchorsch: uh, do i change something there ? that was an accident

hwconf/hw_gesc.c
vedderb: Actually, the functions below expect the configuration to persist (they will store a pointer to it), so it is better to give them the pointer directly from mc_interface. There is probably a better wait to do this in general, maybe I should add something to mc_interface...

CTSchorsch:
> This is a good start, thanks for the PR! Once my RFM95-modules arrive I will start testing a bit too. It would be nice to make it possible to connect them to the comm-port too.
Mh, i know that there are devices which are ready to connect to the uart. the rfm95 has only spi. But it should be possible to integrate the modules in your command/packets system
 
> Here are a few comments about the code. I will give some detailed comments on the files too.
> 
>     * Try to keep the indentation correct and consistent. Use tabs and not spaces.
> 
>     * conf_general.h now always includes the custom app and the version number is decreased. By default no custom app should be selected. Best would be to only add your hardware files to conf_general.h and leave the rest unchanged.
> 
>     * Run the rebuild_all script in the build_all directory to ensure that the changes are compatible with all hardware versions. If that script fails the problems need to be fixed to make everything build.
> 
>     * Instead of modifying app.c to start the uartcomm app after the custom app, you should start the uartcomm app from the custom app itself if you need to have it running. There are several custom apps that use the same pins, so uartcomm should not be running by default.
> 
>     * It is probably better to put the SX1278-files in their own directory with a mk-file, as with the imu code.

I will take care. thanks

hwconf/hw_gesc.c
CTSchorsch: mmh, i tried to use just a pointer without the malloc. this did not work, but i don't know anymore why.

conf_general.h
```diff
@@ -190,7 +196,8 @@
 /*
  * Select default user motor configuration
  */
-//#include			"conf_test.h"
+#include			"mcconf_default.h"
```
vedderb, line 199: The default configuration should not be selected from here. It will be selected later if it isn't overridden here or somewhere else.

conf_general.h
```diff
@@ -209,7 +216,7 @@
 
 // CAN-plotter
 //#define APP_CUSTOM_TO_USE			"app_plot_can.c"
-//#define APPCONF_APP_TO_USE			APP_CUSTOM
+#define APPCONF_APP_TO_USE			APP_NONE
```
vedderb, line 219: APP_NONE should not be selected by default. In most cases it is selected by the hardware file or default config, where UART is the default most of the time.

applications/app.c
```diff
@@ -88,13 +88,13 @@ void app_set_configuration(app_configuration *conf) {
 		break;
 
 	case APP_PPM_UART:
-		hw_stop_i2c();
+        hw_stop_i2c();
```
vedderb, line 91: Still spaces

commands.c
```diff
@@ -75,12 +75,18 @@ static mutex_t print_mutex;
 static mutex_t send_buffer_mutex;
 static mutex_t terminal_mutex;
 static volatile int fw_version_sent_cnt = 0;
+static bool isInitialized = false;
+
+bool commands_is_initialized(void) {
+    return isInitialized;
```
vedderb, line 81: Spaces

applications/app.c
CTSchorsch: Do you know a good tool, to check the indent ?

commands.c
```diff
@@ -75,12 +75,18 @@ static mutex_t print_mutex;
...
+bool commands_is_initialized(void) {
+    return isInitialized; 
+}
 
 void commands_init(void) {
 	chMtxObjectInit(&print_mutex);
 	chMtxObjectInit(&send_buffer_mutex);
 	chMtxObjectInit(&terminal_mutex);
 	chThdCreateStatic(blocking_thread_wa, sizeof(blocking_thread_wa), NORMALPRIO, blocking_thread, NULL);
+    isInitialized = true;
```
vedderb, line 89: Spaces

applications/app.c
```diff
@@ -129,7 +129,6 @@ void app_set_configuration(app_configuration *conf) {
 
 	case APP_CUSTOM:
 #ifdef APP_CUSTOM_TO_USE
-		hw_stop_i2c();
```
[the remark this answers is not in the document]
CTSchorsch: Ah, i need i2c for my custom dpv_app.  will insert again

vedderb:
I have added a few comments. It would be easy for me to fix most of them myself in another commit, but I thought it is worth mentioning what the purpose is of some of the parts to avoid confusion later.

CTSchorsch:
> I have added a few comments. It would be easy for me to fix most of them myself in another commit, but I thought it is worth mentioning what the purpose is of some of the parts to avoid confusion later.

Yes thanks, this kind of cooperation is still new for me and think i need more discipline ;)

applications/app.c
vedderb: there are still spaces :-)

vedderb:
I can fix the spaces. Can you remove the binary files from the PR? They make the git-folder so large over time, so I just compile and upload them when I merge back to master.

CTSchorsch:
> I can fix the spaces. Can you remove the binary files from the PR? They make the git-folder so large over time, so I just compile and upload them when I merge back to master.

Is it okay for you, to add build_all/**/*.bin to .gitignore ?

vedderb:
I would prefer to leave the build in the repo, so that the latest built version of the master firmware can be downloaded from there. It would be good if you just can remove them from the last commit in the PR.

vedderb:
Now the PR deletes the files from the main repo instead. What I meant is to not include them in the PR at all and just leave them as they are in the main repository.

CTSchorsch:
argl, i don't really understand git... 
you mean: rebuild_all, but do not add the bin files to the commit ?

vedderb:
Yes. you should pick the files for each commit manually. Use rebuild_all to make sure that everything builds without warnings and errors, pick the files you want to include with git add and then make a commit. You can also split the changed files into separate commits when it makes sense to make it easier to follow the history.

CTSchorsch:
Ok, but now, i have to add the directories and files to the commit to bring them back ?

vedderb:
I think you have to remove the commits entirely any add them again with just the relevant files. Not sure how to do that though without losing the changes, I'm not that fluent in git either. You could copy the entire repository somewhere else and try a few things there until you figure it out. A quick internet search gave me this site:
https://dev.to/bholmesdev/how-to-revert-your-git-commits-without-panicking-4d1c

CTSchorsch:
I can do a rebase and go back to 2fe6c13 and redo the small changes made after that commit ?

vedderb:
Merged!

I also got my 3 RFM95-boards. What are you using on the other end?

CTSchorsch:
Do you think it is possible to modify the package which is send over LoRa ? It would be nice to transmit the vesc-id as address in the first bytes. The LoRa Chip has some kind of address filtering but the node address has to be the first byte in payload

PR #273, opened 2021-03-28
Fix external IMU

Mitchlol:
Some recent changes to the UART init function cause the I2C to stop functioning (previously it was the start function that stole the pins). So I rewrote the app initialization code to only initialize apps when they run. Certainly not the only solution, and I didn't test every app, so hopefully i got it all right 😅 (I was most unsure about nrf/nunchuck/uart and where the lines are drawn)

*I also fixed the indentation and did some other tiny cleanups

### DIFF
--- a/applications/app.c
+++ b/applications/app.c
@@ -55,6 +55,9 @@ void app_set_configuration(app_configuration *conf) {
 	app_nunchuk_stop();
 	app_balance_stop();
 	app_pas_stop();
+#ifdef APP_CUSTOM_TO_USE
+	app_custom_stop();
+#endif
 
 	if (!conf_general_permanent_nrf_found) {
 		nrf_driver_stop();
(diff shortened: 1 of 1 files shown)


vedderb:
One problem with this approach is that app_nunchuk (which should be renamed to vesc remote) is used for the wand and vesc remotes even via UART and CAN. With this change it won't be configured in most cases, which will be quite confusing when changing configuration in VESC Tool. This is where data comes from in some cases:

https://github.com/vedderb/bldc/blob/dev_fw_5_03/commands.c#L658

This is maybe not the best approach, but it is how things ended up in the end when I tried to reuse stuff.

I think it is better to always configure all apps regardless if they are running, and make sure that the configuration function doesn't do anything unexpected.

Mitchlol:
It looks like @Jfriesen222 moved the pin configuration from start to init. Without know why he did it, I don't feel comfortable reversing it.

https://github.com/vedderb/bldc/commit/e862f29298ec16640bb9e1ab40e5486cfb168c5a

vedderb:
Can you check if this helps?
https://github.com/vedderb/bldc/commit/69d3a1ba33888279c4d6346c15cf09aab581b675

Mitchlol:
All good now, closing this Pull Request.

PR #281, opened 2021-04-10
mc_pwm_foc: simplified math in current PI antiwindup

maxicor88:
In collaboration with https://github.com/vedderb/bldc/pull/91 and https://github.com/vedderb/bldc/commit/feb1be5fbd6aa9916c1ce0ad918afe70c1b55fec

This just avoids doing 2 "SIGN()" and "fabsf()" operations. 

Mathematically the end result in the operation is the same, but use less cpu cycles.


vedderb:
Nice, thanks! Does this approach work as well as the one you sent me before? From my limited testing so far it runs a bit more stable.

maxicor88:
Thanks for merging it. Yes, we have been using this antiwindup code with Axiom boards for around half a year. Non issues detected with it so far.

PR #282, opened 2021-04-11
bemf_decoupling: improved math for IPM motors.

maxicor88:
-In IPM motors (ld_lq_diff > 0) Ld affects in the Q axis, and Lq affects in the D axis.

### DIFF
--- a/mcpwm_foc.c
+++ b/mcpwm_foc.c
@@ -3637,8 +3637,15 @@ static void control_current(volatile motor_all_state_t *motor, float dt) {
 			break;
 
 		case FOC_CC_DECOUPLING_CROSS_BEMF:
-			dec_vd = state_m->iq_filter * state_m->speed_rad_s * conf_now->foc_motor_l * (3.0 / 2.0);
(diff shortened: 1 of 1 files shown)


vedderb:
Are you sure that this is correct, or is it a guess? I can imagine that something like this is correct for salient motors, but it would be good to have some test that confirms it.

maxicor88:
Yes, we tested this code in both the Nissan Leaf motors we use in our dyno setup for about 9 months so far (with up to 175uH of LdLqdiff) and also in other 750W rated motors (with LdLqdiff of 30uHy). 

You may also see that the math used is the same that supports the virtual motor algorithm. 

I tried to look up quickly for a paper that shows this, but could not find it. I found this in the video of Texas Instruments ("Teaching old motors new tricks series, part 3) https://www.youtube.com/watch?v=5eQyoVMz1dY&t=3956s

When he goes through the math, he uses a PSMS motor (so he uses Ls, not Ld nor Lq), but later when he goes to the example, you can see that they show Ld affecting Vq and Lq affeting Vd. (I put the link when he shows the diagram of the example)

If I come across the paper that shows this in a proper manner, I will post it here.

PR #286, opened 2021-04-12
Balance-app: allow jump-dismount at low speed, using half_fault delay

surfdado:
When below the half-fault erpm, we check the full-fault ADC switch timer
also against the half_fault delay value

Before this change, a long full_fault delay could lead to undesired wheel
spinning after jumping off the board, even when stationary.

The other advantage of this change, is that users with only a single sensor
now have the ability to set different delays for speeds above half_fault erpm
and below half_fault erpm.

If "Half Switch Fault Delay" is 0 (=default) then this new behavior is disabled.

### DIFF
--- a/applications/app_balance.c
+++ b/applications/app_balance.c
@@ -197,8 +197,9 @@ bool check_faults(bool ignoreTimers){
 			state = FAULT_SWITCH_FULL;
 			return true;
 		}
-		// at low speed (below half-fault threshold speed), also check half_fault delay:
+		// at low speed (below half-fault threshold speed), also check half_fault delay (if non-zero):
 		else if ((abs_erpm < balance_conf.fault_adc_half_erpm)
+				 && (balance_conf.fault_delay_switch_half > 0)
 				 && (ST2MS(current_time - fault_switch_timer) > balance_conf.fault_delay_switch_half)){
 			state = FAULT_SWITCH_FULL;
 			return true;
--- a/applications/app_balance.c
+++ b/applications/app_balance.c
@@ -192,7 +192,14 @@ bool check_faults(bool ignoreTimers){
 	// Check switch
 	// Switch fully open
 	if(switch_state == OFF){
+		// at any speed, always check full_fault delay:
 		if(ST2MS(current_time - fault_switch_timer) > balance_conf.fault_delay_switch_full || ignoreTimers){
+			state = FAULT_SWITCH_FULL;
+			return true;
+		}
+		// at low speed (below half-fault threshold speed), also check half_fault delay:
+		else if ((abs_erpm < balance_conf.fault_adc_half_erpm)
+				 && (ST2MS(current_time - fault_switch_timer) > balance_conf.fault_delay_switch_half)){
 			state = FAULT_SWITCH_FULL;
 			return true;
 		}


surfdado:
I added a fixup so that if "Half Switch Fault Delay" is 0 (=default) then this new behavior is disabled.

    -added check for fault_delay_switch_half to be non-zero, to avoid unintended
     behavior for single sensor users

vedderb:
Can @Mitchlol have a look at this PR? I consider him responsible for the balance app :-) If he thinks it looks ok I will merge it.

surfdado:
I’ve already pinged him 😊

mattbutlar:
@surfdado couldn't you just add an `|| switch_state == OFF` to the existing half switch state (ie `if((switch_state == HALF || switch_state == OFF) && abs_erpm < balance_conf.fault_adc_half_erpm)`), so it runs the reg half switch logic?  That would accomplish what you are wanting and is tech what it should have always been seeing as a full foot off is also a half foot off and it would tech meet both conditions and then it should be a matter of letting them complete normally based on their settings.

surfdado:
@mattbutlar what you suggest would work too, but then you'd be throwing a HALF-FAULT instead of a FULL-FAULT, which however wouldn't really have any downside from what I can tell. I'm open to either approach

surfdado:
@mattbutlar the other reason I chose my approach is because I use a very low erpm for half-speed faults, but didn't want my board to go nuts for a full second if I jump off at 1.5 or 2mph - so in my own implementation I actually use half-speed-erpm x 4 as the threshold for low-speed vs regular-speed. But that didn't seem like something I could submit to the main repo. The correct way of doing this would be to add more config parameters but I thought in the first version I'd keep it simple.

mattbutlar:
@surfdado right, I get why you want the change and it makes sense as that is similar to what FM does from what I can tell.  My recommendation would do what you want with minimal changes and is really what it should be... then I think adding an independent erpm timeout scale value would be good.  I thought about adding a setting to scale the full foot timeout based on the erpm, so you can set it to like 2000 erpm or something and it will scale from 0ms to your full timeout ms from 0 erpm to 2000 erpm.  That way the slower you are going the quicker it will shut off but when going faster it is longer and less likely to be a death sentence.

Mitchlol:
> @surfdado couldn't you just add an `|| switch_state == OFF` to the existing half switch state (ie `if((switch_state == HALF || switch_state == OFF) && abs_erpm < balance_conf.fault_adc_half_erpm)`), so it runs the reg half switch logic? That would accomplish what you are wanting and is tech what it should have always been, seeing as a full foot off is also a half foot off and it would tech meet both conditions and then it should be a matter of letting them complete normally based on their settings.

I actually like this solution a little better. To me it makes logical sense that if both switches are off, then one switch is off. I guess you could call this a bug. TBH I kinda wondered why jumping off at low speed sucked, but i never put 2 and 2 together.

If it's OK with you @surfdado I'll throw the @mattbutlar implementation on my branch, as I've got a few changes to work through. 

Off topic: I dunno about scaling values, that seems like overkill. But something I have wanted to do for a while was have the switches be sticky even when balancing is off, to make mounting easier.

surfdado:
> If it's OK with you @surfdado I'll throw the @mattbutlar implementation on my branch, as I've got a few changes to work through.

Sure, that's fine with me.

> Off topic: I dunno about scaling values, that seems like overkill. But something I have wanted to do for a while was have the switches be sticky even when balancing is off, to make mounting easier.

I have no idea what you're talking about, must be an EUC thing, I have no idea what your mounting process even looks like...

Mitchlol:
Alright I put in my pull request with matt's suggestion, so i think this one can be closed.

vedderb:
I'm closing this one then.

PR #288, opened 2021-04-14
Introduce adjusted input voltage to componensate for R tolerances

surfdado:
On some VESCs the reported input voltage has a noticeable error.
A correction factor allows users to adjust the measured value.

This feature lets users specify a correction factor, expressed as a
percentage. Now the input voltage used everywhere is Vin * (1 + corr).

The correction factor is limited to values in the range of -10% .. +10%

There's a matching commit in vesc_tool

### DIFF
--- a/applications/app_balance.c
+++ b/applications/app_balance.c
@@ -255,7 +255,7 @@ void calculate_setpoint_target(void){
 		}
 		setpointAdjustmentType = TILTBACK;
 		state = RUNNING_TILTBACK_DUTY;
-	}else if(abs_duty_cycle > 0.05 && GET_INPUT_VOLTAGE() > balance_conf.tiltback_high_voltage){
+	}else if(abs_duty_cycle > 0.05 && mc_interface_get_adjusted_input_voltage() > balance_conf.tiltback_high_voltage){
 		if(erpm > 0){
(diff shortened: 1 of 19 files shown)


vedderb:
I'm a bit concerned introducing a slower input voltage method in the already heavy interrupts. Also, the old mcpwm code relies on the input voltage being scaled in a similar way to how the phase voltage dividers scale it. The new phase filter implementation also kind of does that (I have to study the code carefully again and figure out what the effect is).

I think the best way to update the input voltage reading and handling is having one low-pass filtered version in mc_interface that is updated in the ISR-callback. The 1 kHz timer can then create corrected versions for each motor. Then I have to go through the uses one by one and see which one is most suitable to use.

surfdado:
No worries - If you can find a safe way to add such a feature then please do. I've personally never had an issue with voltages being slightly off, but both my CheapFOCer and my LittleFOCer have fairly small errors. Others, especially BalancePro users have reported errors in excess of 1 Volt. But obviously there's always the workaround of manually adjusting voltage limits accordingly...

I understand you don't want me to take further action on this, right?

surfdado:
Should we just close this? I personally don't feel too strongly about this feature - I'd say we should simply consider this a hardware quality control issue, VESC builders should make sure they use precise resistors...

PR #289, opened 2021-04-17
Axiom field weakening 503

maxicor88:
This code introduces an alternative method for implementing field weakening, based on a PI regulator, often seen in papers about this matter. 

It uses "Field Weakening Duty Start", as the reference input of the PI block, generating an error signal once the duty goes above this threshold. The regulator will increase the "fw_current" up to the limit imposed by "Field Weakening Current Max", in order to try to get this error to 0.  

The Kp and Ki gain variables are set as 0.0 in default configuration, and this will allow the controller to use the ramp method and disables the PI method introduced in this PR. 

Once the user configures a value in Kp or Ki, the PI regulator takes place as the source of field weakening current value.

It has been tested unloaded in Axiom board, the behaviour was pretty similar to the ramp method. We will keep testing both methods in different scenarios. 

A similar branch will also be uploaded to vesc tool repository, to add this pull request variables in the GUI.


vedderb:
This will not work for dual motors. The fw state should be part of the motor state.

Maybe im missing something, but l don't see the main difference from the ramping here. Ramping even works well for me when I set the ramp time to 0, which is like having infinite gain. I can see the use case for a pi controller if the input is e.g. the difference between the set and measured q axis current, as that would only apply fw when needed in order to reach the set curret.

It would be good to have some tests and example that shows where this has advantages before introducing it.

maxicor88:
Thanks for your feedback!

Yes, I pushed the code here so you could take a look and give me your input. I totally agree that this does not justify to be added if it does not show better performance. I will let you know how our tests go.

Cheers

vedderb:
Have you done some more testing? I'm curious to hear how it goes. Let me know if there is anything I can help with.

maxicor88:
We have run a few tests a couple of weeks ago and results were pretty similar in both algorithms. However we still did not push it very hard with the Dyno. The team got busy with side projects but we are going to do more tests the incoming days.

nitrousnrg:
At Luna we have been running Benjamin's field weakening for a couple of weeks on 30+ bikes without any issue. For reference the drives are set for 150 phase Amps and 30A of max field weakening

maxicor88:
We pushed a little bit harder in the dyno and started having some issues with the duty representation in Vesc, that might be also affecting performance.

I explained this issue here, as it is affecting both Field Weakening methods.

https://github.com/vedderb/bldc/pull/91#issuecomment-889542946

PR #297, opened 2021-05-06
Balance App: Autosuspend

surfdado:
@Mitchlol 

Forgetting to turn your board off while in balance mode can be
treacherous when brake current is set. The high power draw can
cause a fully charged battery to drain in a matter of a few days.

Use the SHUTDOWN MODE configuration to automatically stop the
app.

### DIFF
--- a/applications/app_balance.c
+++ b/applications/app_balance.c
@@ -19,6 +19,7 @@
 
 #include "conf_general.h"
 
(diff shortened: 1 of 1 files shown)


Mitchlol:
Hey, can you close the PR as I've implemented an alternative.

PR #299, opened 2021-05-18
Balance 4.3

Mitchlol:
Set current off delay
Add high pass filter
Nose angling refactor


surfdado:
Hi @Mitchlol - would you mind revising this commit so that you reuse the existing kd_pt1_frequency parameter for the low-pass filter? 
Right now the latest beta37 breaks compatibility with old configurations, requiring manual setting of the new parameters even when a user doesn't want to change his settings (yet)

PR #313, opened 2021-06-28
Balance App: introduce new speed based variable tiltback

surfdado:
Nose angle adjustment, similar in purpose as constant tiltback
but it is proportional to erpm aka speed.

@Mitchlol


surfdado:
@Mitchlol said:
Mitch Lustig, [30.06.21 12:11]
Also i looked at your code 2 days ago and had some suggestions but then was too lazy to type em up. 😴

Mitch Lustig, [30.06.21 12:12]
Main one was that if the signs don't match, just flip the sign on the limiter one, and roll with the sign of the angling one. I can't really think of any situations where this would backfire.

surfdado:
Implemented @Mitchlol's change request

applications/app_balance.c
```diff
@@ -419,13 +423,18 @@ static void calculate_setpoint_interpolated(void){
 }
 
 static void apply_constanttilt(void){
-	// Nose angle adjustment
+	// Nose angle adjustment, add variable then constant tiltback
+	float constanttilt_target = 0;
+	if (fabsf(erpm) > tiltback_variable_max_erpm) {
+		constanttilt_target = tiltback_variable * tiltback_variable_max_erpm * SIGN(erpm);
```
Mitchlol, line 429: This line can be simplified to 
`constanttilt_target = balance_conf.tiltback_variable_max * SIGN(erpm);`

applications/app_balance.c
```diff
@@ -419,13 +423,18 @@ static void calculate_setpoint_interpolated(void){
 }
 
 static void apply_constanttilt(void){
```
Mitchlol, line 425: Since you are now doing 2 kinds of adjustment in this function perhaps we can rename it to apply_noseangling or something like that.

Same for the float constanttilt_target variable as it's now the sum of the 2 tilts.

applications/app_balance.c
surfdado: Good suggestion! But remember what we did with the sign earlier? So we'd need to do fabsf of that, right?

applications/app_balance.c
surfdado: I was thinking that too, but didn't want to rename your stuff without your input though. Will do

surfdado:
Requested changes have been pushed

Mitchlol:
Tested on a vehicle and it works for me, also used the graph rendering to monitor it's inputs/outputs and they looked correct as well.

The code looks nice and neat as well.

@vedderb I give this the "OK" signal to be merged.

surfdado:
Thanks guys! Btw, my fixup commits don't look like they've been squashed - do you not do --autosquash prior to merge? Or is this something I should have done? Sorry, I'm not very experienced with the whole pull request process...

PR #315, opened 2021-07-06
Balance App - biquad filters: add config/reset functions

surfdado:
To make the new biquad filter more convenient to use for other purposes I suggest to add config and reset functions too. 
@Mitchlol


vedderb:
@Mitchlol
Can you have a look if this was ok? I thought it was from you so I just accepted the PR, and realized it wasn't a second after.

Mitchlol:
It's probably fine 👍

PR #317, opened 2021-07-10
Fix switching to the MT6816 encoder before reboot

bsilver8192:
encoder_init_mt6816_spi only made it into the switch statement at
startup, so after changing the mc_conf it required a reboot of the VESC
before it worked.


bsilver8192:
This is a fix for #238 for reference.

nitrousnrg:
Thanks Brian, I couldn't figure out how to fix that in the original PR and its obvious now.

Could you change this PR so it patches the development branch instead of the master branch?

bsilver8192:
Done

PR #322, opened 2021-07-18
Make triggering ppm safestart on motor fault configurable.

### LINKED ISSUE #321: PPM median filter causes glitch on startup without safe_start
The Vesc is used heavily in battlebots robots. 
They generally use ppm servo input, and have to turn safe start off because you often get a fault when you get hit, and the time it takes for ppm safe start to recover is not acceptable.
You get faults, but generally do not reboot.

This means you pick up the glitch from median filter equalizing on startup because you can't use safe start.
This causes the motors to twitch on a very big and dangerous robot.

There are probably several ways to solve this:
- Have vesc tool option to not trigger safe_start on motor fault (extra condition in app_ppm.c, line 167)
- When ppm app starts, throw out the first few readings. regardless of safe start setting while the filter equalizes.
- Init the median filter with the first good pulse.

Thoughts? I can probably implement the "don't trigger safe start on motor fault" option.

Teslafly:
Closes #321 (or at least provides a suitable workaround)

Adds option to ignore motor faults in triggering safe_Start. 

Useful in applications like battlebots where you have to recover quickly from a motor fault.
You can't wait for safe start to recover from the fault, but still want safe start when you first power on.

Let me know if I missed anything, but it seems to work.

Further testing in progress for merge.

Vesc tool parameter add PR
https://github.com/vedderb/vesc_tool/pull/150


Teslafly:
I fixed the appconf signature and have now adequately tested this. It works as expected with the companion vesc_tool MR. 

Inverted the logic to make the vesc tool name more straightforward.

Should be ready to merge or have a discussion on.

vedderb:
Can you see if this is what you had in mind?
https://github.com/vedderb/bldc/commit/fc6825a1aefdc9d59379be09cf099ae42a8b875b
I made it this way to save some bytes in the config and to have a similar option for the ADC app. This also allows adding more safe start modes easily in the future.

Teslafly:
> Can you see if this is what you had in mind?

Yes. That also works to accomplish this feature. 

If I was smart I probably would have done it this way too. But was probably too wary of touching existing parameters.

Thank you. I'll close this and the related vesc tool MR.

PR #323, opened 2021-07-19
Add CAN-bus triggered - User-programmed onboard motion sequence

supercrazysam:
**Introduction:**

This pull request introduce a new way for external devices to trigger VESC's onboard user-programmed motion sequence, via CANbus.  

This will allow users to program, script simple motion sequences inside VESC in a "Arduino-like" style,   
utilizing the powerful native `mc_interface.c`, native variables inside VESC firmware, and trigger the programmed sequence by one single simple CAN bus command. 

I have used and tested this simple modification on multiple VESC robot projects over the past two years, it is reliable and I feel like this is a feature/function that every hobbyist will enjoy.

(UART version coming next)

==========
**Motivation:**

The typical way of using VESC, is to use a "main" controller (either microcontroller or computer or some sort of input device), to stream control commands to one/many VESC.

However, in a system, many small functions can be off-loaded to VESC(s), like some pre-defined sequences:

(Simple linear motion:                    ` move to pos 1000, wait 10s, move to pos 0`)
(Simple quadrupled robot joint:      `move to pos 0,  pos 90,  pos 0, pos 90` )
(Current-based limit switch:            `move until current draw >= 10A, set that position as pos 0` )
(Close a door using motor              `rotate at 600rpm, then stop until it closes ~ current draw>=10A` )
(A water pump that can stop itself  `rotate at 600rpm, stop after 1 hour` )         
(....Or a mix of any functions mentioned above, anything in mc_interface can be used!)

Under this pull request, users can directly program and customize the behavior of the VESC when triggered by a specific CAN bus command, to perform different tasks.  

This way, by off-loading simple pre-defined tasks to slave VESC(s), the code on the main controller can be much cleaner.
And if the task logic is simple enough, it is might even be possible to program the entire sequence in VESC!

==========
**Mechanics and Behavior:**

It does NOT disable normal VESC operation.  After the user-programmed code sequence completed, you can use it just like a normal VESC. 

Even when the programmed sequence is running, other input (CAN bus/ ADC/ UART/ PPM...) will still have effect, although you will be fighting for control in this case.  Higher frequency usually wins, although it depends on what kind of code you programmed inside the sequence.

When the sequence is running, it wont be triggered again by the CAN bus command, it can only be triggered after the sequence is completed (When there are no sequence running). 

==========
**Videos:**
  
Demonstration of the pull request's code:  

I did the modification on the latest dev branch, and tested it to make sure everything works.
(my commentary might not be on point) 
[Demo here](http://www.youtube.com/watch?v=fn-P8Sq_9w8 "Demo Here")  

-----
I tested this functions on multiple of my projects,   
below are some of the smaller project / hobbyist task examples:  (Some larger projects detail are non-disclosable)

VESC onboard user-programmed - motor rotation sequence:  
[Demo here](http://www.youtube.com/watch?v=voFFHYZu0LE "Demo Here")    

-----
VESC onboard user-programmed - crossbow object launcher loading sequence:  
[Demo here](http://www.youtube.com/watch?v=pkdx50GuoWQ "Demo Here")  

==========
**Reminder**

Remember to put `chThdSleepMilliseconds`, otherwise the programmed code might risk stucking in a "bricked" manner and might need SWD Debugger to re-flash it.

Remember to `timeout_reset();`

==========
**File changed:**
- app_custom.c
- comm_can.c
- conf_general.h
- datatypes.h

==========

Again, I would like to thank Benjamin (@vedderb)  for creating this wonderful motor controller.

Sam


vedderb:
The app_custom.c file is supposed to just be a template to be used as a basis for custom apps, so it should not include any extra functionality. It is better to make the example under a new name. It should also not be active by default in conf_general.h.

It is already possible to achieve this be adding a callback for standard or extended can-frames in the custom app with comm_can_set_sid_rx_callback or comm_can_set_eid_rx_callback. Here is one exaple:


```
static bool can_sid_callback(uint32_t id, uint8_t *data, uint8_t len);

// In the init-function:
comm_can_set_sid_rx_callback(can_sid_callback);

// The callback itself, which should return true if a valid command was accepted
static bool can_sid_callback(uint32_t id, uint8_t *data, uint8_t len) {

}
```
Then you can even define multiple sequences that are activated with different commands.

Let me know if you have any questions about this approach, I can try to help you make a working example.

PR #327, opened 2021-07-30
Balance: configurable speed/angle for each tiltback type

surfdado:
Using the same tiltback speed/angle for duty as well as high voltage / low voltage tiltback can be dangerous. Let users have different speeds for different events.


surfdado:
Look for matching vesc_tool commit.

@Mitchlol

surfdado:
This commit doesn't include the SIGNATURE change in confgenerator!

vedderb:
One problem with the balance app is that the configuration starts to become a bit large, and soon the appconf will exceed the maximum packet size. I can increase that, but if possible I would like to keep the configuration as small as possible.

1) Are there some parameters that can be removed or merged? That should also make it easier for new users to not be overwhelmed by all the parameters.

2) Can the transfer size of some parameters be reduced? (e.g. by using double16 and scaling the parameter accordingly)
image

surfdado:
Wait... there's a parameter editor GUI? OMG that's awesome! LOL
Yes, many of the Balance parameters could indeed be reduced to double16.

surfdado:
I've converted all tiltback speed and angle parameters to 16bit doubles now

Issue #324, opened 2021-07-20
Startup music tune function


supercrazysam:
As mentioned in the title.

Many commercial BLDC ESCs produce a start-up tune.
Have anyone tried this on VESC? Would like to know more on this.  

Probably will implement it once figured out.

Sam.

surfdado:
VESCs don't have any built-in beeper/buzzer so this is not possible without a hardware change or the addition of external modules

Addy771:
@surfdado it's possible without any hardware changes, only software. Modulating the PWM frequency would work, think about the sounds VESC makes during motor detection or the beeps that hobby ESCs make using the motor.

surfdado:
The sounds I hear on my drone and other ESC based devices come from the motor? Not the ESC? I had no idea

supercrazysam:
> The sounds I hear on my drone and other ESC based devices come from the motor? Not the ESC? I had no idea

I think so.. just not quite exactly sure how they do it, I saw people use stepper motor+driver to produce music tunes, there are some people did that with bldc motors+driver as well

supercrazysam:
My guess is controlling the motor to produce vibration frequency in audible range??
Does anyone have a idea how they actually do it?

arpan0193:
Maybe add some soundwave signal to the reference signal?

hexakopter:
> There was a video I saw a couple years ago where someone had programmed the ESC for their scooter's brushless motor to play music as it powered the scooter along.

I think you are talking about this video: https://www.youtube.com/watch?v=7BQ9lei_24Y 
You can also read a little bit about it on Shanes website: http://scolton.blogspot.com/2010/12/midi-scooter.html

There was also a request on the vesc-project website for a "beep" implementation a few years back. (https://vesc-project.com/node/647) Maybe with the recent HFI implementation some fundamentals are build into the VESC code that can be used to utilize the motor as a speaker more easy.

supercrazysam:
Guys I found out how to do it. Just set the switching frequency to (audible frequencies /2).

Currently writing a prototype to test out the beeping function. :D

supercrazysam:
Submitted a pull request on this, it works really well. Unfortunately I don't have my phone right now, I will upload the demo video once I return home.

supercrazysam:
Here is the video (using a small drone motor):
[...]

PR #328, opened 2021-08-06
Added play music tune function

supercrazysam:
**Motivation:**  

Most of the commercial ESCs on the market, especially drone ESCs, have the ability to produce a tune when powered or in fault state. This is achieved by running the motor in different switching frequencies that is audible to human ears, basically using the motor as a speaker.

This pull request is an simple inital attempt to re-create the same effect in VESC, producing tunes by running the motor in different switching frequencies.

==========
**Mechanics:**  

This pull request contains an example for playing tunes, it uses custom application that plays two tunes, one after another. The first one is a startup tune sequence, the second is a error/warning tune sequence tune.

In the next version, the function will be added to the startup function, to play the startup tune when powered on.

Update:
Tested on 6" inch scooter direct drive hub motor and a small drone motor, with new parameters.

==========
**Video Demos:**  

Demo Video (small drone motor - running 3S):
[Demo here](http://www.youtube.com/watch?v=LB0OpRd8YMk "Demo Here")    

Demo Video (6" inch direct drive scooter motor - runnning 13S):
[Demo here](http://www.youtube.com/watch?v=zjQ2rIybZyA "Demo Here")

=====
Old version demos:  

More explaination [ old version was used in this video, still moves a bit, please refer to new version video above] :  

[Demo here](http://www.youtube.com/watch?v=RGsPfEyQ-WU "Demo Here")  

[Old version, still moves a bit, please refer to new version video above] <del>Demo Video :</del>
[Demo here](http://www.youtube.com/watch?v=pQnYOhD0_OI "Demo Here")    

==========
**Related issue:**  

#324 
https://vesc-project.com/node/647

Sam

### DIFF
--- a/applications/app_play_tune.c
+++ b/applications/app_play_tune.c
@@ -68,7 +68,7 @@ static THD_FUNCTION(custom_script_thread, arg)
                 //play the startup tune
                 for( int a = 3000; a < 5000; a = a + 500 ) {
                      change_sw(a);
-                     mc_interface_set_pid_speed(300.0);
+                     mc_interface_set_pid_speed(10.0);
                      chThdSleepMilliseconds(300);
                      timeout_reset();
                 }
@@ -82,7 +82,7 @@ static THD_FUNCTION(custom_script_thread, arg)
                 for( int i = 0; i < 3; i++ ) {
                      change_sw(5000);
                      chThdSleepMilliseconds(300);
-                     mc_interface_set_pid_speed(300.0);
+                     mc_interface_set_pid_speed(10.0);
                      chThdSleepMilliseconds(500);
                      timeout_reset();
                 }
--- a/applications/app_play_tune.c
+++ b/applications/app_play_tune.c
@@ -60,7 +60,7 @@ static THD_FUNCTION(custom_script_thread, arg)
 		is_running = true;
 		(void)arg;
 	 
-		chRegSetThreadName("APP_CUSTOM_SCRIPT");
+		chRegSetThreadName("APP_PLAY_TUNES");
  
                 //save original switching frequency
                 float original_sw = mc_interface_get_configuration()->foc_f_sw;
--- a/applications/app_play_tune.c
+++ b/applications/app_play_tune.c
@@ -1,0 +1,101 @@
+/*
+	Copyright 2017 Benjamin Vedder	benjamin@vedder.se
+        Copyright 2021 Sam Shum         cshum@andrew.cmu.edu
+
+	This file is part of the VESC firmware.
+
+	The VESC firmware is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    The VESC firmware is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+    */
+
+#include "ch.h" // ChibiOS
+#include "hal.h" // ChibiOS HAL
+#include "mc_interface.h" // Motor control functions
+#include "hw.h" // Pin mapping on this hardware
+#include "timeout.h" // To reset the timeout
+
+#include "mcpwm_foc.h" // Motor control functions
+#include "timer.h" // Motor control functions
+
+static volatile bool stop_now = true;
+static volatile bool is_running = false;
+// custom script thread
+static THD_FUNCTION(custom_script_thread, arg);
+static THD_WORKING_AREA(custom_script_thread_wa, 2048); // 2kb stack for this thread
+
+
+void app_custom_start(void) {
+        if (!is_running)
+        {
+                stop_now = false;
+                is_running = true;
+		chThdCreateStatic(custom_script_thread_wa, sizeof(custom_script_thread_wa), NORMALPRIO, custom_script_thread, NULL);
+        }
+}
+
+void app_custom_stop(void) {
+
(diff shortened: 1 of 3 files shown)


surfdado:
Very interesting - but you set PID speed to 1000rpm to play the tune - won't that be running the motor? On an esk8, how much distance does your board cover by the time it has completed the tune?? I'm a little scared to try this on my Onewheel... waiting for your video...

supercrazysam:
> Very interesting - but you set PID speed to 1000rpm to play the tune - won't that be running the motor? On an esk8, how much distance does your board cover by the time it has completed the tune?? I'm a little scared to try this on my Onewheel... waiting for your video...

Yes, it will run the motor, at a really really low speed for a very short period of time.

This is erpm 1000, divided by pole pairs then it is the real rpm, I set it to 1000 because the default vesc minimum erpm is 900. 
Also it runs for only couple hundred ms, we can further decrease it if you are really concerned.

surfdado:
One a OneWheel motor 1000 erpm comes out to about 3km/h - but for us the minimum erpm is much lower, pretty sure 100 should work too

surfdado:
Also, I guess we could reverse direction after each note to keep board travel to a minimum?

supercrazysam:
> Also, I guess we could reverse direction after each note to keep board travel to a minimum?

Good point.

The reason for it to be 1000 erpm is because I am worried about this part: (And the default min_erpm is 900)
https://github.com/vedderb/bldc/blob/be439a2a55ae5b1ba8cce8ad6b0da2d80466e9df/mcpwm_foc.c#L3749-L3753

But yeah, if it is not a concern I will definitely make it really low.

surfdado:
You're right for PID control there does seem to be a minimum of 900 erpm, even in my own configuration. But what's stopping us from lowering that limit, at least temporarily? Otherwise 300ms at 1000 erpm would definitely travel too much on a big hub motor like we use in Onewheels... I guess I could test it on the bench at first to be safe

supercrazysam:
> You're right for PID control there does seem to be a minimum of 900 erpm, even in my own configuration. But what's stopping us from lowering that limit, at least temporarily? Otherwise 300ms at 1000 erpm would definitely travel too much on a big hub motor like we use in Onewheels... I guess I could test it on the bench at first to be safe

Yeah I guess you are right, safety first.  
After all people that are testing with this feature will probably know the 900 erpm default limit thing anyway :D .

surfdado:
Haha, that was crazy - definitely not a pleasant experience. Motor jerked back and forth that the noise totally overpowered the switching frequency. Never trying that again...

There must be a better solution. During motor and hall sensor detection the motor moves in a super gentle manner. I've definitely heard the motor make noises without rotating, e.g. during R/L/flux detection.

supercrazysam:
> Haha, that was crazy - definitely not a pleasant experience. Motor jerked back and forth that the noise totally overpowered the switching frequency. Never trying that again...
> 
> There must be a better solution. During motor and hall sensor detection the motor moves in a super gentle manner. I've definitely heard the motor make noises without rotating, e.g. during R/L/flux detection.

It worked perfectly with a drone motor, now heading home, will upload the video.

surfdado:
Haha a heavy hub motor is probably the opposite extreme on the motor spectrum. There's no gearbox btw, it's just direct drive.

supercrazysam:
> Haha a heavy hub motor is probably the opposite extreme on the motor spectrum. There's no gearbox btw, it's just direct drive.

Can you upload the model of the motor you are using? A big heavy hub motor for onewheel or ebike, usually have some sort of thin plantery gear inside.

surfdado:
It's a phub188 based "Funwheel" motor - you can see it what it looks like inside: https://youtu.be/4k391oJAeMs

supercrazysam:
> It's a phub188 based "Funwheel" motor - you can see it what it looks like inside: https://youtu.be/4k391oJAeMs

I see, in that case I will also test with a middle size scooter hub motor to test its limits.

(Maybe this is the reason why this sort of play tune function is not present in large motor controllers? )

supercrazysam:
> Haha a heavy hub motor is probably the opposite extreme on the motor spectrum. There's no gearbox btw, it's just direct drive.

Should have mentioned that in the pull request, sorry about that. :(

supercrazysam:
> Haha a heavy hub motor is probably the opposite extreme on the motor spectrum. There's no gearbox btw, it's just direct drive.

This is the demo video:
[Demo here](http://www.youtube.com/watch?v=pQnYOhD0_OI "Demo Here")

surfdado:
Your video shows that the motor rotates quite a bit. I feel like that's not what the ESCs are doing. They produce the noise but the rotors just wiggle a bit, they don't spin.

supercrazysam:
> Your video shows that the motor rotates quite a bit. I feel like that's not what the ESCs are doing. They produce the noise but the rotors just wiggle a bit, they don't spin.

This is the old demo video I took yesterday using 1000 erpm.

supercrazysam:
> 
> 
> Your video shows that the motor rotates quite a bit. I feel like that's not what the ESCs are doing. They produce the noise but the rotors just wiggle a bit, they don't spin.

This is the updated performance, using erpm speed 10

[Demo here](http://www.youtube.com/watch?v=LB0OpRd8YMk "Demo Here")

supercrazysam:
> 
> 
> Haha a heavy hub motor is probably the opposite extreme on the motor spectrum. There's no gearbox btw, it's just direct drive.

I ran it with my 6" inch scooter direct drive hub motor, here is the result. It does not spin.
[Demo here](http://www.youtube.com/watch?v=zjQ2rIybZyA"Demo Here")

surfdado:
For commit 0505c7 you will probably want to follow the mcpwm_foc naming convention and make the whitespace consistent with the rest of the file.
I also noticed several whitespace errors in your app file.

supercrazysam:
Fixed, thank you for pointing out. :D

vedderb:
See my comment here:
https://github.com/vedderb/bldc/pull/330#issuecomment-898302596

PR #330, opened 2021-08-08
Added startup tune

supercrazysam:
==========  
**Although there are already some results, this pull request is still in-progress**   

The only issue right now: VESC tool sometimes become a bit more difficult to connect to. Currently solving.

==========
Based on the pull request #328 I made earlier, I was able to made a VESC startup tune prototype.
This is the result. 
[Demo here](http://www.youtube.com/watch?v=eFeOw0AzH2c "Demo Here")
(I tested and it worked on both small drone motor and my 6" direct drive scooter hub motor.)

I will wait for Benjamin to review/approve #328 first, make necessary changes (if any), then submit the completed version of this startup tune feature in this pull request. 

Also, since "startup tune" will be a general feature, I will do more testing on it first.

==========
- Added config options to enable / disable the startup tune.  
- Does not require user to manually change the min_erpm for it to work, it will lower the min_erpm temporaily and revert back to original value after playing tunes.  
==========
Related issues:
#324 
https://vesc-project.com/node/647

### DIFF
--- a/conf_general.h
+++ b/conf_general.h
@@ -254,6 +254,11 @@
 #endif
 
 /*
+ * Enable startup tune 
+ */
+#define STATUP_TUNE_ENABLE				1    //comment to disable
+
+/*
  * Settings for the external LEDs (hardcoded for now)
  */
 #define LED_EXT_BATT_LOW			28.0
--- a/main.c
+++ b/main.c
@@ -226,7 +226,26 @@ int main(void) {
 
 	ledpwm_init();
 	mc_interface_init();
+#ifdef STATUP_TUNE_ENABLE
+        if (mc_interface_get_configuration()->motor_type==MOTOR_TYPE_FOC) {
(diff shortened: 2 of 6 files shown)


vedderb:
One problem with this approach is that the timer creates glitches when changing the frequency in center-aligned mode. This is in the errata sheet and is the reason that I enforced the switching frequency to be fixed as long as the motor is running. The problem with these glitches is that in the worst case they can cause long pulses where the output is on, which can destroy the power stage with low inductance motors (ask me how I know).

It making the motor beep is important there are a few better approaches.

1) Reconfigure the PWM pins to GPIO and run a code sequence that toggles them to make beeps. Needs some care too, but is by far the easiest.

2) Run the control loop as usual and modulate audio on top of the switching (similar to how HFI is implemented). This is a bit trickier and will take some extra CPU cycles, but allows audio modulation at the same time as the motor is running and even polyphonic or sampled audio.

3) Configure the timer in non-center-aligned mode (as in BLDC) and change the switching frequency similar to how you implemented it.

PR #333, opened 2021-08-19
Tiltback config cleanup

Mitchlol:
Little bit of cleanup for the tiltback config, I wanted to make all the names, variables, defaults, etc uniform


surfdado:
Please let's not change names of existing parameters just for "cosmetics" reasons. Just like in beta38 (pt1 filter got renamed!!) this will cause lots of frustration and confused users because their config suddenly doesn't work when they switch between versions. 

Feel free to change the new parameters I introduced as you see fit, but existing names should ideally be considered "don't touch" imho

Mitchlol:
I knew you would raise this complaint, but I want to keep a clean codebase.

surfdado:
Can't we still achieve a clean codebase within the firmware while also preserving existing parameter names?

surfdado:
I noticed you upped all defaults to 3deg/sec - if we do use non-zero defaults then don't we want to suggest the most reasonable defaults? 

The primary reason why the additional tiltback variants were added in the first place was because high-voltage tiltback is very unpleasant when tilting quickly. So instead of using 3deg/sec for all 3 tiltback variants, we might be better off with 5deg/sec for duty (like in the past), 2deg/sec for HV and then any number between 2 and 3 for LV tiltback, I'd suggest 3 but I have no strong preference an LV tiltback speed.

surfdado:
@vedderb - what is the general VESC Tool/firmware policy on backward compatibility? When users upgrade from 5.2 to 5.3 is it expected that configurations from 5.2 should still work?

vedderb:
The backwards compatibility with configs here is up to you, as you know how people use the app. I usually say that if the old config and firmware works for you there is no reason to upgrade, and if you upgrade you should redo the configuration anyway to get the most out of the new version. That is true for the motor config, which is quite easy to make using the wizards, but I can imagine that the balance app is quite difficult to set up.

It also depends on the type parameter. If these parameters have good default values and most people don't change them I would argue that you might as well change to a more intuitive name as that won't cause problems for most people. It might even help them if they if they never changed them in the old config and the new default values are better.

surfdado:
Okay, in that case I'd say let's get this change included before the next beta version is put out.

PR #341, opened 2021-09-18
Balance App to propagate DUTY faults to MC Interface

surfdado:
When a balance duty fault occurs (duty cycle > configured limit) the consequence is a nosedive in 99% of cases! 
Furthermore it is usually difficult to prove that this happened even when logs are available.

Propagating this fault over to mc_interface allows duty cycle related problems to show up in logs and get reported by apps.

Too many onewheel builders are getting unexplained nosedives, especially the guys with weak batteries (Floatwheel 12s1p!!)

I realize this may not be the right way to do this, but I wanted to get the conversation started at least

@Mitchlol


applications/app_balance.c
```diff
@@ -787,6 +787,8 @@ static THD_FUNCTION(balance_thread, arg) {
 				// Otherwise duty fault will clear itself as soon as motor pauses, then motor will spool up again.
 				// Rendering this fault useless.
 				check_faults(true);
+				mc_interface_fault_stop(FAULT_CODE_OVER_DUTY, false, false);
```
Mitchlol, line 790: The loop runs continuously during a fault, that will spam this function. Is that intended/required? Perhaps it would be better to call this function alongside the detection and setting the state to FAULT_DUTY

surfdado:
I now realize that we don't need this fault at all, sounds like we should set duty fault to 100% and we won't need to worry about it

PR #342, opened 2021-09-25
Support for Spintend ubox single

tharon-rivera:
Would like to include this 75/300 based controller with internal IMU. There are a few balance board builders using this currently.

### DIFF
--- a/build_all/rebuild_all
+++ b/build_all/rebuild_all
@@ -566,3 +566,21 @@ touch conf_general.h
 make -j8 build_args='-DHW_SOURCE=\"hw_raiden7.c\" -DHW_HEADER=\"hw_raiden7.h\"' USE_VERBOSE_COMPILE=no
 cd $DIR
 cp $FWPATH/build/BLDC_4_ChibiOS.bin $COPYDIR/VESC_default.bin
+
+#################### UBOX_SINGLE ########################
+
+COPYDIR=UBOX_SINGLE
+mkdir -p $COPYDIR
+rm -f $COPYDIR/*
+
+# default
+cd $FWPATH
+touch conf_general.h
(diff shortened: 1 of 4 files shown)


surfdado:
This commit appears to be incomplete. For UBox Single to be included you also need the rebuild_all script to be updated. I suggest you look at the git history to see how other boards were added in the past.

surfdado:
Sorry, ignore my previous comment

surfdado:
Yes, I'm sorry - you're right. Hence I removed my comment. My apologies. Your commit looks good!

luiscarone:
Hi @tharon-rivera do you know if this firmware will work with the ubox 75 300 dual? Thanks

surfdado:
> Hi @tharon-rivera do you know if this firmware will work with the ubox 75 300 dual? Thanks

I'm afraid this commit is exclusively for the Spintend Ubox Single, like the subject states.

When I looked at Spintend's code it looks like they made a royal mess out of the original 75_300 driver rather than adding a new proper driver following the same pattern of other VESCs.  What was done in this pull request is to create a compliant HW support for Ubox Single. The same would need to be done for Spintend's other VESCs.

tharon-rivera:
@luiscarone Like @surfdado said, the other ubox boards are a little different.  The major difference except for enforcing max value limits I saw for the dual options was a different V_REG.  On the current 75-200 dual page on their site, it says "we recommend to use Spintend custom 100A firmware, if you're using stock firmware, there's battery voltage reading discrepancy and relative symptom."  The single also has internal IMU configured that isn't present in the duals.

surfdado:
> @surfdado, the mess you mean is that they change the default 75/300 instead of creating a new one? or there is something else that I'm missing?

Yes, that's what I meant. They just hacked their way to a working firrmware without any regard for upstreaming their code, hoping that one of their customers will do it for them...

PR #345, opened 2021-10-02
Add A50S hardware. Correct resistance measurement.

TechAUmNu:
Adds A50S.h/c and build script.

Resistance measurement corrected to use phase-phase resistance /2. Rather than 2/3. 
This gives much better results on all motors and hardware tested. 
I am not sure why it was 2/3 to begin with. For example the TI datasheet https://www.ti.com/lit/ds/symlink/mcf8316a.pdf?ts=1633169249754 for MCF8316A on page 43 states: 

image

In addition the tooltip for motor resistance in VESC Tool also says it should be half. 

image

The phase is also changed to give better detection on dual shunt hardware. The measured voltage, current and calculated resistance are printed for information.

### DIFF
--- a/build_all/rebuild_all
+++ b/build_all/rebuild_all
@@ -277,6 +277,30 @@ make -j8 build_args='-DDISABLE_HW_LIMITS -DHW_SOURCE=\"hw_hd75.c\" -DHW_HEADER=\
 cd $DIR
 cp $FWPATH/build/BLDC_4_ChibiOS.bin $COPYDIR/VESC_default_no_hw_limits.bin
 
+#################### HW A50S 6S ########################
+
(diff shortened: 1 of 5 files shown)


TechAUmNu:
>  Did you leave the debug print in the resistance measurement code on purpose? 
Yes I think it is useful information and no reason not to show it in the console.

>  Also, don't you think these two commits warrant two separate pull requests as they are completely unrelated?
Yeah I wanted to do that, but GitHub won't let me do a pull request with specific commits. I could make another branch I guess and do a separate pull for each item.

vedderb:
Thanks for looking into this! I have commented about it here:
https://vesc-project.com/node/3225#comment-8281

PR #352, opened 2021-10-08
Use convenience macros for deg/rad conversions

kubark42:
This cleans up the code, improves readability, and reduces the odds of a math typo.

I'm not super in love with `RPM2RADPS_f` to mean RPM to radians/second (and the vice-versa `RADPS2RPM_f`), if there's a better suggestion I'd be happy to update the PR.

THIS IS UNTESTED

Old:
```
 275200	   3196	 140900	 419296	  665e0	build/BLDC_4_ChibiOS.elf
```

New:
```
 274960	   3196	 140900	 419056	  664f0	build/BLDC_4_ChibiOS.elf
```

I think the memory savings are coming from the compiler having an easier time optimizing, as well as avoiding accidental promotion to `double` during certain operations.


mcpwm_foc.c
```diff
@@ -3928,8 +3928,8 @@ static float correct_hall(float angle, float dt, volatile motor_all_state_t *mot
 	motor->m_hall_dt_diff_now += dt;
 
 	float rad_per_sec = (M_PI / 3.0) / motor->m_hall_dt_diff_last;
-	float rpm_abs_fast = fabsf(motor->m_speed_est_fast / ((2.0 * M_PI) / 60.0));
-	float rpm_abs_hall = fabsf(rad_per_sec / ((2.0 * M_PI) / 60.0));
```
kubark42, line 3932: Note that this was a division by `RPM2RADPS`, so I turned it into a simple multiplication of `RADPS2RPM`.

utils.h
```diff
@@ -81,6 +81,13 @@ const char* utils_hw_type_to_string(HW_TYPE hw);
 #define UTILS_IS_NAN(x)		((x) != (x))
 #define UTILS_NAN_ZERO(x)	(x = UTILS_IS_NAN(x) ? 0.0 : x)
 
+// Handy conversions for radians/degrees and RPM/radians-per-second
+#define DEG2RAD_f ((float)(M_PI / 180.0))
+#define RAD2DEG_f ((float)(180.0 / M_PI))
+#define RPM2RADPS_f ((float)((2.0 * M_PI) / 60.0))
+#define RADPS2RPM_f ((float)(60.0 / (2.0*M_PI)))
```
kalvdans, line 88:
```suggestion
#define RADPS2RPM_f ((float)(60.0 / (2.0 * M_PI)))
```

kubark42:
> All conversions looks to be equivalent to what was before. I haven't tested for real though.

Thanks. I think they are all correct as well, I've proof-read them a bunch of times. However, I don't have any way to test on hardware, so it's kind of dangling a little.

Unit tests would be useful here, but barring if enough eyes pedantically check every line I think we can trust it.

vedderb:
This is one of the cases where I lean towards just spelling out what is done rather than having a macro that I have to look up in a separate file. Should generate similar code after optimization. My opinion is not that strong though, so we can discuss why one or the other is better.

kubark42:
> `#define RADPS2RPM_f(rad_per_sec) (rad_per_sec * (float)(60.0 / (2.0 * M_PI)))`
> 
> If you change them to that style and change branch from master to dev_fw_5_03 I will merge the PR :)

Thanks! I was having a lot of fun with https://github.com/vedderb/bldc/pull/358, so am only now getting to this to resolve the merge conflicts. I will get it done tomorrow.

kubark42:
Done.

virtual_motor.c
```diff
@@ -214,7 +214,7 @@ static void connect_virtual_motor(float ml , float J, float Vbus){
 																							GET_GATE_DRIVER_SUPPLY_VOLTAGE();
 		}
 #endif
-		virtual_motor.phi = mcpwm_foc_get_phase() * M_PI / 180.0;
+		virtual_motor.phi = DEG2RAD_f(mcpwm_foc_get_phase());
```
kalvdans, line 217: One reason to write macros with CAPITALS, is to be aware that the argument could be evaulated twice. That would be bad if the argument evaluation is slow or have side-effects. Don't know if we should care about that here and use a temporary variable, or if we should use inline functions instead of macros. It is fine with me the current way though, just saying the history why macros are written CAPITALIZED.

See for example `UTILS_IS_NAN(x)` and `UTILS_NAN_ZERO(x)` that evaluates their argument twice.

utils.h
```diff
@@ -81,6 +81,15 @@ const char* utils_hw_type_to_string(HW_TYPE hw);
 #define UTILS_IS_NAN(x)		((x) != (x))
 #define UTILS_NAN_ZERO(x)	(x = UTILS_IS_NAN(x) ? 0.0 : x)
 
+// Handy conversions for radians/degrees and RPM/radians-per-second
+#define DEG2RAD_f(deg) (deg * (float)(M_PI / 180.0))
```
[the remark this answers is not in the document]
kubark42: Ooof, good catch!

virtual_motor.c
kubark42: You mean `DEG2RAD_f(DEG2RAD_f(x))` as a risk?

This makes somewhat of an argument for returning to pure multiplication. There are no sharp edges to `#define DEG2RAD_f (180.0f/M_PI)`.

vedderb: I still prefer the macros with arguments over multiplications. It is easier to see how to use them and it is less likely to make mistakes with parentheses when using them compared to multiplications. These macros are not evaluating arguments twice either, so that is a general statement about macros that sometimes is true but not here.

kubark42: I'm not one to let the perfect get in the way of the good. If you're happy with it and it tests out okay, I think it's ready for merge.

PR #353, opened 2021-10-08
Create Unit Test scaffold, via rewriting the Makefile

kubark42:
This PR adds unit test capabilities to the project. I'm hoping that they will be helpful in assuring firmware stability, especially as people seek to optimize or otherwise improve algorithms. 

This is a somewhat ambitious PR, as a lot of plumbing had to be done in order to make it all silky smooth. It replaces the existing `Makefile` with a wholly new version, one which is built around managing the project instead of building the firmware. The original Makefile is split, with some parts remaining in the new version and some pushed into `make/fw.mk`.

The default for `make` is now to pull up the `help` guide.
* `make`  # displays help
* `make help`  # does the same thing as `make`

The new way to build firmware is documented in the `README.md` changes, but in short:

* `make fw`  # bog-standard build, equivalent the the prior Makefile's `make` command
* `make fw -j8`  # to use a lot more cores
* `PROJECT=MyNewVESC make fw`  # to build a project by the name of `MyNewVESC`

This new Makefile concept also allows for automagic installation of development tools. This is nice because it means that different versions of the tools no longer clobber each other, and the user doesn't have to export any environmental variables.

* `make armk_sdk_install`  # installs the ARM SDK into a local directory
* `make gtest_install`          # installs the Google Test package into a local directory

It also allows for running unit tests
* `make all_ut_run`  # build and run all unit tests
* `make ut_utils`  # builds and runs the `utils` unit test

This is tested to build things correctly, but no firmware blobs are tested on actual hardware.

Here is the output of a unit test. Interestingly, it already found an unexpected result in the `utils_norm_angle_rad()` and `utils_norm_angle()` functions!
```
vesc % make all_ut_run                                         
*NOTE*      Sanitized disallowed variable 'TMPDIR' from environment
*NOTE*     Parallel make disabled by all_ut_run target so we have sane console output
utils.c
tests/utils/unittest.cpp
tools/gtest-1.11.0/src/gtest_main.cc
tools/gtest-1.11.0/src/gtest-all.cc
[1]  + done       gitk bob
build/unit_tests/utils/utils.elf
 TEST RUN    build/unit_tests/utils/utils.elf
Running main() from /Users/kenz/Documents/vesc/tools/gtest-1.11.0/src/gtest_main.cc
[==========] Running 6 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 3 tests from NormAngle_deg
[ RUN      ] NormAngle_deg.ValInRange
[       OK ] NormAngle_deg.ValInRange (0 ms)
[ RUN      ] NormAngle_deg.ValBelowRange
[       OK ] NormAngle_deg.ValBelowRange (0 ms)
[ RUN      ] NormAngle_deg.ValAboveRange
[       OK ] NormAngle_deg.ValAboveRange (0 ms)
[----------] 3 tests from NormAngle_deg (0 ms total)

[----------] 3 tests from NormAngle_rad
[ RUN      ] NormAngle_rad.ValInRange
unittest.cpp:118: Failure
The difference between 3.14159265358979323846264338327950288 and inputVal is 6.2831851561837873, which exceeds eps, where
3.14159265358979323846264338327950288 evaluates to 3.1415926535897931,
inputVal evaluates to -3.1415925025939941, and
eps evaluates to 9.9999999747524271e-07.
[  FAILED  ] NormAngle_rad.ValInRange (0 ms)
[ RUN      ] NormAngle_rad.ValBelowRange
unittest.cpp:142: Failure
Expected equality of these values:
  -3.14159265358979323846264338327950288
    Which is: -3.14159
  inputVal
    Which is: 3.14159
[  FAILED  ] NormAngle_rad.ValBelowRange (0 ms)
[ RUN      ] NormAngle_rad.ValAboveRange
[       OK ] NormAngle_rad.ValAboveRange (0 ms)
[----------] 3 tests from NormAngle_rad (0 ms total)

[----------] Global test environment tear-down
[==========] 6 tests from 2 test suites ran. (0 ms total)
[  PASSED  ] 4 tests.
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] NormAngle_rad.ValInRange
[  FAILED  ] NormAngle_rad.ValBelowRange

 2 FAILED TESTS
make[1]: *** [run] Error 1
make: *** [ut_utils_run] Error 2
```

In order to be truly valuable, the unit test will have to be extended to cover many more functions, but now there is a nice framework for it.

I expect that this major a rewrite of how people are used to doing business will take some thought and some back-and-forth in order to iron out any issues which might arise. I've tested it as much as I can on my end, but that's nowhere near enough coverage to know that this could be rolled out without breaking other people's setups.

The PR could be split into a more atomic approach, where the Makefile is changed just far enough to test existing workflows, and then once that's proven out we can add the Unit Test and SDK install capabilities back in. I put them all in at once in order to make a more compelling case that this Makefile encompassing vision was more than half-baked.


Makefile
```diff
@@ -1,314 +1,202 @@
...
-# If enabled, this option allows to compile the application in THUMB mode.
-ifeq ($(USE_THUMB),)
-  USE_THUMB = yes
-endif
-
-# Enable this if you want to see the full log while compiling.
-ifeq ($(USE_VERBOSE_COMPILE),)
-  USE_VERBOSE_COMPILE = yes
+# Makefile for Rootloader project
+.DEFAULT_GOAL := help
```
kalvdans, line 2: Remove; make runs the first defined target if none is given on the command line.

make/fw.mk
```diff
@@ -0,0 +1,286 @@
...
+ifeq ($(USE_OPT),)
+  USE_OPT = -O2 -ggdb -fomit-frame-pointer -falign-functions=16 -std=gnu99 -D_GNU_SOURCE
+  USE_OPT += -DBOARD_OTG_NOVBUSSENS $(build_args)
+  USE_OPT += -fsingle-precision-constant -Wdouble-promotion -specs=nosys.specs
+endif
+
+# C specific options here (added to USE_OPT).
+ifeq ($(USE_COPT),)
+  USE_COPT =
+endif
```
kalvdans, line 16: Make variables can be overridden without `ifeq` noise, all variables defined on the command-line takes precedence over the ones defined in the Makefile.

make/system-id.mk
```diff
@@ -0,0 +1,47 @@
+# shared.mk
```
kalvdans, line 1: remove copy-pasted comment.

kalvdans (commented):
Great with unit tests!

I don't like recursive make files though, and also I don't like macros, they will increase the barrier level of people that want to contribute. I will see if I have time to help out to eliminate those.

If people like the verbosity clutter I'm fine with that. I usually tell people to run `make -s` if they don't want to see the commands that are being run :)

Using google test framework is maybe a bit overkill but I'm fine with that.

kubark42:
> Great with unit tests!

Thanks! They're coming in handy as I try to implement http://www.coranac.com/2009/07/sines/, which is a response-- and an extension-- of the algorithm which is currently used by `utils_fast_sincos_better()`.

> I don't like recursive make files though, and also I don't like macros, they will increase the barrier level of people that want to contribute. I will see if I have time to help out to eliminate those.

There's definitely no perfect build system! I have based this one off a generic scaffold I learned at another project. It worked exceptionally well for me, hiding all the dev environment and build complexity when I was just starting off. At the time, I was wholly unable to understand what was going on in the background and was ecstatic to be able to jump right in and start modifying the codebase.

My experience is that when the Makefile is well done like this, it becomes something people never interface with beside typing `make ...`. For instance, it can track all the targets so the user can know, via `make help`, exactly what commands can be issued. In the below example, it shows that there are four unit tests which can be run with make `ut_XXX`. If we can find a way to make this rough approach work, I would want to extend it so that the board targets are treated the same way as the unit test targets.

```
> make help
.
.
.
   [Unit Tests]
     ut_<test>            - Build unit test <test>
                            supported targets are:  utils approximate_math rotations foc
.
.
.
```

make/fw.mk
kubark42: These were inherited from moving the original Makefile to `fw.mk`. I'm happy to get rid of them.

Makefile
kalvdans: It is easier to learn make's default recipe selection than to learn an additional builtin variable. I would be very confused why make wouldn't pick my target as default and it would probably take lot of time until I know I have to grep for this variable name.

vedderb:
Also, so far I have put unit tests here:
https://github.com/vedderb/bldc/tree/dev_fw_5_03/tests

I think I prefer that over having a complex makefile. It also makes sense to keep the makefile similar to how they come with chibios, as that makes it much easier to move to later versions of it when the time comes.

PR #357, opened 2021-10-09
[RFC] Split monolithic terminal module into several files

kubark42:
This is for discussion. I was going through the terminal.c module and realized I was having trouble because of its size. This proposal is simply to split it's functionality across several files. This makes for easier maintenance, as changes to individual terminal commands won't touch the main terminal.c module.

I tried to group the files by theme, I'm not sure how successful I was.

There should be no functionality changes.

### DIFF
--- a/terminal.c
+++ b/terminal.c
@@ -29,10 +29,6 @@
 #include "utils.h"
 #include "timeout.h"
 #include "encoder.h"
-#include "drv8301.h"
-#include "drv8305.h"
-#include "drv8320s.h"
-#include "drv8323s.h"
 #include "app.h"
 #include "comm_usb.h"
 #include "comm_usb_serial.h"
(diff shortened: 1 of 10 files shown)


kubark42:
> the main file doesn't need to include so many headers.

Ahhh, thanks for reminding me. Pruning that long list was something I was putting off till the end and I clearly forgot.

> Please put function definitions in .c files and their declaration in header files that correspond to the .c file name.

I waffled on this one. I originally thought to do it in this style, but it winds up with a lot of header cruft. It's not against ANSI-C guidelines to define the function in the header, so this was my solution. My thinking is that I would use a `.c` file if I had something which was truly modular, and perhaps had several private internal functions. But this isn't a strongly held opinion.

I'm fine to go either way. Lemme know which is preferred!

PR #358, opened 2021-10-13
Create optimized trigonometry for THUMB-2 fixed-point calculation

kubark42:
Uses a 6-th order sine/cosine approximation, and employs fixed-point math to greatly reduce computation time.

Exact numbers are hard to come by, but it seems there's a 3-5x speed up as well as a global 50% improvement in error.

The speed gains are in part due to the [ARM THUMB-2 SMLAWy](https://www.keil.com/support/man/docs/armasm/armasm_dom1361289900041.htm) instruction, which allows a signed multiply and accumulate, combined with a bitshift.

The fixed-point conversion takes advantage of a mapping from 2PI radians to 2[units]. The interested reader should inspect the code comments for a better description.

This was inspired by the work at http://www.coranac.com/2009/07/sines/, but departs heavily from the source material.

Below is the API comment, in hopes it will allow the reader an easier time understanding the fixed-point and assembly-foo.

```
/**
 * @brief _SinCos66_FP This function uses a 6th order taylor expansion for atomically calculating
 *                     sine(x) and cos(x). It uses fixed-point calculations, and thus is sensitive
 *                     to the scale of the arguments. (Please see the below notes.) It uses an
 *                     ARM THUMB-2 assembly instruction, and so is not portable to platforms which
 *                     do not
 *
 *                     Coefficients were chosen by setting:
 *                       *) cos(0) = 1
 *                       *) cos(pi/2) = 0
 *                       *) d/dx[cos(x=pi/2)] = -1
 *                       *) Minimizing the root-mean square error
 *
 *                     For the system
 *
 *                        cos(x) ~= a - x^2*((b - x^2*(c - d*x^2))
 *
 *                     the following constants are found through closed-form analysis
 *
 *                        G=pi/2
 *                        a=1
 *                        b = (-45045/G^4*(G^6+2*G^4-156*G^2+360)/(G^3) - (-655*G-4504))/(952*G^2);  // 0.499833652837794
 *                        c = 1/(2*G^4)*(G-6 + 4*b*G^2);  // 0.041388888146338884
 *                        d = 1/(G^6)*(G/2 - 2 + b*G^2);  // 0.0012440842722145661
 *
 *                     Sine is calculated with the same equaiton by recognizing that sin(x) = cos(x+pi/2)
 *
 *                     A transformation of z = x/pi is used to map the circle into the field
 *                     of [-1, 1), which leads itself to easy fixed-point processing.
 *
 *
 *                     NOTE: Only arguments between (-2*pi and +2pi) are valid. Values outside
 *                           that range will cause integer overflow and the algorithm will silently
 *                           fail.
 *                     NOTENOTE: Please read the above note on the valid range.
 *                     NOTENOTENOTE: Srsly, only values in [pi, -pi) should be used, and never, ever
 *                                   exceed (-2*pi+1) and (2*pi-1).
 * @param angle_rad[float] The angle in [radians], in the range of [-pi,pi)
 * @param sin_out[float] The 6th order approximation of sin(x)
 * @param cos_out[float] The 6th order approximation of cos(x)
 */
```

This has been tested on an NRF52840 platform, and shows a speed increase of 3-10x, depending. The accuracy is also improved, frequently by ~10x, although in some places it's as low as only 2x.

Hand-counting the assembly instructions, it looks like the new form has about 44, of which only 36 will be used in a worst-case scenario. Only 11 are floating point operations. Contrast this with the original, which has 62 instructions of which 27 are FPU operations.

This has been unit tested somewhat, I am in the middle of elaborating a full test case.

NOTE: THIS HAS NOT BEEN TESTED ON AN STM32.

### DIFF
--- a/math/optimized_trig.h
+++ b/math/optimized_trig.h
@@ -1,0 +1,266 @@
+/*
+	Copyright 2021 - Kenn Sebesta
+
+	This file is part of the VESC firmware.
+
+	The VESC firmware is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    The VESC firmware is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+    */
+
+#ifndef OPTIMIZED_TRIG_H_
+#define OPTIMIZED_TRIG_H_
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <math.h>
+#include <stdlib.h>
+
+/**
+ * @brief cosine6TaylorApproximation Fixed-point 6th-order taylor approximation which returns cosine(x)
+ * @param AQ     aq * 2^qN
+ * @param BQm   -bq * 2^qN
+ * @param CQ     cq * 2^qN
+ * @param DQm   -dq * 2^(16 + qN - 2*qX)
+ * @param X2    x^2 * 2^(2qX)
+ * @param X2_16 x^2 * 2^16
+ * @return
+ */
+__attribute__((always_inline))
+inline static int32_t cosine6TaylorApproximation(const int32_t AQ, const int32_t BQm, const int32_t CQ, const int32_t DQm, const int32_t X2, const int32_t X2_16) {
+   /*
+    * out = a - bx^2 + cx^4 - dx^6
+    *     = a - x^2*((b - x^2*(c - d*x^2))
+    *     = a - x^2*(((-bm) - x^2*(c - (-dm)*x^2))
+    *     = a + x^2*(bm + x^2*(c + dm*x^2))
+    */
+
+   int32_t output;
+#if __thumb2__  // Only compile this branch when the thumb instructions are present
+   // ARM THUMB-2 assembly blob which takes advantage of the SMLAWb operator, which performs a multily and accumulate
+   asm ("SMLAWb %[Y3], %[X2], %[D], %[C]\n"       // (y1 = cq  + dqm*(x)^2)
+        "SMLAWb %[Y3], %[Y3], %[X2_16], %[B]\n"   // (y2 = bqm +  y1*(x16)^2)
+        "SMLAWb %[Y3], %[Y3], %[X2_16], %[A]\n"   // (y3 = aq  +  y2*(x16)^2)
+        : [Y3] "=&r" (output)
+        : [A] "r" (AQ), [B] "r" (BQm), [C] "r" (CQ), [D] "r" (DQm), [X2] "r" (X2), [X2_16] "r" (X2_16)
+        :);
+
+#else
+   int Y3=(((int64_t)X2*DQm)>>16)+CQ; // (y1=c-d*x^2)
+   Y3=(((int64_t)Y3*X2_16)>>16)+BQm; // (y2=-(b-y1*x16^2))
+   Y3=(((int64_t)Y3*X2_16)>>16)+AQ; // (y3=y2*x16^2+a)
+   output = Y3;
+
+#endif
+
+   return output;
+}
+
+
+/**
+ * @brief _SinCos66_FP This function uses a 6th order taylor expansion for atomically calculating
+ *                     sine(x) and cos(x). It uses fixed-point calculations, and thus is sensitive
+ *                     to the scale of the arguments. Please see the below notes.
+ *
+ *                     Coefficients were chosen by setting:
+ *                       *) cos(0) = 1
+ *                       *) cos(pi/2) = 0
+ *                       *) d/dx[cos(x=pi/2)] = -1
+ *                       *) Minimizing the root-mean square error
+ *
+ *                     For the system
+ *
+ *                        cos(x) ~= a - x^2*((b - x^2*(c - d*x^2))
+ *
+ *                     the following constants are found through closed-form analysis
+ *
+ *                        G=pi/2
+ *                        a=1
+ *                        b = (-45045/G^4*(G^6+2*G^4-156*G^2+360)/(G^3) - (-655*G-4504))/(952*G^2);  // 0.499833652837794
+ *                        c = 1/(2*G^4)*(G-6 + 4*b*G^2);  // 0.041388888146338884
+ *                        d = 1/(G^6)*(G/2 - 2 + b*G^2);  // 0.0012440842722145661
+ *
+ *                     Sine is calculated with the same equaiton by recognizing that sin(x) = cos(x+pi/2)
+ *
+ *                     A transformation of z = x/pi is used to map the circle into the field
+ *                     of [-1, 1), which leads itself to easy fixed-point processing.
+ *
+ *
+ *                     NOTE: Only arguments between (-2*pi and +2pi) are valid. Values outside
+ *                           that range will cause integer overflow and the algorithm will silently
+ *                           fail.
+ *                     NOTENOTE: Please read the above note on the valid range.
+ *                     NOTENOTENOTE: Srsly, only values in [pi, -pi) should be used, and never, ever
+ *                                   exceed (-2*pi+1) and (2*pi-1).
+ * @param angle_rad[float] The angle in [radians], in the range of [-pi,pi)
+ * @param sin_out[float] The 6th order approximation of sin(x)
+ * @param cos_out[float] The 6th order approximation of cos(x)
+ */
+__attribute__((always_inline))
+inline static void _SinCos66_FP(const float angle_rad, float *sin_out, float *cos_out) {
+   const uint8_t qN = 28;
+   const uint8_t qX = 16;
+   const uint8_t qQ = 30;
+
+   const float PI_INVERSE = 1.0f/M_PI;  // Forces the compiler to precompute the inverse of PI
+   const int32_t FP_PI_Q = (1<<qQ);  // qQ
+   const int32_t FP_PI_HALVES_Q = FP_PI_Q>>1; // qQ
+   const int32_t FP_PI = (1<<qX);  // qX
+   const int32_t FP_PI_HALVES = FP_PI>>1;  // qX
+
+   // This is the point, in [%/100], where the 1st order approximation is more
+   // accurate than the 6th order fixed-point approximation.
+   const float approximationCrossoverPoint = 0.9658377177;
+
+   const double G=M_PI/2;
+   const double G2=G*G;
+   const double G3=G2*G;
+   const double G4=G2*G2;
+   const double G6=G4*G2;
+
+   // a - x^2*((b - x^2*(c - d*x^2))
+   // Coefficients found by minimizing the RMS error, and ensuring that the sine wave passes perfectly by 0 and 1.
+   const double a=1;
+   const double b = (-45045/G4*(G6+2*G4-156*G2+360)/(G3) - (-655*G-4504))/(952*G2); //# 4.933160419860462,;  // 0.499833652837794
+   const double c = 1/(2*G4)*(G - 6 + 4*b*G2);  // 0.041388888146338884
+   const double d = 1/(G6)*(G/2 - 2 + b*G2);  // 0.0012440842722145661
+
+   // Transform x to q, where q = x/pi
+   // aq - q^2*((bq - q^2*(cq - dq*q^2))
+   const double aq = a;
+   const double bq = b * G2*4;
+   const double cq = c * G4*16;
+   const double dq = d * G6*64;
+
+   // Convert the coefficients to fixed point, and inject a negative sign for certain ones. This negative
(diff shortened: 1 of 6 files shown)


math/optimized_trig.h
```diff
@@ -0,0 +1,222 @@
+/*
+	Copyright 2016 - 2019 Benjamin Vedder	benjamin@vedder.se
```
kalvdans, line 2: Update copyright

math/optimized_trig.h
```diff
@@ -0,0 +1,230 @@
...
+   // out = a - bx^2 + cx^4 - dx^6
+   //     = a - x^2*((b - x^2*(c - d*x^2))
+   //     = a - x^2*(((-bm) - x^2*(c - (-dm)*x^2))
+   //     = a + x^2*(bm + x^2*(c + dm*x^2))
+
+   int32_t output;
+#if 1  // TODO: This needs to be #ifdef'ed for platforms which do not have the appropriate THUMB-2 instructions
+   // ARM THUMB-2 assembly blob which takes advantage of the SMLAWt operator, which performs a multily and accumulate
+   asm ("SMLAWb %[Y3], %[X2], %[D], %[C]\n"  // (y1=c-d*x^2
+        "SMLAWb %[Y3], %[Y3], %[X2_16], %[B]\n"   // (y2=-(b-y1*x16^2))
```
kalvdans, line 49: Perhaps you mean 2^16 in the comment? Also the same below in the C version.

math/optimized_trig.h
kubark42: Fixed!

math/optimized_trig.h
```diff
@@ -0,0 +1,230 @@
...
+
+         int32_t X2 = X*X;                 // 2*qX
+         int32_t X2_16 = X2 >> (2*qX-16);  // 16
+
+         output = cosine6TaylorApproximation(AQ, BQm, CQ, DQm, X2, X2_16);
+      }
+
+      // Check if the original angle is in the 2nd or 3rd quadrant
+      if ((is2ndOr4thQuadrant == true && !is3rdOr4thQuadrant == true) || // the 2nd quadrant
+          (!is2ndOr4thQuadrant == true && is3rdOr4thQuadrant == true)) { // the 3rd quadrant
```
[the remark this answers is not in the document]
kubark42: `!a == true`, what was I thinking? At the very least `a==false` would be cleaner! I'll do it your suggested way as the cleanest.

kubark42:
With a global inline replacement of all `utils_fast_sincos_better()`

```
 302352	   3444	 144820	 450616	  6e038	build/BLDC_4_ChibiOS.elf
```

With a global `.c`-based replacement of all `utils_fast_sincos_better()`
```
 300880	   3444	 144820	 449144	  6da78	build/BLDC_4_ChibiOS.elf
```

Compared to the original using only `utils_fast_sincos_better()`
```
 300848	   3444	 144820	 449112	  6da58	build/BLDC_4_ChibiOS.elf
```

Looks like when inlined it adds 1.5k.

vedderb:
If it works well I think you can just replace the functions in utils, it is not worth keeping unused code. I also think it is worth adding angle normalization to the beginning to make it less likely to get overflows even if that has some overhead. The old normalization should be fine:

```
while (angle < -M_PI) { angle += 2.0 * M_PI; }
while (angle >  M_PI) { angle -= 2.0 * M_PI; }
```

kubark42:
Those FPU operations would negate much of the optimization and would be redundant with https://github.com/vedderb/bldc/pull/358/files#diff-96067bfcfc31af06a30c0f3a1f1186440eb2a7e680a44cce52c71059d17e4961R155

That above section could be taken out, somewhat mitigating the additional FPU operations.

It would be better, IMHO, to look upstream to ensure that all calls into this function are sanitized. If there are calls which use values which are not contained then the angle can be normalized beforehand. 

It would also be possible to introduce a new function which wraps the call to _SinCos66_FP in the proposed guards. Then the raw form could be used when it's certain that the bounds can never be exceeded.

langwadt:
http://smd.hu/Data/Analog/DSP/21xx/DSP%20Applications%20I/CHAP_4.PDF section 4.2 might have a few ideas

math/optimized_trig.h
kalvdans: Oh I see, it is fine the way it stands. I wasn't reading the whole documentation. Perhaps some spaces around binary operators will help readability. We can close this thread!

kubark42: Maybe `(y2=-(b-y1*(x_16)^2))`?

kalvdans:
> that would be the focus of another PR,

I completely agree.

langwadt:
@kubark42 I guess I should have said "**negate** a twos-complement number and clear the sign bit". It basically flips 0x40000000-0x0 into 0x0-0x40000000  (similar to your line 173)

I'm not sure it is better, I haven't seen what the compiler output is. It just seemed like a lot of code and many ifs for something I remembered as a few negates and bitands

math/optimized_trig.h
```diff
@@ -0,0 +1,233 @@
...
+/**
+ * @brief _SinCos66_FP This function uses a 6th order taylor expansion for atomically calculating
+ *                     sine(x) and cos(x). It uses fixed-point calculations, and thus is sensitive
+ *                     to the scale of the arguments. Please see the below notes.
+ *
+ *                     Coefficients were chosen by setting:
+ *                       *) cos(0) = 1
+ *                       *) cos(pi/2) = 0
+ *                       *) d/dx[cos(x=pi/2)] = -1
+ *                       *) Minimizing the root-mean square error
```
ledvinap, line 78: It may be better to minimize maximum absolute/relative error instead of RMS (infinite-norm). Mean error is a problematic is there is small part of argument space where error is large. Also both relative and absolute error is important .. 

Look at https://github.com/samhocevar/lolremez, IICR it was quite easy to optimize polynomial coefficients for (floating point) sin/cos approximation

math/optimized_trig.h
```diff
@@ -0,0 +1,233 @@
...
+
+   // Convert the coefficients to fixed point, and inject a negative sign for certain ones. This negative
+   // sign will be useful when using the multiply & accumulate assembly instruction.
+   const int32_t AQ = aq * (1UL<<qN);  // qN
+   const int32_t BQm = -(bq * (1UL<<qN) + 0.5L);  // qN
+   const int32_t CQ = cq * (1UL<<qN) + 0.5L;  // qN
+   const int32_t DQm = -(dq * (1UL<<((48-(32-qN)-2*qX))) + 0.5L);  // ((16 + qN - 2*qX)))
+
+   // Convert the angle to fixed-point.
+   int32_t angle_FP = (angle_rad*PI_INVERSE)  * (1<<qQ); //  Do not round, as it is important not to reach +2^15
```
ledvinap, line 148: Low order bits are discarded later. It shall be possible to allocate some bits for range outsize -2pi..2pi and normalize argument using bitmasking/simple integer math. 
It will also allow rounding here, increasing accuracy a bit (but later shift must be considered - rounding before 1<<16 division has almost no effect)

math/optimized_trig.h
```diff
@@ -0,0 +1,233 @@
...
+
+   // Shift X for proper scaling for use in the fixed-point math
+   int32_t X = Q >> (qQ-qX);  //qX
+
+   // Cosine(x). This MUST be done first
+   {
+      int32_t output;  // qN
+
+      // If the angle is close to pi/2, then use the simple approximation cos(pi/2-eps) ~= pi/2-eps
+      if (Q > (int32_t)(FP_PI_HALVES_Q*.95)) {
```
ledvinap, line 183: It will be great to use cutoff where accuracy os simple and 6th order approximation meet. Also take care to avoid discontinuity at this point.
Also different running time may theoretically cause nasty bugs if calling code timing is important.

ledvinap (commented):
Just some thoughts

math/optimized_trig.h
kubark42: I had never seen LolRemez, I love it.

Having looked at the graphs, relative error is acceptably small, like <1 e-4 across much fo the range. The biggest driver of error is the limits of integer precision and dynamic range. When the value of `cos(x)` starts getting very close to 0 the relative error gets really bad because there's only so much you can express with an integer value which itself is close to 0. 

This is why I switch to the approx `cos(x) ~= pi/2-x` for values of `x` close to pi/2. This in theory adds a tiny jump in the graphed curve, however in reality this jump is lost in the noise of cos(x) values which are less than 1e-3.

math/optimized_trig.h
kubark42: > It will be great to use cutoff where accuracy os simple and 6th order approximation meet. Also take care to avoid discontinuity at this point.

The cutoff is around there. I didn't explore the space carefully, but it's around the 95% mark.

I had thought about redoing my coefficient calculation in order to avoid potential discontinuity, but ultimately decided it wasn't important because the discontinuity in C(0) and C(1) is small compared to the quantization errors on that scale.

> Also different running time may theoretically cause nasty bugs if calling code timing is important.

I am somewhat wary about that as well. It's easy to see how optimization could lead developers to believe there is more free CPU than is safe to use, ultimately resulting in a 1202 problem. But it feels unlikely in this case, since the optimization will only affect, at most, one of the two sine(x)/cos(x) operations and so the bulk of the function still runs no matter what.

math/optimized_trig.h
kubark42: @ledvinap, that's an excellent observation. It can save some additional cycles because there's no floating-point division to scale back to radians. Plus, the quantization error as cos(x)--> 0 completely disappears.

math/optimized_trig.h
```diff
@@ -0,0 +1,233 @@
...
+         output = cosine6TaylorApproximation(AQ, BQm, CQ, DQm, X2, X2_16);
+      }
+
+      // Check if the original angle is in the 3rd or 4th quadrant
+      if (is3rdOr4thQuadrant == true) {
+         output = -output;
+      }
+
+      // Scale the output from qN to a float
+      *sin_out = (float)output / (1<<qN);
```
ledvinap, line 229: (you can use multiplication with inverse here. gcc has to use division unless relaxed math is enables)

kubark42: I'll do that. The current output with -O2 is smart enough to recognize that this is a conversion to floating-point with a shift, but the multiplication solves any reliance on the compiler to do the right thing.

```
      *sin_out = (float)output / (1<<qN);
 8028ac2:	ee07 3a90 	vmov	s15, r3
 8028ac6:	eefa 7ac2 	vcvt.f32.s32	s15, s15, #28
 8028aca:	edc4 7a0e 	vstr	s15, [r4, #56]	; 0x38
```

kubark42: This looks reasonable as a solution:

```
      // Scale the output from qN to a float
#if __thumb2__  // Only compile this branch when the thumb instructions are present
      *sin_out = (float)output / (1<<qN);
#else
      const float den = 1/(1<<qN);
      *sin_out = output * den;
#endif
```

math/optimized_trig.h
kubark42: The alternative is to just do
```
         // Ensure angle is in [-pi,pi)
         while (angle_rad < -M_PI) { angle_rad += 2.0 * M_PI; }
         while (angle_rad >  M_PI) { angle_rad -= 2.0 * M_PI; }

         if ((is2ndOr4thQuadrant && !is3rdOr4thQuadrant) || // the 2nd quadrant
             (!is2ndOr4thQuadrant && is3rdOr4thQuadrant)) { // the 3rd quadrant
            output = -M_PI + angle_rad;
         } else {
            output = angle_rad;
         }
```

While that has those while loops, they are (1) unlikely to be looped more than once because the function shouldn't be -passed values in the wrong range-- and if it is then the fixed-point math generally breaks so we no longer care about efficiency-- and (2) it helps even out the processor requirements for both estimation branches (esp. since this branch no longer has the `*1/1<<qN` factor.

I'm inclined to go with this option, as it is much more readable. The downsides only occur when we have extra CPU budget.

vedderb:
I just want to point something out before you dive too deep into optimizations: The obvious way to make the code run faster and have the best accuracy is to represent the angle as sine and cosine directly in mcpwm_foc. Almost every place, except the position controller, start by using the sincos function, so you might as well store it as sine and cosine.

It would be nice to have a faster and more accurate sincos function, but it is not strictly necessary. It is not the bottleneck in speed, we are not running out of CPU cycles yet, and more importantly in my case: my time is very limited and I try to focus on things that are the most important and add the most value first.

At the moment I want to get a new stable and tested release out as there are so many significant improvements in dev_fw_5_03 over master. Major changes, like how the angle is represented, take a lot of testing and delay the release, so before we go there I will focus on the release. We can still change branch on the pull requests then, I'm just pointing it out as I won't give this the highest priority now.

vedderb:
That sounds good, I'm happy to hear that. I also enjoy learning and experimenting with these things. I just wanted to point this out as I like to encourage people who try to dive into the code and improve it, and I don't want to disappoint you if it takes some time to get it merged.

kubark42:
> don't want to disappoint you if it takes some time to get it merged.

The value of a contribution is not measured by the speed it is contributed. 😃 

This is a conversation for a separate issue, but have you considered adding other project collaborators? I can see this codebase has attracted some top-notch contributors, if one or several could be trusted with PR testing and review, it would free you to focus on core tasks.

PR #370, opened 2021-10-27
[Math][Optimization] Provide a CPU-efficient boxcar average

kubark42:
This is the classic CPU-efficient implementation of a moving average filter. Rather than recompute the entire sum for `i=k..k+N`, it exploites the fact that the sum from `i=k-1..k+N-1` is known from the previous iteration. Thus, all that remains is to do is remove `val[k-1]` and add `val[k+N]`. So an arbitrarily large O(n) filter becomes size invariant with only 2 addition operations.

I have not looked to see if there were other opportunities for using this CPU-optimized moving window filter. It's not clear to me how often this ADC thread runs, but I think it might be 500Hz. So this saves 6-12 floating point additions per loop, depending on if the ADC filter is turned on. If I'm right that it's a 500Hz loop, that's 3000-6000FPU operations/sec. That's not terrible!

It's also important to point out that previously, when the filter was first turned on the initial results were compromised because the filter was dividing by the *nominal* number of samples, but until the filter is completely filled it should be divided by the actual number of samples.

### Before
```
 300976	   3444	 144884	 449304	  6db18	build/BLDC_4_ChibiOS.elf
```
### After
```
 301152	   3444	 144892	 449488	  6dbd0	build/BLDC_4_ChibiOS.elf
```
NOTE: Not tested, not even unit tested.

### DIFF
--- a/applications/app_adc.c
+++ b/applications/app_adc.c
@@ -128,21 +128,14 @@ static THD_FUNCTION(adc_thread, arg) {
 
 		read_voltage = pwr;
 
-		// Optionally apply a mean value filter
+		// Optionally apply a moving mean value filter
 		if (config.use_filter) {
-			static float filter_buffer[FILTER_SAMPLES];
-			static int filter_ptr = 0;
+			static float filter_buffer[FILTER_SAMPLES];  // ANSI-C guarantees this will initialize to all 0s. This is important insider the filter.
+			static uint32_t filter_current_idx = 0;
+			static float filtered_pwr_accum = 0;
+			static bool isBufferFull = true;
 
-			filter_buffer[filter_ptr++] = pwr;
-			if (filter_ptr >= FILTER_SAMPLES) {
-				filter_ptr = 0;
-			}
-
-			pwr = 0.0;
-			for (int i = 0;i < FILTER_SAMPLES;i++) {
-				pwr += filter_buffer[i];
-			}
-			pwr /= FILTER_SAMPLES;
+			pwr = utils_moving_average_filter_f(pwr, &filtered_pwr_accum, filter_buffer, &filter_current_idx, &isBufferFull, FILTER_SAMPLES);
 		}
 
 		// Map the read voltage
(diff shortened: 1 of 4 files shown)


vedderb:
I like the trick with the accumulator, but there are a few things I don't like with this PR.

* It is untested. Before I can accept it I have to get a setup with an analog throttle and see if it works. It would be really helpful if that was already done and confirmed working.
* If there is a choice, readability and easy of use always comes before performance. Storing the inverse of the same value that has to be initialized manually for avoiding one division is not worth it in most cases. When it is, it would be better to have an init function that takes care of that.
* If this filter would be used in a critical section the worst case execution time is more important than the average time. If the ADC DMA interrupt in mcpwm_foc takes too long that update will be too late and the motor will glitch. The extra loop for initializing the accumulator can make this function run much slower if it has a lot of samples.

The following approach would be nice:
* Remove the buffer and just use an accumulator.
* Use an init function that accept an initial value and number of samples, then sets up the accumulator and pre-divides the number of samples.
* When the filter is off keep updating the accumulator and return the input value. Then it is ready to go the next time it is switched on.
* If possible, a unit test with a makefile in the tests directory would be nice.
* Put the filter in digital_filter instead of utils.

vedderb:
Just realized the buffer is needed, sorry about the confusion.

vedderb:
The problem with the numerical instability also worries me a bit, as any error will just be left in the accumulator. If the accumulator has to be recalculated from the buffer on a regular basis it makes this filter quite useless as the worst case execution time is what matters the most.

vedderb:
It would be nice to approximate a moving average with an exponential moving average, as that can be done with a first order IIR filter similar to the UTILS_LP_FAST macro.

According to
https://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average
the approximation can be done by choosing filter_constant = 2 / (N + 1)

A new macro can be introduced for that:
```
#define UTILS_MOVING_AVG_APPROX(value, sample, N)	(value -= (2.0 / (N - 1)) * ((value) - (sample)))
```

That is a convenient way to replace moving average filters with something faster that has a similar behavior.

vedderb:
Here is an update:
https://github.com/vedderb/bldc/commit/e9cf6eb4e8ed86b6dacba216604b2c205af177a0

It now uses the approximation instead. It also keeps the filter running when it isn't being used, so that it is ready when it is enabled. That should not matter so much as the motor is switched off when writing the config anyway and the filter will have time to converge, but it makes the ADC app cleaner and easier to follow.

kubark42:
> Here is an update...

I'm sold. 48LoC to 25LoC, and it solves all the toggling problems we've outlined in this thread. Ship it!

In fact, I like this approach more as it is more inline with my way of seeing the world (Kalman Filters). Moving average filters always bother me as they inevitably delay the signal, unlike the class of Luenberger observers.

Regarding testing, I understand and agree with where you're coming from. I sensed this PR would be fertile ground for innovation and so I put-off testing which I felt would be premature. As it turns out, testing this PR's original code would not have been valuable since ultimately the solution looks very different. 

Apologies if this wasn't clear from the beginning. I have opened https://github.com/vedderb/bldc/issues/371 as a suggestion for how contributors might be able to label PRs so that it's clear they are in an early stage.

Closing this PR as the exponential moving-average filter has been merged into the mainline.

langwadt:
> > Here is an update...
> 
> I'm sold. 48LoC to 25LoC, and it solves all the toggling problems we've outlined in this thread. Ship it!
> 
> In fact, I like this approach more as it is more inline with my way of seeing the world (Kalman Filters). Moving average filters always bother me as they inevitably delay the signal, unlike the class of Luenberger observers.

A moving average and exponential filter is going to have similar delay for similar "N", just look at the step response

kubark42:
> A moving average and exponential filter is going to have similar delay for similar "N", just look at the step response

I agree. 

My comment was related to how the resulting code looks. A KF (which is a form of Luenberger Observer), takes the discrete-time form `x_k+1 = x_k + K*y`, where `y` is the error between measurement and state prediction. This is closer in code form to `x_k+1 = alpha * x_k + (1-alpha)*y`, where `y` is the measurement, than it is to `x_k+1 = sum(y_k-n... y_k)`.

ElwinBoots:
I agree that switching the boxcar average to a simple low-pass filter is a good choice. What I don't really understand is why you stick to naming it a moving average approximation. Wouldn't it be better to just call it low pass filter and use the function you already have for this: UTILS_LP_FAST?

The moving average filter has one distinct feature that the lowpass doesn't have: it will fully filter out certain frequencies from your input signal. By naming it moving average one might think this filter has this attribute, but it won't. I don't know whether it is important for this use case, but I have used this property of the moving average to prevent excitation of the eigenmodes of a system in the past.

vedderb:
The reason for the naming is that the one property a moving average filter has is the number of samples N, and this macro makes something that behaves similar in that regard. That makes it easier to replace or re-implement moving average filters without thinking about how to choose an appropriate filter constant if that choice already was done before in the form of N.

PR #374, opened 2021-11-07
Turn FPU math into truth table

kubark42:
I noticed that the below code was evaluated completely from the signs of `float` arguments:

```
mod_alpha_filter_sgn = (1.0 / 3.0) * (2.0 * SIGN(ia_filter) - SIGN(ib_filter) - SIGN(ic_filter));
mod_beta_filter_sgn = ONE_BY_SQRT3 * (SIGN(ib_filter) - SIGN(ic_filter));
```

Thus this is really just a truth table. This PR eliminates all the FPU math involved, resulting in an extremely streamlined calculation.

This is approximately 9x more performant in testing. On the testbed (nrf52840) it trims ~380 cycles per iteration down to ~47 cycles. It also compiles 96 bytes smaller, to boot.

It is not immediately easy to read, I have done my best to comment the crap out of this code so that it's obvious what's going on.

If anyone can fill in the reason for calculating this quantity, I would like to drop a comment in to explain how the outputs of this function are used.

Note that this PR makes bitwise use of the IEEE 754 spec by exploiting the fact that the last bit is reserved for the sign. This optimization is portable: it will work on big- and little-endian processors, and to the best of my knowledge there are no readily available processors which do not adhere to IEEE 754.

### DIFF
--- a/math/optimized_helper_functions.h
+++ b/math/optimized_helper_functions.h
@@ -1,0 +1,95 @@
+/*
+	Copyright 2021 Kenn Sebesta
+
+	This file is part of the VESC firmware.
+
+	The VESC firmware is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    The VESC firmware is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+    */
+
+#ifndef OPTIMIZED_HELPER_FUNCTIONS_H_
+#define OPTIMIZED_HELPER_FUNCTIONS_H_
+
+#include <stdbool.h>
+#include <stdint.h>
+#include "datatypes.h"
+#include <math.h>
+#include <stdlib.h>
+
+// This takes advantage of the IEEE 754 standard, which requires that the sign
+// bit of single-precision floats is the last bit.
+// It is faster than `signbit()` because it does not have a bit shift and does no type checking.
+// On the flip side, it is dangerous because it does no checking of any kind. It will
+// not care if the input is not a `float`, if it is `NaN`, etc...
+// It is guaranteed to work on all systems which adhere to IEEE 754.
+// This works on both big- and little-endian systems.
+// The macro returns 1 if a `float` is negative, 0 if otherwise
+#define IS_FLOAT_IEEE754_NEGATIVE(x) (*(uint32_t *)(&(x)) & (1U<<31))
+
+
+/**
+ * @brief calculate_mod_alpha_beta_filter_sign Exploits the fact that two of the three
+ * currents will always have the same sign, and the third will always have the opposite sign.
+ * @param mod_alpha_filter_sgn Returns a value on the set of [-4/3, -2/3, 2/3, 4/3]
+ * @param mod_beta_filter_sgn Returns a value on the set of [-2/sqrt(3), 0, 2/sqrt(3)]
+ * @param ia_filter The filtered i_a
+ * @param ib_filter The filtered i_b
+ * @param ic_filter The filtered i_c
+ */
+inline __attribute__((always_inline)) void calculate_mod_alpha_beta_filter_sign(float *mod_alpha_filter_sgn, float *mod_beta_filter_sgn, float ia_filter, float ib_filter, float ic_filter) {
+	// This calculates the below code, which is effectively a truth table with floats as outputs:
+	//    mod_alpha_filter_sgn = (1.0 / 3.0) * (2 * SIGN(ia_filter) - SIGN(ib_filter) - SIGN(ic_filter));
+	//    mod_beta_filter_sgn = ONE_BY_SQRT3 * (SIGN(ib_filter) - SIGN(ic_filter));
+
+	/* Perform a compile-time assert, checking that a float and a uint32_t are the same size on this machine. */
+	_Static_assert(sizeof(float) == sizeof(uint32_t), "This compilation is not compatible. Please fix the IEEE754 relationship between floats and uints");
+
+#pragma GCC diagnostic push  // Stores any GCC directives
+#pragma GCC diagnostic ignored "-Wstrict-aliasing"  // Disables the strict-aliasing directive. This is necessary for the IS_FLOAT_IEEE754_NEGATIVE() macro.
+
+	// Check if both `ib_filter` and `ic_filter` have the same sign.
+	if (IS_FLOAT_IEEE754_NEGATIVE(ib_filter) == IS_FLOAT_IEEE754_NEGATIVE(ic_filter)) {
+		// Both `ib_filter` and `ic_filter` have the same sign, so they cancel out
+		*mod_beta_filter_sgn = 0;
+
+		// Check if `ia_filter` is negative
+		if (IS_FLOAT_IEEE754_NEGATIVE(ia_filter)) {
+			// ia_filter is negative, which means both `ib_filter` and `ic_filter` are positive.
+			*mod_alpha_filter_sgn = -4.0 / 3.0;
+		} else {
+			// ia_filter is positive, which means both `ib_filter` and `ic_filter` are negative.
+			*mod_alpha_filter_sgn =  4.0 / 3.0;
+		}
+	} else  {
+		// `ib_filter` and `ic_filter` have opposite signs
+
+		if (IS_FLOAT_IEEE754_NEGATIVE(ia_filter)) {
+			// ia_filter is negative. `ib_filter` and `ic_filter` have opposite signs so they cancel out.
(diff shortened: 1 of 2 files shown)


kubark42:
This PR trades readability for speed. My hunch is that it's a bridge too far for this project at this point in time. I hope it is a useful discussion point, though.

mcpwm_foc.c
```diff
@@ -3557,6 +3558,77 @@ static THD_FUNCTION(pid_thread, arg) {
 	}
 }
 
+// This takes advantage of the IEEE 754 standard, which requires that the sign
```
kubark42, line 3561: This function declaration is not in the right place in the file. Should it go higher up or lower down, though?

vedderb:
That is an interesting solution. I think it would be nice to find a style to include optimizations like this. It would also be good to have a way to quantify how much difference they make. For doing that it is important to use the same compilation flags, namely -mfloat-abi=hard -mfpu=fpv4-sp-d16 -fsingle-precision-constant

I guess the function has to be in the same file for the inline to work. Maybe we can include a C (or H) file in the mcpwm_foc.c file with optimizations like this so that they don't take too much space in the file itself (as the file is long already and cryptic optimizations don't make it easier to read). Then we can have an ifdef that enables or disables them. Disabling them would use the old solution, which also serves as documentation about what is being done there.

For proper regression testing I think a hardware testbench will be needed in the end that runs before and after and compares the performance. That is something I want to build soon, but that is a topic for another day.

vedderb:
A short description on why it looks the way it does:

The code is for dead-time compensation. When the half-bridges switch between high and low there is a short pause where both switches are off. This pause is needed to prevent shorting out the power rail and is called the dead-time. The dead-time will cause the modulation to be slightly different from what we set, which causes some distortion at low modulated voltages. This makes it harder to track the motor at low speed.

The compensation relies on the fact that when both switches are off, the voltage in that half-bridge can be determined directly from the current on that phase. If the current is flowing towards that half-bridge, it will raise the voltage to the bus voltage as the body diode in the top MOSFET starts to conduct then, which makes it look like that half-bridge is on during the dead-time and thus it is added to the modulation. Likewise, if the current is flowing away from that half-bridge the opposite happens. That is why we can use a Clarke-transformed truth table based on the current signs to compensate the modulation for the effects of dead-time.

kubark42:
> Maybe we can include a C (or H) file in the mcpwm_foc.c file with optimizations like this so that they don't take too much space in the file itself (as the file is long already and cryptic optimizations don't make it easier to read). Then we can have an ifdef that enables or disables them.

I like this. 1) It's useful as part of a unit test, to ensure identical results to the original math and 2) it makes it easy to keep the code portable, since in general I would not blindly trust bitwise optimizations on new platforms.

kubark42:
I moved the optimizations into a separate header file, and added a `#ifdef FLOAT_BITWISE_OPTIMIZATIONS_TURNED_OFF` to turn them off if so desired.

langwadt:
> I noticed that the below code was evaluated completely from the signs of `float` arguments:
> 
> ```
> mod_alpha_filter_sgn = (1.0 / 3.0) * (2.0 * SIGN(ia_filter) - SIGN(ib_filter) - SIGN(ic_filter));
> mod_beta_filter_sgn = ONE_BY_SQRT3 * (SIGN(ib_filter) - SIGN(ic_filter));
> ```
> 
> Thus this is really just a truth table. This PR eliminates all the FPU math involved, resulting in an extremely streamlined calculation.
> 
> This is approximately 9x more performant in testing. On the testbed (nrf52840) it trims ~380 cycles per iteration down to ~47 cycles. It also compiles 96 bytes smaller, to boot.
> 

must depend on the -O setting

https://godbolt.org/z/bxovW4c9r

kubark42:
>must depend on the -O setting

It could also depend on the platform. I only tested on an nRF, so while I know that it was compiled to use the FPU, I don't know how it performs compared to the STM32. It could be that Nordic's silicon is nowhere near as advanced.

langwadt:
> > must depend on the -O setting
> 
> It could also depend on the platform. I only tested on an nRF, so while I know that it was compiled to use the FPU, I don't know how it performs compared to the STM32. It could be that Nordic's silicon is nowhere near as advanced.

there was nothing STM32 specific in the options. The nRF use the same Cortex-M4F core so the only difference is possibly a slightly different timing on fetching instructions

kubark42:
> The nRF use the same Cortex-M4F core so the only difference is possibly a slightly different timing on fetching instructions

Perhaps. If the instruction is executed from flash, it's an order of magnitude slower. STM has its fancy ART accelerator to try to load instructions pre-emptively, I don't know if nRF has anything similar.

Once I get a chance, I'll try to do the same test where I used the STM32's built-in hardware CPU Cycle Counter. It might prove enlightening.

langwadt:
> > The nRF use the same Cortex-M4F core so the only difference is possibly a slightly different timing on fetching instructions
> 
> Perhaps. If the instruction is executed from flash, it's an order of magnitude slower. STM has its fancy ART accelerator to try to load instructions pre-emptively, I don't know if nRF has anything similar.
> 
> Once I get a chance on a VESC, I'll try to do the same test as where I used the nRF's built-in hardware CPU Cycle Counter. It might prove enlightening.

afaict the nRF has 2k instruction cache

math/optimized_helper_functions.h
```diff
@@ -0,0 +1,100 @@
...
+#include <stdlib.h>
+
+// This takes advantage of the IEEE 754 standard, which requires that the sign
+// bit of single-precision floats is the last bit.
+// It is faster than `signbit()` because it does not have a bit shift and does no type checking.
+// On the flip side, it is dangerous because it does no checking of any kind. It will
+// not care if the input is not a `float`, if it is `NaN`, etc...
+// It is guaranteed to work on all systems which adhere to IEEE 754.
+// This works on both big- and little-endian systems.
+// The macro returns 1 if a `float` is negative, 0 if otherwise
```
ledvinap, line 36: Actually it returns 1<<31 or 0

math/optimized_helper_functions.h
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * @brief calculate_mod_alpha_beta_filter_sign Exploits the fact that two of the three
+ * currents will always have the same sign, and the third will always have the opposite sign.
+ * @param mod_alpha_filter_sgn Returns a value on the set of [-4/3, -2/3, 2/3, 4/3]
+ * @param mod_beta_filter_sgn Returns a value on the set of [-2/sqrt(3), 0, 2/sqrt(3)]
+ * @param ia_filter The filtered i_a
+ * @param ib_filter The filtered i_b
+ * @param ic_filter The filtered i_c
+ */
+inline __attribute__((always_inline)) void calculate_mod_alpha_beta_filter_sign(float *mod_alpha_filter_sgn, float *mod_beta_filter_sgn, float ia_filter, float ib_filter, float ic_filter) {
```
ledvinap, line 49: Just maybe:
What about collecting sign bits into 3bit integer and use it to index 8 element table with mod_alpha_filter_sgn / mod_beta_filter_sgn values?

math/optimized_helper_functions.h
kubark42: Excellent point, I'll fix that before taking this PR out of draft mode.

math/optimized_helper_functions.h
kubark42: So something like:

```
#define SIGN_BIT_FLOAT_IEEE754(x) (*(uint32_t *)(&(x)) & (1U<<31))

// Possible output values of the truth table
const float A[6] = {0, 1, 3, 4, 5, 6};  // NOTE: Not correctly populated yet
const float B[6] = {0, 1, 3, 4, 5, 6};

void calculate_mod_alpha_beta_filter_sign(float *mod_alpha_filter_sgn, float *mod_beta_filter_sgn, float ia_filter, float ib_filter, float ic_filter) {

    // Create a truth table index in 0b000..0b111.
    int idx = (SIGN_BIT_FLOAT_IEEE754(ia_filter) >> 29) | (SIGN_BIT_FLOAT_IEEE754(ia_filter) >> 30) | (SIGN_BIT_FLOAT_IEEE754(ia_filter) >> 31);

    *mod_alpha_filter_sgn = A[idx];
    *mod_beta_filter_sgn = B[idx];
}
```

That's beautiful in its simplicity, and has the nice effect of being constant in execution time. I'll program it up for real and test it.

math/optimized_helper_functions.h
```diff
@@ -0,0 +1,100 @@
...
+    */
+
+#ifndef OPTIMIZED_HELPER_FUNCTIONS_H_
+#define OPTIMIZED_HELPER_FUNCTIONS_H_
+
+#include <stdbool.h>
+#include <stdint.h>
+#include "datatypes.h"
+#include <math.h>
+#include <stdlib.h>
```
kalvdans, line 27: Include only what you need. I think you need `<stdint.h>` and the macros `ONE_BY_SQRT3` and `SIGN`.

math/optimized_helper_functions.h
ledvinap: Personally, I would use 
```
const struct { float a,b; } tab[8] = {{1,2},...};
```
Also it is good idea to specify all 8 possibilities, ideally with 'safe' value for 'impossible' input combination ... 

As for generated code - ARM is well optimized for bitfield access, so shift operations are nearly free. And avoiding branch saves a cycle.

PR #376, opened 2021-11-08
Frequency response function

kubark42:
Developed in collaboration with @ErwinBoots, this was the outgrowth of https://vesc-project.com/node/3257.

It has been used a fair amount on a VESC6 Mk3, and seems to work well. 

This needs to be merged alongside https://github.com/vedderb/vesc_tool/pull/177.

One limitation which we would like to move past at some point soon is the ability to get multiple samples sequentially. Ideally a user should be able to get several cycles of data and average them together for a cleaner signal.

@vedderb could you let us know how this looks? In particular, notice how we have hijacked the `hfi` configuration to configure this as well. Obviously that's not okay, but I'm not really sure the right way to do this.

### DIFF
--- a/mc_interface.c
+++ b/mc_interface.c
@@ -98,7 +98,7 @@ static volatile motor_if_state_t m_motor_2;
 #endif
 
 // Sampling variables
-#define ADC_SAMPLE_MAX_LEN		2000
+#define ADC_SAMPLE_MAX_LEN		2048
 __attribute__((section(".ram4"))) static volatile int16_t m_curr0_samples[ADC_SAMPLE_MAX_LEN];
 __attribute__((section(".ram4"))) static volatile int16_t m_curr1_samples[ADC_SAMPLE_MAX_LEN];
 __attribute__((section(".ram4"))) static volatile int16_t m_ph1_samples[ADC_SAMPLE_MAX_LEN];
--- a/mc_interface.c
+++ b/mc_interface.c
@@ -2053,7 +2053,13 @@ void mc_interface_mc_timer_isr(bool is_second_motor) {
 			m_vzero_samples[m_sample_now] = zero;
 			m_curr_fir_samples[m_sample_now] = (int16_t)(current * (8.0 / FAC_CURRENT));
 			m_f_sw_samples[m_sample_now] = (int16_t)(f_samp / 10.0);
-			m_frf_samples_mV[m_sample_now] = (int16_t)(mcpwm_foc_get_vd() * 1000 + 0.5);
+
+			if (conf_now->foc_prbs_channel == PRBS_CHANNEL_D) {
+				m_frf_samples_mV[m_sample_now] = (int16_t)(mcpwm_foc_get_vd() * 1000 + 0.5);
+			} else if (conf_now->foc_prbs_channel == PRBS_CHANNEL_Q) {
+				m_frf_samples_mV[m_sample_now] = (int16_t)(mcpwm_foc_get_vq() * 1000 + 0.5);
+			}
+
 			m_status_samples[m_sample_now] = (mcpwm_foc_get_prbs() << 7) | (((mcpwm_read_hall_phase() << 3) | mcpwm_get_comm_step()) & 0x7F);
 
 			m_sample_now++;
--- a/mc_interface.c
+++ b/mc_interface.c
@@ -109,6 +109,7 @@ __attribute__((section(".ram4"))) static volatile uint8_t m_status_samples[ADC_S
 __attribute__((section(".ram4"))) static volatile int16_t m_curr_fir_samples[ADC_SAMPLE_MAX_LEN];
 __attribute__((section(".ram4"))) static volatile int16_t m_f_sw_samples[ADC_SAMPLE_MAX_LEN];
 __attribute__((section(".ram4"))) static volatile int8_t m_phase_samples[ADC_SAMPLE_MAX_LEN];
+__attribute__((section(".ram4"))) static volatile int16_t m_frf_samples_mV[ADC_SAMPLE_MAX_LEN];
 
 static volatile int m_sample_len;
 static volatile int m_sample_int;
@@ -2052,7 +2053,8 @@ void mc_interface_mc_timer_isr(bool is_second_motor) {
 			m_vzero_samples[m_sample_now] = zero;
 			m_curr_fir_samples[m_sample_now] = (int16_t)(current * (8.0 / FAC_CURRENT));
 			m_f_sw_samples[m_sample_now] = (int16_t)(f_samp / 10.0);
-			m_status_samples[m_sample_now] = mcpwm_get_comm_step() | (mcpwm_read_hall_phase() << 3);
+			m_frf_samples_mV[m_sample_now] = (int16_t)(mcpwm_foc_get_vd() * 1000 + 0.5);
+			m_status_samples[m_sample_now] = (mcpwm_foc_get_prbs() << 7) | (((mcpwm_read_hall_phase() << 3) | mcpwm_get_comm_step()) & 0x7F);
 
 			m_sample_now++;
 
@@ -2714,7 +2716,7 @@ static THD_FUNCTION(sample_send_thread, arg) {
 		}
 
 		for (int i = 0;i < len;i++) {
-			uint8_t buffer[40];
+			uint8_t buffer[39];  // 1 + 4 + 4 + 4 + 4 + 4 + 4 + 4 + 4 + 4 + 1 + 1
 			int32_t index = 0;
 			int ind_samp = i + offset;
 
@@ -2747,6 +2749,7 @@ static THD_FUNCTION(sample_send_thread, arg) {
 			}
 
 			buffer_append_float32_auto(buffer, (float)m_f_sw_samples[ind_samp] * 10.0, &index);
+			buffer_append_float32_auto(buffer, (float)m_frf_samples_mV[ind_samp], &index);
 			buffer[index++] = m_status_samples[ind_samp];
 			buffer[index++] = m_phase_samples[ind_samp];
 
--- a/mcpwm_foc.c
+++ b/mcpwm_foc.c
@@ -3930,12 +3930,26 @@ static void control_current(volatile motor_all_state_t *motor, float dt) {
 	utils_truncate_number_abs((float*)&vq_tmp, max_vq);
 	state_m->vq_int += (vq_tmp - vq_presat);
 
-	float noiseScale = conf_now->foc_hfi_voltage_max;
 
 	// Generate PRBS noise
+	float noiseScale = conf_now->foc_prbs_voltage;
 	int noise = prbsGenerator11_increment();
-	state_m->vd = vd_tmp + noise * noiseScale;
-	state_m->vq = vq_tmp;
+
+	// Check that the mode is turned on and if the D channel is active
+	if (conf_now->foc_prbs_mode == PRBS_MODE_VOLTAGE &&
+	    (conf_now->foc_prbs_channel == PRBS_CHANNEL_D || conf_now->foc_prbs_channel == PRBS_CHANNEL_DQ)) {
+		state_m->vd = vd_tmp + noise * noiseScale;
+	} else {
+		state_m->vd = vd_tmp;
+	}
+
+	// Check that the mode is turned on and if the Q channel is active
+	if (conf_now->foc_prbs_mode == PRBS_MODE_VOLTAGE &&
+	    (conf_now->foc_prbs_channel == PRBS_CHANNEL_Q || conf_now->foc_prbs_channel == PRBS_CHANNEL_DQ)) {
+		state_m->vq = vq_tmp + noise * noiseScale;
+	} else {
+		state_m->vq = vq_tmp;
+	}
 
 	utils_saturate_vector_2d((float*)&state_m->vd, (float*)&state_m->vq, max_v_mag);
 
--- a/mcpwm_foc.c
+++ b/mcpwm_foc.c
@@ -294,6 +294,38 @@ static volatile bool pid_thd_stop;
 #define M_MOTOR(is_second_motor)  (((void)is_second_motor), &m_motor_1)
 #endif
 
+// This is the noise bit for the PRBS generator
+static bool prbsNoiseBit;
+
+bool mcpwm_foc_get_prbs(void) {
+   return prbsNoiseBit;
+}
+
+/**
+ * @brief prbsGenerator Generates a -1/+1 noise signal from a pseudo-random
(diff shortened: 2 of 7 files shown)


mc_interface.c
```diff
@@ -97,7 +97,7 @@ static volatile motor_if_state_t m_motor_2;
 #endif
 
 // Sampling variables
-#define ADC_SAMPLE_MAX_LEN		2000
+#define ADC_SAMPLE_MAX_LEN		2048
```
kubark42, line 100: This has a big memory hit, as all the below arrays grow 2%.

kubark42, line 100: This has a big memory hit, as all the below arrays grow 2%.

mc_interface.c
```diff
@@ -108,6 +108,7 @@ __attribute__((section(".ram4"))) static volatile uint8_t m_status_samples[ADC_S
 __attribute__((section(".ram4"))) static volatile int16_t m_curr_fir_samples[ADC_SAMPLE_MAX_LEN];
 __attribute__((section(".ram4"))) static volatile int16_t m_f_sw_samples[ADC_SAMPLE_MAX_LEN];
 __attribute__((section(".ram4"))) static volatile int8_t m_phase_samples[ADC_SAMPLE_MAX_LEN];
+__attribute__((section(".ram4"))) static volatile int16_t m_frf_samples_mV[ADC_SAMPLE_MAX_LEN];
```
kubark42, line 111: I don't love this. It's a large memory allocation for a feature which won't be used often.

mc_interface.c
```diff
@@ -2553,7 +2555,7 @@ static THD_FUNCTION(sample_send_thread, arg) {
 		}
 
 		for (int i = 0;i < len;i++) {
-			uint8_t buffer[40];
+			uint8_t buffer[39];  // 1 + 4 + 4 + 4 + 4 + 4 + 4 + 4 + 4 + 4 + 1 + 1
```
kubark42, line 2558: I am confused why this value was 40. I counted the bytes and with the addition of 4 bytes it should now be 39. Could someone double-check?

mcpwm_foc.c
```diff
@@ -3778,22 +3811,32 @@ static void control_current(volatile motor_all_state_t *motor, float dt) {
...
-	float vq_presat = state_m->vq;
-	utils_truncate_number_abs((float*)&state_m->vq, max_vq);
-	state_m->vq_int += (state_m->vq - vq_presat);
+	// max_vq^2 + vd_tmp^2 = max_v_mag^2, where v_d and max_v_mag come from prior in the code
+	float max_vq = sqrtf(SQ(max_v_mag) - SQ(vd_tmp));
+	float vq_presat = vq_tmp;
+	utils_truncate_number_abs((float*)&vq_tmp, max_vq);
+	state_m->vq_int += (vq_tmp - vq_presat);
+
+	float noiseScale = conf_now->foc_hfi_voltage_max;
```
kubark42, line 3834: Hijacking other system configurations for fun and profit!

mcpwm_foc.c
```diff
@@ -3823,10 +3866,6 @@ static void control_current(volatile motor_all_state_t *motor, float dt) {
 	float mod_beta  = c * state_m->mod_q + s * state_m->mod_d;
 
 	update_valpha_vbeta(motor, mod_alpha, mod_beta);
-    
-    // Dead time compensated values for vd and vq. Note that these are not used to control the switching times. 
-	state_m->vd = c * motor->m_motor_state.v_alpha + s * motor->m_motor_state.v_beta;
```
kubark42, line 3828: Is it okay to remove this? The variables are never used, and clobbering the variable which is actually sent downstream for controlling the PWM, in favor of this one, seems like the wrong thing to do.

If it's necessary to keep this `vd` around, it might be best to give it another name.

mcpwm_foc.c
```diff
@@ -286,6 +286,38 @@ static volatile bool pid_thd_stop;
...
+/**
+ * @brief prbsGenerator Generates a -1/+1 noise signal from a pseudo-random
+ *        binary sequence with an 11-bit period. Every time this function is
+ *        called, the binary sequences advances one tick.
+ *        C.f. https://blog.kurttomlinson.com/posts/prbs-pseudo-random-binary-sequence
+ *        C.f. https://en.wikipedia.org/wiki/Pseudorandom_binary_sequence
+ *        C.f. https://en.wikipedia.org/wiki/Linear-feedback_shift_register#Example_polynomials_for_maximal_LFSRs
+ * @return -1 or 1
+ */
+int32_t prbsGenerator11_increment(void) {
```
kubark42, line 305: `static`

kubark42: I also don't love this being where it is in the code. I feel this should be migrated to `utils.c` or something like that.

mcpwm_foc.c
vedderb: It might be better to rename the other one to vd_control or this one to vd_measured. The measured value is used in vesc tool and the detection code, where measuring it will give slightly better results.

Is the set voltage better to use for FRF than the measured one?

vedderb:
I think it is better to make a union with the sample buffers and the FRF buffer as they most likely won't be used at the same time. Either you do sampling or FRF analysis. Then they can be floats and the averaging can be done directly in the firmware.

For the injection, I think a dedicated command would be better. It can send all parameters (noise magnitude, averaging samples), then the injection and averaging can be done and finally the result is sent back. I can look into that. As this works according to your tests I can reproduce it quite easily.

mcpwm_foc.c
ElwinBoots: > Is the set voltage better to use for FRF than the measured one?

Yes, it is essential for the calculation of the loops that it is the setpoint itself.

ElwinBoots:
Yes that sounds good Benjamin. We what we also would need ot the ability to change the injection point. 
1. Vd
2. Vq
3. Iq

Tracing is needed of the noise itself, the signal directly after the sum of the injection, id, iq and pid loop error. The currents can also be just the already available i1 and i2 and the phase angle (this I used untill now).

Averaging in the vesc would be really helpful. We have to remember about the bit depth, or maybe the the averaging might overflow.

Making it a function sounds like a good idea. Would it then also be possible to have a constant velocity while measuring?

mcpwm_foc.c
kubark42: I'll rename the original to `vd_control` and the other to `vd_measured`.

kubark42:
> I think it is better to make a union with the sample buffers and the FRF buffer as they most likely won't be used at the same time. Either you do sampling or FRF analysis. 

There might be a better way to resolve this, but it involves kicking the can down the road.

I would like to see the firmware be able to accept an arbitrarily large data request from the UI, and then it simply streams in batches until all requested samples are sent. This would lead to the firmware buffers decreasing in size, freeing up RAM.

In addition, the UI would format its command for which fields it wanted to query. This would turn the buffer of dedicated arrays into a generic buffer which can be packed arbitrarily.

> Then they can be floats and the averaging can be done directly in the firmware.

Averaging in the firmware will be somewhat complex because it either requires a new subsystem which knows how to wait until the average is complete, or it requires more RAM dedicated to the task. Considering that the value of FRF comes from averaging several fields simultaneously, it might be reasonable to average all values, e.g. `i_d`, `i_q`, etc... However, it might also be the case that the analysis requires the latest versions of some values and the averaged of the others. 

The upshot is it might be much better to implement the strategy described above and do all this in the UI.

vedderb:
> I would like to see the firmware be able to accept an arbitrarily large data request from the UI, and then it simply streams in batches until all requested samples are sent. This would lead to the firmware buffers decreasing in size, freeing up RAM.

How would you imagine that? USB is not fast enough to stream all samples in real-time, so things that have to be samples simultaneously would need some clever way to do that. Possibly rely on assumptions that things are periodic and pick the same things over the period repetition by repetition. I don't really like that...

> Averaging in the firmware will be somewhat complex because it either requires a new subsystem which knows how to wait until the average is complete, or it requires more RAM dedicated to the task. Considering that the value of FRF comes from averaging several fields simultaneously, it might be reasonable to average all values, e.g. i_d, i_q, etc... However, it might also be the case that the analysis requires the latest versions of some values and the averaged of the others.

I think it is perfectly fine to make the firmware aware of how to do the averaging. That is already done in a way with the sampled data and the different trigger types, decimation and sample lengths. The FRF sampling would look very similar I think: Mode (Injection point, other modes?), noise amplitude, averaging iterations.

> Making it a function sounds like a good idea. Would it then also be possible to have a constant velocity while measuring?

As with the sampling, it could be done under any running condition.

> Averaging in the vesc would be really helpful. We have to remember about the bit depth, or maybe the the averaging might overflow.

If we use floats that should be fine, but the buffers will require more bits per sample. Now it uses int16 to conserve memory.

> Tracing is needed of the noise itself, the signal directly after the sum of the injection, id, iq and pid loop error. The currents can also be just the already available i1 and i2 and the phase angle (this I used untill now).

Using the same amount of memory as the sampling in a union, we can do the following:

1x 32-bit float for the signal after injection (vd, vq, id or iq I assume?)
1x 32-bit float for id
1x 32-bit fload for iq
1x 32-bit float for pid_error (do we need both id and iq, or just the one we inject on?)
1x 8-bit position
1x 8-bit with some other state (like the noise bit), or use 16 bits for position

Averaging can be done on all of (or any combination of) the floats. Elwin, do you think that would work?

ElwinBoots:
Yes that sounds perfect.

For measured current in principle only id or iq is fine, depending on where you put the noise indeed.

I think the streaming idea could work if the full 12 Mbit/s is available: 6 floats at 60 kHz. But I am not sure if the full bandwidth can be used.

kubark42:
> How would you imagine that? USB is not fast enough to stream all samples in real-time...

I think it depends on how many samples we want to send, and how frequently. We might find that we can use two buffers on the VESC and transfer several buffers in a row before we get overrun by data.

On an optimized full-speed bulk transfer operation, the USB bus can realistically transfer a sustained 1MB/s. Of course, that's not our case, so let's downrate by at least 90%, leaving 100kB/s of streaming. If the system is filling the buffer at 50kHz, then this is ~2 bytes per cycle continuous. 

That doesn't sound like a lot, but if we are only need to transfer four int16 per cycle, then that means only a 2x speed differential. In the time it takes to fill two buffers of size X, we would have transmitted a full buffer's worth. This means we could get three buffers for the price of two. And if we slow down to 25kHz, or if we can get 20% of best real-world throughput, then we can do real-time for four int16.

By custom-tailoring the streaming output so only the values of interest are sent, and by aggregating data for large bulk transfer efficiency can be improved. I feel like this data budget could be stretched quite some distance.

Do you know what the realistic max throughput of the VESC is?

> make the firmware aware of how to do the averaging.

There's nothing wrong with the processor doing the average, per se. However, for me, if math doesn't need to be done on the device I like to avoid doing it there. It's simultaneously the hardest platform to support and debug, the most fragile and prone to unforeseen interactions, and the highest CPU cost per operation. The one advantage it holds-- which is nothing to sneeze at-- is that it's the one generating all the data.

Still, since the device won't consume the FRF data, I think it's better if the user has the flexibility to process the data in any way s/he would like. For instance, a mean-value windowed filter is one kind of average, but there are many others as well which might be interesting. It'll be easiest from a firmware dev perspective if the user gets the data directly, and the dev doesn't have to support all the mathematical operations a data scientist can dream up.

> This [averaging] is already done in a way with the sampled data and the different trigger types, decimation and sample lengths. The FRF sampling would look very similar I think...

I agree, with the caveat that the other averages are done in rolling real-time and so for those the answer is ready on demand. In this case, the averaging would either be done in a rolling manner, requiring N*2048 floats in RAM, or it would have to delay the transmission of data until the average is complete. Neither is impossible, but I think they're both more complex than bulk streaming the data in realtime.

vedderb:
The USB speed varies on different systems and for some reason it is much slower on windows. Also, I often only have access over CAN or UART, which would make this useless unless the buffers can store almost all samples. If this relies on quite fast USB it already is useless on half of my setups where I can't access the USB port easily. Some of my larger controllers rarely have access to the USB port and I do almost all debugging over CAN, and there it probably is even more interesting to run FRF.

> In this case, the averaging would either be done in a rolling manner, requiring N*2048 floats in RAM, or it would have to delay the transmission of data until the average is complete. Neither is impossible, but I think they're both more complex than bulk streaming the data in realtime (if in the end that's possible).

Not sure I'm following here. There are at least two ways to do the average without storing all samples:

1) Clear the samples in the beginning, add every new sample to each value in the buffer and divide by the number of samples when done.

2) Add to every sample using something like UTILS_LP_FAST and send them as they are when done.

Both of these sound much simpler to me than finding a good way to stream data with efficient ring buffers in-between.

I think general fast data sampling support for analysis is a different topic, and I think it takes much more effort making that generic and well working rather than just focusing on the FRF now. This depends so much on the specifics of the firmware anyway, especially with how the noise is injected.

kubark42:
> Clear the samples in the beginning, add every new sample to each value in the buffer and divide by the number of samples when done.

This requires delaying until the average is complete. I don't see a problem with that, but it is an extension of the code and it forces the developer to make choices about what to do with the other samples which are gathered at the same time. Should they also be averaged too, or should the last one be sent, or perhaps the first one? My intuition is that all those answers will be limiting in some way.

> Add to every sample using something like UTILS_LP_FAST and send them as they are when done.

I suggested that to @ElwinBoots, but he said a low-pass wouldn't work. I'll let him chime in here if he sees it any differently now that he's gotten some experience with the feature.

> I think general fast data sampling support for analysis is a different topic, and I think it takes much more effort making that generic and well working rather than just focusing on the FRF now.

Agreed. The only way it impacts FRF is deciding where is the best use of effort for implementing additional features in firmware. Sounds like you'd like to use FRF on systems with low speed telemetry busses, so I agree that real-time streaming isn't a solution.

FWIW, it might make sense to put this aside for the moment. @ElwinBoots and I have managed to get our results by simply clicking the UI's "Send" button multiple times and averaging the results. It's not ideal, but it works well enough that I think the question of how to efficiently transmit/average could be moved to a different feature PR. It might not make sense for us to tackle that immedaitely because we're unfamiliar with the telemetry system and could make erroneous assumptions (like the USB-only streaming).

kubark42:
@vedderb, thoughts on moving forward with the telemetry as is? If that's okay, then the only thing which needs fixing at this moment is to not hijack the HFI settings.

vedderb:
For now I'm focusing on testing and finishing the 5.03 release, then I will revisit this and get it implemented with filtering in the firmware probably.

kubark42:
We figured out the telemetry side of things. We'll test it this weekend and then upload the changes if everything works out.

We're also testing triggering multiple trace grabs, in order to allow certain types of averaging.

PR #377, opened 2021-11-12
Minor adjustments for HW_RD2 & the build tool-chain

rodale75:
For the tool chain, turned of verbose build output and added a target size rule. For the hw_rd2, naming adjustments of the inclusion guard and the get_temperature function (old legacy names, not valid for this target), and some minor changes/cleanup of the accompanied config params.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ endif
 
 # Enable this if you want to see the full log while compiling.
 ifeq ($(USE_VERBOSE_COMPILE),)
-  USE_VERBOSE_COMPILE = yes
+  USE_VERBOSE_COMPILE = no
 endif
 
 # If enabled, this option makes the build process faster by not compiling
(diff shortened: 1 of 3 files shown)


kubark42:
@bmellstrom-radinn, could you give an example of what the `size` rule is for?

bmellstrom-radinn:
@kubark42 It's basically just a tab-completable shortcut for the following command:
<pre>
$ arm-none-eabi-size build/BLDC_4_ChibiOS.elf 
 303104	   3436	 143148	 449688	  6dc98	build/BLDC_4_ChibiOS.elf
</pre>
It's nice to have when you want to quickly iterate between different code options and see how it affects the flash and ram size. Text means code in this case, data is non-zero-initialized memory and bss is zero-initialized memory. So text+data will determine how much flash (rom) memory is required and data+bss determines how much ram memory is required. Hope this helps!

PR #378, opened 2021-11-13
SVM sign fix

ElwinBoots:
The original SVM code output was (1-dutycycle) instead of dutycycle, while the code needs dutycycle.
To make the math correct this was "corrected" by having the mod_alpha and mod_beta inputs negative.

This proposal now instead really fixes the output to dutycycle. With this code the input should not be taken negative anymore.

Tested in simulation and on VESC 6 MKV with motor. As expected, no difference found.

Renamed the PWMHalfPeriod variable as this confused me into thinking it is somehow the half duty cycle, while it really is the peak.


kubark42:
Much better, me likes!

PR #379, opened 2021-11-16
Increase max imu rate

Mitchlol:
Small PR to increase the max allowed IMU update rate. I've found I can get a slight performance gain with a higher rate. Some of the supported IMUs have output data rates up to 8khz so I figured I'd set the max to 10khz.

### DIFF
--- a/imu/bmi160_wrapper.c
+++ b/imu/bmi160_wrapper.c
@@ -80,9 +80,12 @@ static bool reset_init_bmi(BMI_STATE *s) {
 	}else if(s->rate_hz <= 800){
 		s->sensor.accel_cfg.odr = BMI160_ACCEL_ODR_800HZ;
 		s->sensor.gyro_cfg.odr = BMI160_GYRO_ODR_800HZ;
-	}else{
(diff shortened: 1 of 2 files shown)


vedderb:
I think it is better to have an enum like this when allowing so high rates:
https://github.com/vedderb/bldc/blob/dev_fw_5_03/mcpwm_foc.c#L3541
As the sleep time always is a multiple of CH_CFG_ST_FREQUENCY the higher rates will be rounded to the closest value and the steps are large. E.g. the next value after 10k will be 5k, then 3.3k, then 2.5k etc.

Also, it would be good to limit the maximum rate for each IMU depending on the interface (SPI, I2C), what it supports and how much CPU overhead it gives.

Running the i2c_bb driver at 10 kHz might consume all CPU cycles, so before allowing that it would be good to test it. One way to deal with the faster IMUs would be to update the i2c_bb driver to allow a configurable i2c_delay so that the bus runs faster and the code spends less time sleeping (which is where it spends the most time). The reason that the SW SPI driver spends so much less cycles is that it clocks faster and spends less time sleeping.

PR #392, opened 2021-12-28
Luna bbshd support

nitrousnrg:
Here is the code support for the Luna Cycle  bbshd[ ludi v2](https://lunacycle.com/blog/luna-bbshd-ludicrous-v2-released/) controller
image

The big contribution is the code for the display. It provides basic support for Bafang serial protocol, and most notably it works with Eggrider displays over UART:
image
I have not implemented the packets to read/write settings as vesc settings are vastly more advanced than bafangs so it made little sense to code that, but for example the eggrider bluetooth app can freak out when it tries to load controller settings.

These displays show:
* Speed
* Battery current
* Battery voltage
* Battery level
* PAS level

They also have + - buttons to set the assist level from 0 to 9 which affects the max phase current that can be delivered from pedal assist.

These displays are not just plug&play, they need a special connector, a >12V supply (bagang needs way more volts) some conditioning on the uart lines and a circuit to turn it on.

The cadence-based pedal assist got more rugged against shocks and its worth noting that [here](https://github.com/vedderb/bldc/compare/dev_fw_5_03...luna-cycle:luna_bbshd_support?expand=1#diff-8e566af6be43fda0f0f76cdf994f9608bccf81f89d3ffd017485e843f8cc712bR232) I inadvertently added a hook for torque-based pedal assist that is coming in the next PR.

This PR also includes a default motor config for the BBSHD drive unit and the luna QML image


r3n33:
I'm excited to see the Ludi v2 PR! Are there plans to support the display of the trip or odometer values?

nitrousnrg:
> I'm excited to see the Ludi v2 PR! Are there plans to support the display of the trip or odometer values?
As far as I know, the display keeps track of odometer values based on the wheel speed reported by the motor controller, so no need to send odometer data from the controller.

r3n33:
Thank you. I'm just learning how these displays work. In my case (X2 with Ludi v2) the C240 display seems to retrieve the wheel diameter from the controller. Would that be required for the display to calculate distance over time? :thinking:

nitrousnrg:
> Thank you. I'm just learning how these displays work. In my case (X2 with Ludi v2) the C240 display seems to retrieve the wheel diameter from the controller. Would reporting the wheel diameter be required for the display to calculate distance over time? thinking

Ah yes, you have the M600 V2 controller, this pull request is for the BBSHD V2 controller, they are quite different controllers. BBSHD display uses serial communication and M600 display uses CANbus and does require the controller to send odometer data to the display.

The M600 code sits on top of the BBSHD code so if we can get this merged I can work on the M600 PR. If I did something wrong with the BBSHD its likely its also wrong in the M600 code. For example the directory where the qmlui is placed, etc. The display code is coded like a hardware-specific thing because I think its not suitable for a general purpose controller, but someone might argue against that.

vedderb:
Nice work!

I moved all luna-files to one place, hope that makes it easier. Also changed the qmlui to HW instead of APP as I think it is specific to this hardware.

Did you forget to include the appconf in the pr?

nitrousnrg:
Thanks! Will take a look at the qmlui as I tried to have it in HW but didn't work out back in that time.

Ugh, yes I forgot to add the appconf, will set up another PR for that one

PR #393, opened 2021-12-29
[MCPWM_FOC] Make rotor lock feature use D axis

kubark42:
Changes rotor lock feature from Q axis to D axis. This has two advantages:
1. Now the angle you put in is the angle you really get in the alpha beta frame (no 90 degree offset).
2. Now when the rotor is being locked, the D axis is really the D axis and Q is really the Q axis. This helps for FRF measurements.

HT to @ElwinBoots

Tested extensively by @mxlemming and @ElwinBoots.

### DIFF
--- a/mcpwm_foc.c
+++ b/mcpwm_foc.c
@@ -890,7 +890,8 @@ void mcpwm_foc_set_pid_pos(float pos) {
 void mcpwm_foc_set_current(float current) {
 	motor_now()->m_control_mode = CONTROL_MODE_CURRENT;
 	motor_now()->m_iq_set = current;
-
+	motor_now()->m_id_set = 0;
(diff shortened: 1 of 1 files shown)


Guillaume227:
For consistency, should m_iq_set and m_id_set be swapped when locking the rotor in mcpwm_foc_measure_resistance?

kubark42:
> For consistency, should m_iq_set and m_id_set be swapped when locking the rotor in mcpwm_foc_measure_resistance?

Good catch. @ElwinBoots, what's your take on that? It certainly makes sense to me, and from a user perspective it shouldn't make any tangible difference.

ElwinBoots:
Yes that would make sense.

PR #395, opened 2022-01-03
Balance App: bugfix for speed based variable tiltback config

surfdado:
Avoid division by zero

### DIFF
--- a/applications/app_balance.c
+++ b/applications/app_balance.c
@@ -214,7 +214,11 @@ void app_balance_configure(balance_config *conf, imu_config *conf2) {
 
 	// Variable nose angle adjustment / tiltback (setting is per 1000erpm, convert to per erpm)
 	tiltback_variable = balance_conf.tiltback_variable / 1000;
-	tiltback_variable_max_erpm = fabsf(balance_conf.tiltback_variable_max / tiltback_variable);
+	if (tiltback_variable > 0) {
+		tiltback_variable_max_erpm = fabsf(balance_conf.tiltback_variable_max / tiltback_variable);
+	} else {
+		tiltback_variable_max_erpm = 100000;
+	}
 
 	// Reset loop time variables
 	last_time = 0;


applications/app_balance.c
```diff
@@ -214,7 +214,10 @@ void app_balance_configure(balance_config *conf, imu_config *conf2) {
 
 	// Variable nose angle adjustment / tiltback (setting is per 1000erpm, convert to per erpm)
 	tiltback_variable = balance_conf.tiltback_variable / 1000;
-	tiltback_variable_max_erpm = fabsf(balance_conf.tiltback_variable_max / tiltback_variable);
+	if (tiltback_variable > 0)
```
kubark42, line 217: I don't know the project's coding style standards, but in embedded I would always avoid non-braced `if().

```
if (tiltback_variable > 0) {
	tiltback_variable_max_erpm = fabsf(balance_conf.tiltback_variable_max / tiltback_variable);
} else {
	tiltback_variable_max_erpm = 100000;
}
```

Adds only one line and prevents bad things from happening if someone misses that there's no bracing while adding an extra line to the `else` clause.

surfdado: Fixed

applications/app_balance.c
```diff
@@ -214,7 +214,11 @@ void app_balance_configure(balance_config *conf, imu_config *conf2) {
 
 	// Variable nose angle adjustment / tiltback (setting is per 1000erpm, convert to per erpm)
 	tiltback_variable = balance_conf.tiltback_variable / 1000;
-	tiltback_variable_max_erpm = fabsf(balance_conf.tiltback_variable_max / tiltback_variable);
+	if (tiltback_variable > 0) {
+		tiltback_variable_max_erpm = fabsf(balance_conf.tiltback_variable_max / tiltback_variable);
+	} else {
+		tiltback_variable_max_erpm = 100000;
```
kubark42, line 220: When using a magic number, it's good to justify why this number and not another. Is `100000` chosen at (almost) random and `99999` and `100001` work just as well? Or is it very sensitive and was the result of 15 minutes of sitting at the bench playing with numbers to get something which was just right?

(Sorry, I should have suggested this earlier)

surfdado: It's just a high number - what's your suggestion?

surfdado: There will be no tiltback anyway because tb_variable is 0, so the number doesn't really matter too much - but it ensures that the condition return false each time when variable tiltback is disabled. If anything, higher would be better

surfdado: I guess MAX_INT would be the right choice here...

kubark42: I don't have a suggestion, bit based on what you say I'm sure any number would work fine. You might add that as a comment:

```
tiltback_variable_max_erpm = 100000;  // This number was chosen without any particular care. It's big without being obnoxiously hard to write.
```

That might sound a little silly, but it really dives home to readers that this is a variable they can change without worrying it will upset something sensitive.

kubark42: > I guess MAX_INT would be the right choice here...

Mmm, yeah, that's always a good one for this kind of thing.

PR #405, opened 2022-01-21
Encoder split

Jaycobe96:
In order to make further work on a particular encoder and adding new encoders easier, the old encoder.c file has been split such as every type of encoder has its own .c and .h files. All encoder files are wrapped in new encoder.c file. Encoder files and encoder.c wrapper are placed in encoder directory.


Teslafly:
Is there any sort if forum discussion for this?
Or at least a description that's basically what you would put in the changelog?

It looks like you are just splitting up the encoders for easier readability?
Are there any other justifications or functionality improvements?

Teslafly:
Looks like this is probably a continuation of https://github.com/vedderb/bldc/pull/386 on the new branch.

Shackmeister:
Hi @Teslafly 
The refactoring is for easier readability, and easier addition of new encoders.
next step is to allow for multiple instances of encoders.

The plan for the refactor was made together with Vedder

Teslafly:
Glad to hear it. 

I just really don't like pull requests without a description. Too often I've seen or had to deal with someone just throwing a pull request your way and expecting you to merge it with no explanation. and then being salty when you don't.

No PR message also doesn't help those of us who want to review (say because we are also working on encoder things), but weren't privy to the conversation you had with vedder.

or if this is squashed and the PR message becomes the commit message.

Shackmeister:
Hi

Makes perfect sense, I'll have Jakub update the PR with a proper
description :)
Are you also working on encoder code?
Goal for this PR is to split the encoders tl seperate files (inspired by
IMU structure) and also unify all the software SPI to a common library.

Kindest regards
Frank Severinsen

lør. 22. jan. 2022 01.29 skrev Teslafly ***@***.***>:

Teslafly:
I had made a pcb for the TLE5012 encoder (https://github.com/Teslafly/TLE5012-Hall-Encoder-Module) and had planned on adding support for it to vesc. Needs hw spi support. Though I've been dragging my feet on it as other things have taken priority. I only got as far as duplicating the encoder functions and adding register definitions.

I also have a suggested change for future vesc hw hall pins. You can fit both the abi encoder timer and hw spi on the same 3 hall pins. MOSI can be set to bidirectional for ssc interfaces and can also support plain data receive (AS5047).

- hall3 = pb4 (tim3-1 A, csq/n) alt miso,
- hall2 = pb5 (tim3-2 B, dat/mosi,) 
- hall1 = pb3 (timrst, I, sck) alt swo

```
// Hall/encoder pins
#define HW_HALL_ENC_GPIO1        GPIOB
#define HW_HALL_ENC_PIN1        3
#define HW_HALL_ENC_GPIO2        GPIOB
#define HW_HALL_ENC_PIN2        5
#define HW_HALL_ENC_GPIO3        GPIOB
#define HW_HALL_ENC_PIN3        4
#define HW_ENC_TIM				TIM3
#define HW_ENC_TIM_AF			GPIO_AF_TIM3
#define HW_ENC_TIM_CLK_EN()		RCC_APB1PeriphClockCmd(RCC_APB1Periph_TIM3, ENABLE)
#define HW_ENC_EXTI_PORTSRC		EXTI_PortSourceGPIOC
#define HW_ENC_EXTI_PINSRC		EXTI_PinSource8
#define HW_ENC_EXTI_CH			EXTI9_5_IRQn
#define HW_ENC_EXTI_LINE		EXTI_Line8  //check/fix
#define HW_ENC_EXTI_ISR_VEC		EXTI9_5_IRQHandler
#define HW_ENC_TIM_ISR_CH		TIM3_IRQn
#define HW_ENC_TIM_ISR_VEC		TIM3_IRQHandler
```

This slightly changes the pinout for ABI encoders, but all other sensors remain the same.

Shackmeister:
Very nice
Me and Jakub also looked into supporting TLE5012B. Is it strictly nescessary to use HW SPI for it? On a seperate project we made a 3 pin BB SPI due to an issue in the stm driver (F7).
I'm hoping at some point to make it easier to switch between HW and SW SPI

Teslafly:
It's probably not strictly necessary. but it requires bidirectional communication and each transaction is 48 bits long  (16 bit command, 16 bit data, 16 bit safety word/crc). The max spi frequency is 8mhz. The MT6816 is a similar sensor and it only supports hw spi (though not the single data line ssc interface).
I would worry about the software spi thread starting to take up too much time. It would probably work at low rpm's with a low sensorless switchover speed though.

nitrousnrg:
My 2c: I have a strong preference towards HW SPI because I find difficult to provide design assurance on a sensorless based system, so most -if not all- my controllers run sensored from 0 to max RPM, and bitbanging the SPI does take considerable time.

image

When I took that capture years ago the cpu usage of the bb spi was 1/3 of the time it took to run the full FOC loop. That's a slow IGBT system switching at 10kHz, I usually do 21kHz so its nice to see the peripheral offloading the cpu :)

vedderb:
When it comes to hw spi you cannot just start the hw and wait for it to finish as the CPU will still spend the same time in a busy wait loop instead of using the same cycles to move the bits. To actually offload the cpu you would have to implement the spi for each encoder using interrupts and/or dma and coordinate with the FOC code, which makes the implementations much more hw specific and complex.

In the majority of cases sensorless works better at higher speeds than using an encoder (e.g. the AS5047P only works to 28k RPM and the AS5047D to 14K RPM). The intention is to use encoders at lower speed and read them with lower priority than the FOC interrupt at 5-10 kHz, then switch to sensorless at higher speed. That makes the burden of implementing encoder support much smaller than writing very hardware specific DMA/interrupt code for each encoder. Dealing with the different/lack of hw pin support for all already existing hardware out there in the driver makes this an even bigger task.

With the split in the driver it leaves room for making hw-accelerated drivers for suitable encoders separately and leave the existing ones with sw support at low speeds intact and avoids breaking things for other people with different hw versions. One example of an encoder that has full hw support and samples at the full speed is the quadrature (ABI) encoder, which would work as it is if you really need to run on an encoder at full speed.

Teslafly:
Hardware accelerated drivers running at control loop speed would be ideal where it can be supported. These hall encoders are much more desirable than the abi encoder since they don't have to move the motor to zero out on boot. They can give you an absolute angle right away.
Hopefully this encoder refactoring can make it less of a pain to implement/maintain control loop coupled encoder polling.

On difficult motors I've definitely had to turn up the sensored to sensorless switchover erpm when the observer had trouble tracking the motor even at moderate speeds. Though this was on v4.2 hw with 2x shunts. Control loop tuning is also better now.

----------------
On another note, It would definitely be nice to be able to choose sw vs hw spi/ssc without recompiling the firmware. If you don't have access to the hw spi pins (using uart/adc with current pinout) you can use sw spi and just make your starting/low speed performance better. But if you have access to hw spi pins you can get hall encoder position potentially at control loop speed. (for properly hw accelerated drivers)

Also being able to choose between spi 4 wire (with mosi and miso) and ssc 3 wire (1 bidirectional data line) without recompiling would be a good goal for the MT6816 and TLE5012. Currently the MT6816 only supports 4 wire spi.
Though pin assignments get a bit messy for each hardware when you effectively have 4 ways of encoder communication using different peripherals.

Being able to switch these is possibly something to keep in mind on the vesc-tool side of this?

Jaycobe96 (commented):
All encoders implemented, AS504x working (tested)
TODO tests: 
-ABI 
-AD2S1205
-ENC_SINCOS
-MT6816
-TS5700N8501

Jaycobe96:
I am not sure why i get this warning
_encoder/encoder.c:35:1: warning: missing braces around initializer [-Wmissing-braces]_

Teslafly:
It would be nice if the encoder files could have the functions in the same order. 
to me, this makes the most sense:

1. init
2. deinit
3. routine
4. helper functions
5. getter functions

Though as long as they are consistent, order probably isn't that important.
If you want, this is what I did in my branch: https://github.com/Teslafly/vesc_firmware/tree/frankentest_tle5012_dev

PR #410, opened 2022-02-10
Support for v0.9 version of the Cheap FOCer

surfdado:
90% of Cheap FOCers in circulation are v0.9, due to chip
shortage this won't change anytime soon...

To build v0.9 do #define CFOC_IS_V09 in conf_general.h
Rebuild all now includes a V09 version as well.


hwconf/hw_Cheap_FOCer_2.h
```diff
@@ -20,7 +20,12 @@
 #ifndef HW_Cheap_FOCer_2
 #define HW_Cheap_FOCer_2
 
+
+#ifdef CFOC_IS_V09
+#define HW_NAME					"Cheap_FOCer_2_v0.9"
```
vedderb, line 25: The dot will not work with vesc tool, it has to be the same name as the directory name for vesc tool to list the correct firmware if the firmware is to be included in vesc tool tool.

surfdado: Gotcha, will fix it

Teslafly:
Since the structure of the mcconf and build all method had changed in the master branch dev firmware, this should be re targeted there and updated or the pr closed.

Teslafly:
If https://github.com/vedderb/bldc/pull/577 replaces this PR, should this be closed?

PR #413, opened 2022-02-11
Add the BissC encoder on HW SPI.

manoukianv:
Hello,

i had implemented a new interface encoder : BISSC to be compatible with a lot of industrial encoder.
A quick video here to explain how to : https://youtu.be/G4rkW7bhovg

i use the same implementation than TS5700N8501 and MT6816 : it uses COMM port to wires the encoder.
i push also a PR for the vesc tools. Both come together to synchronize MCCONF signature.

This encoder uses RS422, si it has be connected to the COMM port for high speed communication. A RS422-transceiver such as the MAX490 is required, where CLK and MISO are used as click and data input lines. To use this encoder, you have to make sure that no app uses UART or ADC1. The ABI resolution field is used to setted the BISSC encoder accuracy : 2^(BissC Resolution)

There 2 implementations selectable by a compilation flag : BISSC_ISR_IMPL
By default the BISSC interface uses ISR handle like ABI and other and i reduce the refresh rate at 10k/s
A second implementation is available by thread, like the TS5700N8501, sample rate is also 10k/s
Both works well but i don't know which is the better, so i let option available for technical design.

I Hope this can help community and increase motor compatibility.

Best regards
Vincent


vedderb:
Nice work!

Can you update the PR against the new encoder structure? We did a major rewrite, so it should be easier to add encoders from now on.

manoukianv:
Thanks.
I will migrate it on new architecture and push it ;)

manoukianv:
I have check the new architecture of encoder. I think i can migrate the bissc encoder on the new architecture.
I saw that the new rate for encoder via isr is 10000. The implementation i have done via isr works at this speed. I'll migrate the 2 impl in a first time (isr & thread).

manoukianv:
I close this PR and open a new one.

PR #424, opened 2022-03-01
[Package] Simple script to copy appropriate targets into package directory

kubark42:
The resulting output is packaged into a VESC-Tool release.

It requires a directory layout like https://github.com/vedderb/bldc/pull/423 produces.

It's output looks like: 

```
bldc/package/46/VESC_default.bin
bldc/package/48/VESC_default.bin
bldc/package/410/VESC_default.bin
bldc/package/410/VESC_default_no_limits.bin
bldc/package/60/VESC_default.bin
bldc/package/60/VESC_default_no_limits.bin
bldc/package/60_mk3/VESC_default.bin
bldc/package/60_mk3/VESC_default_no_limits.bin
bldc/package/60_mk4/VESC_default.bin
bldc/package/60_mk4/VESC_default_no_limits.bin
bldc/package/60_mk5/VESC_default.bin
.
.
.
```

### DIFF
--- a/package/package_firmware.py
+++ b/package/package_firmware.py
@@ -51,8 +51,12 @@ package_targets = ['46', '48',
 
 # Iterate over each target
 for target in package_targets:
-    # Set the target filename and destination
-    file_name = target + '-' + git_hash + '.bin'
+    # Set the target filename
+    if "_no_limits" in target:
+        file_name = "VESC_default_no_limits.bin"
+    else:
+        file_name = "VESC_default.bin"
+
     # Set the target destination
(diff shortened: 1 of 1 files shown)


Teslafly:
One thing I notice is that this is not OS independent. Is there any intention for this to work on windows at all? If so, using the python os fileutils to build the file path would be a good idea vs string concatenation.

kubark42:
> Is there any intention for this to work on windows at all?

It has an audience of one, but I don't see any reason to make it broken-by-design for windows. Let me build that path the way you suggest.

PR #426, opened 2022-03-03
xESC2 (+TMC6200 gate driver) support

ClemensElflein:
Hi,
I created a smaller, lower power, lower cost version of the VESC for my (hobby) robotics projects. Until now I used VESC 4 controllers, but they were too large and hard to use in robotics projects (you can't just put them on a PCB).

Unfortunately, due to the semiconductor shortage, the gate drivers used by the VESC are sold out and therefore I had to look for alternatives. I found Trinamic's TMC6200 IC would be perfect a perfect fit for my VESC build: It contains gate drivers and in-line shunt current amplifiers all in one package. The only drawback is, that it does not contain a step down 5V regulator.

Long story short: My design is ready and I implemented the ESC hardware config and the TMC6200 support in the VESC firmware. It would be great if you could merge this into the main firmware.

You can find the ESC here: https://github.com/ClemensElflein/xESC

BTW: thank you for the great project and this awesome firmware!


vedderb:
Thanks for the PR!

Can you make it against the beta release? Also, is it really necessary to include all of that code when only a fraction of it is used? It would be better to put only the functions and defines you need in tmc6200.c and tmc6200.h and leave out the rest of the files and code that never get used.

ClemensElflein:
Thank you for the feedback and sorry for the late reply. I will implement the changes and update the PR soon.

ClemensElflein:
I've modified the code so that it can be merged in the master branch. I also removed all compile time warnings (unused variables for terminal commands) and removed the unused Trinamic gate driver sources.
The firmware builds and works on the ESC as expected.
Currently I'm testing using the precompiled VESC tool, but I'm installing qt and maybe I get the development version of VESC tool to run. I don't think that any problems will arise, though.

danilolattaro:
Can this now merge to close the PR?

Teslafly:
This looks good for merging to me?

PR #427, opened 2022-03-06
Make update

vedderb:
Build system updates from @kubark42


package_firmware.py
```diff
@@ -21,64 +21,63 @@ def get_git_revision_short_hash() -> str:
 package_dict = {}
 package_dict["46_o_47"] = [['46', default_name],
                     ['46_33k', 'VESC_33k.bin'],
-                    ['46_0005ohm', 'VESC_0005ohm.bon']]
```
kubark42: Well, that's an embarrassing typo I made. Thanks for fixing it!

PR #428, opened 2022-03-06
[README] Update README.md instructions

kubark42:
This updates the readme to reflect https://github.com/vedderb/bldc/pull/423.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ There are also many other options that can be changed in [conf_general.h](conf_g
 ## Prerequisites
 
 ### On Ubuntu (Linux)/macOS
-- Tools: `git`, `wget, `make`, `python3`, and `pip`
+- Tools: `git`, `wget, and `make`
 - Additional Linux requirements: `libgl-dev` and `libxcb-xinerama0`
 - Helpful macOS tools: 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,155 +17,94 @@ This is the source code for the VESC DC/BLDC/FOC controller. Read more at
...
+6. Continue with [On all platforms](#on-all-platforms)
 
+### On all platforms
 
-```bash
-export PATH="$PATH:/Users/your-name/gcc-arm-none-eabi-8-2019-q3-update/bin/"
-```
+7. `git checkout origin/master`
+8. `make arm_sdk_install`
+9. `make qt_install`
```
vedderb, line 73: qt_install is probably good to move to an optional section as it isn't needed unless you want to build vesc tool or use qt creator. pip install aqtinstall can be moved there too. Seeing the pip-command in build instructions always puts me in fight-against-python-dependency-hell-mode.

kubark42: It sets up Qt Creator for firmware compilation. That's pretty handy if people don't already have an IDE that they prefer.

kubark42: Since Qt is necessary for the VESC-Tool, it makes sense to me to have it be part of a standard setup.

vedderb: I think it is quite common to just compile the firmware and use pre-compiled vesc tool, so it makes more sense to have them separate. Especially since qt is so huge compared to the other tools needed. I also wonder if this script has a way to get qt 5.12 as that is what I'm using, and if it gets all qt modules correctly.

kubark42: @vedderb fixed.

PR #429, opened 2022-03-06
[Unit tests][Utils.c] Fix unit test compilation

kubark42:
The test compilation were broken by the addition of some threading utilities.


kubark42:
@vedderb I wonder if we could simply move the thread utilities to their own separate file. Or perhaps it would make even more sense to move all the math functions out of `utils.c` and into a dedicated file strictly reserved for algorithms.

In fact, I like that second option a lot. It makes for a small math lib, and we can easily isolate and unit test those without worrying about functions which depend on the ChibiOS subsystem.

PR #436, opened 2022-03-07
Create hw_410_005ohm.h

adilrepas:
HW4 targets
vesc 410/412/420 with 5mohm shunt

### DIFF
--- a/hwconf/trampa/vesc4/hw_410_005ohm.h
+++ b/hwconf/trampa/vesc4/hw_410_005ohm.h
@@ -1,0 +1,26 @@
+/*
+	Copyright 2015 Benjamin Vedder	benjamin@vedder.se
+
+	This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


hwconf/trampa/vesc4/hw_410_005ohm.h
```diff
@@ -0,0 +1,26 @@
...
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+    */
```
kubark42, line 16: Outside the scope of this PR, but we really should get rid of this boiler plate from all of these header files. It has no legal validity, and it just gunks up text searches.

PR #437, opened 2022-03-09
[Utils] Return `bool` instead of `int`

kubark42:
No functional changes. Just makes the API contract a little clearer.

Of the 70-odd usages of these truncation functions, the return value is used in precisely 0 places, so this really has little to no effect on the output firmware.

Tested for sys ident and running on EDF, no differences noted.


kalvdans (approved):
Please document what the function is supposed to return in the header file, so that users can understand it without reading the definition.

kubark42:
> Please document what the function is supposed to return in the header file, so that users can understand it without reading the definition.

I'm not sure I completely follow. There is currently no documentation aside from the API signature, and I wholeheartedly agree with adding docu and am happy to do that if requested during review.

However, I rarely see docu in the header file, and then only as part of a comprehensive, holistic approach. Header docu would seem something which should be done project-wide, or not at all.

kalvdans:
> Header docu would seem something which should be done project-wide, or not at all.

I disagree, partial documentation is better than no documentation. But it doesn't have to be done in this PR :)

PR #442, opened 2022-03-12
Windows powershell fixes

kubark42:
This fixes some Windows build issues which were related to having MS Powershell (not installed by default) vs Windows Powershell (installed by default).

Tested on Windows 10 machines with and without WSL.

### DIFF
--- a/ChibiOS_3.0.5/os/common/ports/ARMCMx/compilers/GCC/rules.mk
+++ b/ChibiOS_3.0.5/os/common/ports/ARMCMx/compilers/GCC/rules.mk
@@ -312,6 +312,13 @@ clean:
 #
 # Include the dependency files, should be the last of the makefile
 #
--include $(shell mkdir .dep 2>/dev/null) $(wildcard .dep/*)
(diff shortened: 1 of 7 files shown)


ChibiOS_3.0.5/os/common/ports/ARMCMx/compilers/GCC/rules.mk
```diff
@@ -312,6 +312,13 @@ clean:
 #
 # Include the dependency files, should be the last of the makefile
 #
--include $(shell mkdir .dep 2>/dev/null) $(wildcard .dep/*)
```
kubark42, line 315: I really don't enjoy changing ChibioOS core files, this just makes life harder when upgrading ChibiOS.

kubark42:
Tested on Windows 10 machines with and without WSL.

kubark42:
@vedderb this is ready for merge.

dylancm:
tested after rebase without `pwsh` and without pip path stuff and it works end-to-end

PR #443, opened 2022-03-12
Make fixes

kubark42:
Minor Makefile fixes dealing with Python on all platforms and removing spurious warning when there is no ARM compiler installed.

Testing coverage:
- [x] Checked on Windows
- [x] Checked on POSIX


dylancm:
Verified this works without `aqt` being on my path.

PR #450, opened 2022-03-16
Split `utils.c/.h` into `math_utils.c/.h` and `sys_utils.c/.h`

kubark42:
This makes it a little clearer and enables unit testing on the pure math functions without having to struggle to compile the system functions.

A downside is that the `utils_` prefix has lost some of its meaning.

There are no functional changes to the code, this is simply light reorganization. I have tried to make it so that review is very easy by looking at the individual commits, so that the combined changeset is not overwhelming.

- [x] Passes unit testing
- [x] Passes hardware test
- [x] All targets updated to new code


kubark42:
Motor detects and runs fine.

RTM.

kubark42:
Passes `make all_fw`, good for review/merge!

PR #454, opened 2022-03-21
[Make][Windows] Force the use of `cmd` shell for ChibiOS rules.mk

kubark42:
This guarantees that it won't matter if Git for Windows's sh.exe is on the path and aliases `make`'s sh.exe.


kubark42:
This resolves a problem @Teslafly was having.

PR #459, opened 2022-03-25
`utils_saturate_vector_2d()` unit test

kubark42:
Implements a test for `utils_saturate_vector_2d()` 

Unit test passes:

```
[----------] 4 tests from Saturate2dVector
[ RUN      ] Saturate2dVector.ValsInRange
[       OK ] Saturate2dVector.ValsInRange (0 ms)
[ RUN      ] Saturate2dVector.ValsOnEdgeOfRange
[       OK ] Saturate2dVector.ValsOnEdgeOfRange (0 ms)
[ RUN      ] Saturate2dVector.ValsBeyondRange
[       OK ] Saturate2dVector.ValsBeyondRange (0 ms)
[ RUN      ] Saturate2dVector.ValsAreDegenerate
[       OK ] Saturate2dVector.ValsAreDegenerate (0 ms)
[----------] 4 tests from Saturate2dVector (0 ms total)
```

PR #460, opened 2022-03-25
[Utils] Streamline `utils_saturate_vector_2d()`

kubark42:
There was no need to do the `sqrt()` unless the absolute magnitude is exceeded.

Passes unit tests but shouldn't be merged until #459 is merged.


vedderb:
It looks like the worst case, when saturation is needed, will still take the same amount of time. Also, sqrtf is quite fast on the FPU.

For optimizations like this one it would be good to set up a way to profile how much difference they make. Maybe this can be done by running a test on the STM32 with some build option and plot the result in VESC Tool.

PR #462, opened 2022-03-31
Fix AS504x DIAG and MAGN addresses.

raidoz:
According to AS5047D datasheet (https://ams.com/documents/20143/36005/AS5047D_DS000394_2-00.pdf), DIAAGC is 0x3FFC and MAG is 0x3FFD, in enc_as504x.c they are, however, incremented by one, reading back MAG for DIAG and ANGLEUNC for MAG.


Teslafly:
Just checking, was this tested? If so using what setup?

vedderb:
I currently have no way to test this as I didn't add miso to any of my as5047 test boards. It would be good if someone who has been using the as504x with miso can give some feedback on this.

Teslafly:
Should this be included the the v6.0 firmware release?

Issue #319, opened 2021-07-14
Use realitive instead of absolute position for pid position control


Teslafly:
I see you have somewhat recently reworked/touched pid position control, so now may be a good time to bring this up.
Right now, pid position control normalizes all pid angles to a positive value 0-360 degrees. 

This leads to some issues where the motor tries to turn around completely when reaching the end of the 0-360 range, and also means that "position angle division" setting to make 360 degrees = 2+ turns does not work.

My application is using the vesc like a servo with an attached output gearbox. I need to turn the motor -1800 to +1800 degrees (4 turns) after setting a zero point with an endstop. (call: mcpwm_foc_get_tachometer_value(reset=True)) 

I have made a branch that implements this pid position behavior at https://github.com/Teslafly/bldc/tree/gearbox_servo_app_dev
See my changes to mcpwm_foc.c
I am using the as5047 spi hall encoder to track the motor position which gives good results.

### So basically:
`mc_interface_set_pid_pos()` should not use `utils_norm_angle(&pos);` 
You should be able to send a positive or negative value of any angle, even over 360 degrees, and be able to treat the vesc as a servo going to those values. 

"position angle division" > 1 should act as an angle multiplier, with "position angle division" = 7 making the motor spin 7 degrees for each 1 degree of command. (that, or if the angle is meant to be divided by it, the behavior is just inverted.)
This also removes any need for special behavior based on the value of `p_pid_ang_div`. As it is always tracking the motor in a relative way.

I am wondering what your thoughts are on this behavior change? Is there a reason you made it absolute originally? If we switch to relative behavior, do you want to do it? or I can make a pull request. It may take me a little while to update my method though.

supercrazysam:
Absolute positioning is very useful for robotics application. You can use absolute position value to do relative task, but using relative position to do absolute task would be really troublesome. VESC's native support for absolute encoder saved a lot of my projects, honestly I prefer absolute positioning behavior a lot more.

Also if you are looking for "multi-turn absolute" I saw and used a mod somewhere that does absolute positioning multi-turn.
You may want to look into how he changed the encoder.c
https://github.com/raess1/vesc-FW

I followed his mod and successfully implemented multi-turn absolute position control a while back, and used it on many of my projects over the years, you can look at how I modified the code:

https://github.com/supercrazysam/bldc_speed_custom
https://github.com/supercrazysam/bldc_speed_custom/commit/23d7f27bd9b71d604eef9c947361f699beff2060

[...]

Issue #205, opened 2020-08-20
Position contol in ppm mode


Teslafly:
Couple Battlebots friends are asking for position control based on a ppm input. 

Going to take a stab at implementing this.

ppm mapping would map pulse width to rotation angle.

josephduchesne:
I'm also interested 👍

PR #464, opened 2022-04-07
Draft: Add basic ppm position control mode

Teslafly:
Closes https://github.com/vedderb/bldc/issues/205
Probably closes https://github.com/vedderb/bldc/issues/319 too.

Adds a position control mode to the PPM app that enables the vesc to become a ppm/pwm input servo.

This tries to be somewhat intelligent during starts. After safe_start elapses, it will wait until the pwm angle and motor angle are close before switching to position control mode. This should prevent large jerks and big sudden movements when powering on or recovering from a safe_start signal loss event.

Will attach qt xml when PR ready.


datatypes.h
```diff
@@ -550,12 +550,13 @@ typedef enum {
...
 	PPM_CTRL_TYPE_CURRENT_NOREV,
 	PPM_CTRL_TYPE_CURRENT_NOREV_BRAKE,
+	PPM_CTRL_TYPE_CURRENT_BRAKE_REV_HYST,
+	PPM_CTRL_TYPE_CURRENT_SMART_REV,
 	PPM_CTRL_TYPE_DUTY,
 	PPM_CTRL_TYPE_DUTY_NOREV,
 	PPM_CTRL_TYPE_PID,
 	PPM_CTRL_TYPE_PID_NOREV,
-	PPM_CTRL_TYPE_CURRENT_BRAKE_REV_HYST,
-	PPM_CTRL_TYPE_CURRENT_SMART_REV
```
vedderb, line 558: It is best to not move enums around as that will break old configs. Always add new enums to the end. This is why the order of the enums doesn't always make sense.

Teslafly: I was thinking since each config data structure has a unique hash and all settings are erased on firmware upload, that that wouldn't be a problem. But I guess you store configs by enum value instead of by name on the vesc tool side.
```
<app_ppm_conf.ctrl_type>0</app_ppm_conf.ctrl_type>
```

vedderb: For the serialization it does, but not when you load a config stored as xml. Then every enum is stored by value.

Teslafly: Reverted the enum reordering.

applications/app_ppm.c
```diff
@@ -333,6 +338,32 @@ static THD_FUNCTION(ppm_thread, arg) {
...
 
+		case PPM_CTRL_TYPE_PID_POSITION: // make this an actual type.
+			if (fabsf(servo_val) < 0.001) {
+				pulses_without_power++;
+			}
+
+			float angle = servo_val * 360;
+
+			if (!(pulses_without_power < MIN_PULSES_WITHOUT_POWER && config.safe_start)) {
+				// try to more intelligently safe start by waiting until
```
vedderb, line 349: I think safe start makes very little sense in position control mode. It will be confusing if it does not work until you give just the right input and then it suddenly starts working. It is also likely to cause more harm than good if it is used for e.g. a steering servo. Any glitch or fault will make the steering not work until you give it just the right range of input to start working again. I think it is best to just skip the safe start entirely for position control.

Teslafly: So you can just turn safe start off if you don't want it. But most applications I think you want to disable position control if the signal is lost. (battlebots seems the most interested in this for big swing-y or lifter weapons.) You can just disable safe start on fault if that is an issue for you for something like a steering servo.

I want to make sure that you don't get a fast, potentially violent rotation immediately after power up as it tries to go to the right position. 
I want your hands on the controls where you can try to do something/away from the motor. 
Wiggling the sticks after power up when you want things to move is a fairly natural thing to do. And as you pass through the position it will "latch on".

I can make it configurable if you really think that is needed though.

Teslafly: As of right now, to start position control, you need to:
1. Return the control stick to the "resting" position (center ppm for 180 mode, minimum ppm for 360 mode) - your remote should rest in this position.
2. Sweep the control until you hit the position of the motor and it starts tracking.

I think this provides the most amount of safety and convenience, as long as you know what you need to do.
It is possible to disable the first "safe start"  part of this, but not the position matching.

kubark42: > I want to make sure that you don't get a fast, potentially violent rotation immediately after power up as it tries to go to the right position.

100% agreed with the need for this.

applications/app_ppm.c
```diff
@@ -333,6 +333,34 @@ static THD_FUNCTION(ppm_thread, arg) {
 			}
 			break;
 
+		case PPM_CTRL_TYPE_PID_POSITION_180: // -180 to 180. center ppm safestart
+		case PPM_CTRL_TYPE_PID_POSITION_360: // 0 to +360. minimum ppm safestart
+			if (fabsf(servo_val) < 0.02) {
```
kubark42, line 338: Could you drop some comments about what is intended here? Why are you incrementing `pulses_without_power` and why is `0.02` the right value, as opposed to `0.15` or `0.25`?

Teslafly: The incrementing pules without power is on every mode. it's part of the safe start feature.
The 0.02 would correspond to 2% of full scale movement. I was having trouble with the default 0.1% range (probably because i'm using a servo tester, not a remote) and I have a "second layer" of safe start, so I increased it to 2%.

Teslafly: @kubark42 does this answer your question? This section in the other modes is equally as commented.

Teslafly:
Tonight I tried to put together a coherent help section for vesc tool on this.

image

Since github doesn't allow attaching xml files, xml here:
https://github.com/Teslafly/vesc_tool/blob/add_ppm_position_control_settings_help/res/config/6.00/parameters_appconf.xml

Any suggestions here? Too verbose?

I think I like the logic where it is at for now. Any other thoughts?

applications/app_ppm.c
Teslafly: @vedderb 
What are your thoughts on the safe start as it stands now? especially with the proposed vesc tool help section? Do you think I need to add a setting to disable the position based safe start?

Teslafly: Also, if you didn't move the motor from when you powered it off the last time and it was centered. The regular safe start and position safe start windows line up and the motor starts position control right away.

applications/app_ppm.c
kubark42: It does, although I had meant more for the comments to be in the code.

Teslafly:
Is there anything else that must be done to get this merged? I think it is in a good state.

PR #467, opened 2022-04-10
FOC minor optimizations

kubark42:
1. Several precalculations
2. Replace a division by a multiplication
3. Fix some `;;` typos
4. Abstract out a magic number
5. Eliminate some instances of `volatile` keyword.


mcpwm_foc.c
```diff
@@ -2541,8 +2535,10 @@ void mcpwm_foc_adc_int_handler(void *p, uint32_t flags) {
 
 	UTILS_LP_FAST(motor_now->m_motor_state.v_bus, GET_INPUT_VOLTAGE(), 0.1);
 
-	volatile float enc_ang = 0;
```
kubark42, line 2544: @vedderb I can't see why this `volatile` would have been necessary. The variables are used immediately after in the function, so there's not opportunity for things to be updated elsewhere. The functions themselves shouldn't be optimized away, not unless the underlying variables they're using should somehow themselves be volatile.

foc_math.c
```diff
@@ -613,6 +613,9 @@ void foc_precalc_values(motor_all_state_t *motor) {
 	const mc_configuration *conf_now = motor->m_conf;
 	motor->p_lq = conf_now->foc_motor_l + conf_now->foc_motor_ld_lq_diff * 0.5;
 	motor->p_ld = conf_now->foc_motor_l - conf_now->foc_motor_ld_lq_diff * 0.5;
-	motor->p_inv_ld_lq = (1.0 / motor->p_lq - 1.0 / motor->p_ld);
-	motor->p_v2_v3_inv_avg_half = (0.5 / motor->p_lq + 0.5 / motor->p_ld) * 0.9; // With the 0.9 we undo the adjustment from the detection
+	motor->p_inv_inv_ld_lq = 1.0 / (1.0 / motor->p_lq - 1.0 / motor->p_ld);
+
+	motor->p_inv_lo_current_max = 1.0 / conf_now->lo_current_max;
```
vedderb, line 618: The lo-limits are special as they are updated in mc_interface when the different limits are hit, so they cannot be precalculated here. There should probably be some comment explaining that in the struct. This works because mcpwm_foc does not have it's own instance of the configuration - it is mc_interface that owns it and mcpwm_foc only has a pointer to it. The mcpwm_foc_set_configuration-function is only used to tell mcpwm_foc that the configuration has changed so that some checks and updates can be done on it if needed. This function is not called when the lo_limits are updated.

It might be worth rethinking this part at some point if there is a simpler/better solution.

kubark42: It might be worth moving things which are not explicitly user configuration out of the `conf` object and into a separate struct. The memory could be allocated in a similar way, so that there is only one instance shared globally.

I see how this variable is set in `update_override_limits()`. I think it would make sense to have these variables in a globally accessible struct which is written through setters. Modules could subscribe to update notifications, so they would recalculate only when the value has changed.

mcpwm_foc.c
```diff
@@ -3705,14 +3698,18 @@ static void control_current(motor_all_state_t *motor, float dt) {
 	UTILS_LP_FAST(state_m->id_filter, state_m->id, conf_now->foc_current_filter_const);
 	UTILS_LP_FAST(state_m->iq_filter, state_m->iq, conf_now->foc_current_filter_const);
 
+	// TODO: Document this section
 	float d_gain_scale = 1.0;
 	if (conf_now->foc_d_gain_scale_start < 0.99) {
-		float max_mod_norm = fabsf(state_m->duty_now / max_duty);
```
vedderb, line 3710: The name max_mod_norm makes a lot of sense here. After the changes it is much more difficult to see what this section does.

kubark42: Can we fix that with comments? TBH, `max_mod_norm` didn't speak to me, and even after working with the math it's not clear to me what this section is doing. The math is clear, but the why of it all is obscure without a deeper understanding of the code. For instance, `d_gain_scale` is set and modified up to three times.

It's also not clear if this section runs once every never, or it runs almost every loop. Answering those questions might help make it clear if it's worth saving a division or not.

PR #475, opened 2022-04-18
Watchdog reset recovery by switching to UART app

surfdado:
To prevent bootloops, we automatically switch to the UART app to when we're booting from a WDT reset. The first commit is the main one, keeping the switch to UART app temporary, i.e. a hard restart or power cycle will load the original settings. The second commit makes the change to UART permanent. I've left it as a separate optional commit as I'm not sure if there might be use cases where an automatic permanent switch to the UART app is not desirable.

### DIFF
--- a/main.c
+++ b/main.c
@@ -265,6 +265,14 @@ int main(void) {
 	app_uartcomm_initialize();
 	app_configuration *appconf = mempools_alloc_appconf();
 	conf_general_read_app_configuration(appconf);
+
+	// check whether we're rebooting from a watchdog reset
+	if (timeout_had_IWDG_reset(false)) {
+		// switch to uart app when recovering from an IWDG reset
+		// (but don't write it back to allow power cycling)
+		appconf->app_to_use = APP_UART;
+	}
+
 	app_set_configuration(appconf);
 	app_uartcomm_start(UART_PORT_BUILTIN);
 	app_uartcomm_start(UART_PORT_EXTRA_HEADER);
--- a/main.c
+++ b/main.c
@@ -269,8 +269,9 @@ int main(void) {
 	// check whether we're rebooting from a watchdog reset
 	if (timeout_had_IWDG_reset(false)) {
 		// switch to uart app when recovering from an IWDG reset
-		// (but don't write it back to allow power cycling)
 		appconf->app_to_use = APP_UART;
+		// store the changed app permanently, forcing the user to re-selct the desired app
+		conf_general_store_app_configuration(appconf);
 	}
 
 	app_set_configuration(appconf);
--- a/timeout.c
+++ b/timeout.c
@@ -172,12 +172,20 @@ void timeout_configure_IWDT(void) {
 	}
 }
 
-bool timeout_had_IWDG_reset(void) {
+/**
+ * Check if we've rebooted after a watchdog timer reset
+ *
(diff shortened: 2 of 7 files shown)


surfdado:
In case the appeal of this feature isn't clear: people with sealed up boards are pretty screwed if they try a firmware that causes a boot loop due to WDT. Now they will be able to load different firmware without messing with their build. No need for ST-Link

timeout.c
```diff
@@ -172,12 +172,20 @@ void timeout_configure_IWDT(void) {
 	}
 }
 
-bool timeout_had_IWDG_reset(void) {
+/**
+ * Check if we've rebooted after a watchdog timer reset
+ *
+ * @param preserve_flag   set to true to allow subsequent check
```
kubark42, line 178: I like the school of thought on bools that they should answer a true/false question. So `should_preserve_flag` would be my preference.

main.c
```diff
@@ -265,6 +265,15 @@ int main(void) {
 	app_uartcomm_initialize();
 	app_configuration *appconf = mempools_alloc_appconf();
 	conf_general_read_app_configuration(appconf);
+
+	// check whether we're rebooting from a watchdog reset
+	if (timeout_had_IWDG_reset(false)) {
```
kubark42, line 270: I can see this working out poorly if someone expects the system to come back online after a reset, only to find that their app isn't being used. This makes fault tolerance somewhat trickier.

On the flip side, you're dealing with the very real problem of people who cannot access their SWD port and find themselves locked out of their board. Would a happy compromise be to add some more conditions before setting the app to UART? Perhaps checking the number of boot loops vs. how long it was since the last reset (under the assumption that if there is a 60 second window when someone could upload new firmware, that's sufficient to reset the boot loop counter)

surfdado: is there a way to count boot loops and measure time since reset? In my experience the boot loop has a period of just a few seconds, i.e. a subsequent reboot gets triggered right after startup. Are you suggesting to add a WDT reboot counter to persistent memory and then clear it once we've been up for more than x seconds?

In my experience loading firmware requires a lot more time than 60 seconds (bluetooth), but at the same time the reboots I've seen usually happen after less than 10 seconds, so having enough time for a firmware download isn't really important.

timeout.c
surfdado: "should" doesn't instill much confidence imo, maybe "do" would be better, but I'm fine with whatever name people prefer

main.c
Teslafly: Is the app thread not terminated when you start a firmware upload? Terminating the app and mcpwm thread on starting a fw upload seems like that may be a good idea. After ensuring the motor is at 0 current.

surfdado: Pretty sure the app isn't terminated. I remember the early version of 5.3 had my motor wiggling (some weird brake current behavior) while firmware was uploading. If I recall correctly I actually submitted a patch to deal with it but it get rejected because Mitch added a different workaround/timeout for the brake behavior

vedderb:
The problem with this change is that the reset command uses the watchdog to reset, which would disable the app on the next boot. Other things like an undervoltage in the wrong moment or a current spike on hardware with bad layout can also cause watchdog resets. In those cases it is better that the next boot picks the previous app as it can leave you stranded until you have a phone or computer to change the config.

If there is an issue where certain configurations in an app cause boot loops those issues should be solved in the app.

PR #479, opened 2022-04-21
Add cascading PID to the balance app

Mitchlol:


vedderb:
I have been thinking about how to deal with all the parameters, especially after realizing that @surfdado also needs a whole set of new parameters and probably a different app for balance skateboards as they are quite different. All of that will not fit in the appconfig the way it is now and changing the sizes will slowly eat all the ram that is left as each parameter in appconfig needs to have probably at least 10x the space as there are so many interfaces that can send things at the same time. It would also break ble-modules out there now as they also need larger buffers.

One way to approach this would be to only expose the very generic parameters in the appconf that normal users are likely to tweak (which probably includes 90% of all parameters). The rest of the parameters could be configured using a lisp script that runs on each boot (all of them do). All you have to do then is to add a few lisp extensions to the balance app similar to this:
https://github.com/vedderb/bldc/blob/master/lispBM/README.md#configuration
Then when you have a config that works and you want to share it you just share that script as a text file.

This also has a few other advantages:
* You don't need to make changes to VESC Tool when you add or change parameters.
* You won't break compatibility with the firmware version, meaning that you can change the balance app on the stable release with new parameters when you feel like it without impacting the rest.
* You can make it so that you can update the parameters without stopping the motor, which could be interesting for tuning parameters live. One way to do that would be to add some potentiometers to the ADCs that you read from lisp and update the parameters as you go. You can also make a qml-app that interacts with the lisp app with some sliders for tuning.
* If you want to make a balance setup ui in qml it is quite easy to make that interact with the lisp code to change parameters, and as it is completely independent you can update and maintain it without being bound by the firmware releases.

I can start on it for a few parameters if it sounds interesting so that you can see what I mean, and if it looks good you can proceed with the rest of them.

Mitchlol:
I'm not really sure I see a clean way to split parameters. Would it make sense to move all the parameters? What about rewriting the balance app as a lisp script? Can they be apps or are they just configuration scripts? I don't know anything about this stuff 🤯

I don't think I plan on adding a ton of more parameters. In terms of features it's pretty good, really just trying to get performance as good as I can. If we split onewheeled skateboards and unicycles into 2 apps, I can remove a bunch of features from this one. Tho I'm not so sure that it needs to be 2 different apps.

PR #482, opened 2022-04-23
Pass reply_func to sample_send_thread

TechAUmNu:
When data is being sampled another command can be received on a different port and cause the sampled data to be sent to the wrong port. This change saves the reply_func for the requester so other commands can be processed on other ports without issue.


vedderb:
Nice fix!

PR #486, opened 2022-05-02
Add unity parallel output config

Teslafly:
Unity never got a parallel hw config .h like stormcore and someone on the discord asked for it.

### DIFF
--- a/hwconf/hw_unity_parallel.h
+++ b/hwconf/hw_unity_parallel.h
@@ -1,0 +1,25 @@
+/*
+	Copyright 2016 Benjamin Vedder	benjamin@vedder.se
+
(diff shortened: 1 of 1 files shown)


Teslafly:
Compilation is not currently working. wants a hw_unity_parallel.c file but can't figure out why.

Teslafly:
got some help from Aeroboticist. needs a "core" fw file structure.

vedderb:
It would be good if someone can try this out before merging it. Would be bad to have a config in the repo that makes the unity catch fire. Dual parallel is quite risky in general as both power stages have to be in perfect sync for it to not short out.

Teslafly:
Am less motivated to try and get this working. closing for now.

PR #488, opened 2022-05-04
Improve fault handling during detection routines 

TechAUmNu:
Added more checks for faults during detection routines. 
Always return an error if a fault occurred. 
Handing for returned errors so later steps do not get run.


TechAUmNu:
I have tested this with a current limited supply so it causes an under_voltage fault when running the detection routines. 
Previously some of the routines would continue even with a fault being active.

PR #490, opened 2022-05-05
Set abs max to reasonable value during detection

TechAUmNu:
During the FOC wizard the motor config can be reset to default. 
The motor max current can be set by the wizard to be higher than the default abs max current.

This change sets the abs max current to the motor config default before detection, 
then to 1.5x the detected max motor current after detection.


surfdado:
Yes, please! This will help prevent so many nosedive onewheel crashes going forward... many users have hit abs-max because they set it to e.g. 120A or less with 100A motor current

Teslafly:
Is this feature a candidate for the 6.0 firmware release? or should it be postponed?

PR #495, opened 2022-05-13
Log faults if active on startup

avlasic:
Currently if fault is active on startup it won't be logged to terminal so it is' hard to notice faults like booting from watchdog reset.
Also current offset and unbalace faults shoudn't be triggered before dccal , since after that vairables are filled with real data from adc.

### DIFF
--- a/mc_interface.c
+++ b/mc_interface.c
@@ -239,6 +239,12 @@ void mc_interface_init(void) {
 		break;
 	}
 
+	//Clear fault, to log first fault to terminal
+	m_motor_1.m_fault_now = FAULT_CODE_NONE;
+#ifdef HW_HAS_DUAL_MOTORS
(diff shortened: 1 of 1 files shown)


vedderb:
Can you see if this commit gives the same result?
https://github.com/vedderb/bldc/commit/1309ece84be5f69968f7fc149d0f9e241613e082
Then you don't have to check dccal_done in the isr-code. Hopefully the comment also explains why the faults behave the way they do.

avlasic:
That commit does give me same result, logs both first fault (like watchdog) that is gone or still active as it should.

As for dccal_done on current offsets, that check is done in 1ms task, not isr code so I think it should be fine to do it, since I still get FAULT_CODE_HIGH_OFFSET_CURRENT_SENSOR with run calibration on boot to off.

vedderb:
That sounds good. Is it ok to close this one then?

PR #510, opened 2022-07-20
Set lsm6ds3 odr nearest to configured frequency

Mitchlol:


surfdado:
How do we tell the user which frequencies are "recommended" based on the IMU they have? BMI160 likes 800Hz while LSM6 runs best @833Hz? I'm still not sure if having the frequency balance/IMU frequency match the native frequency of the IMU really does make a difference or if it's placebo, but it sounds like it should be real?

Mitchlol:
If we want to inform users right now, the only way to do it that i see would be in the help text of the vesc tool.

However i think the more long term proper way would be to have each IMU use its own lisp plugin, and have its own QML GUI where it can have a completely custom configuration.

As far as matching the frequencies, something that is very real and significant, is that for both the lsm6ds3 and bmi160 the imu data has less noise at lower frequencies. If you look at the raw gyro x/y/z you can see the waveform get way smaller when selecting lower frequencies, and this does translate to a significant performance increase.

In terms of synchronization, the imu and balance app threads are not explicitly synchronized, so there is probably always gonna be some small percentage of duplicate and skipped reads. But when adjusting at 1khz, if one adjustment is wrong, it hardly matters. if you select 800hz balance loop on an 833hz imu, its probably fine, but if you select 833hz balance loop on an 800hz gyro then you'll get ~4% duplicate reads which is probably still fine. But I think if you mess with it too much, you'll probably see some performance degradation. There is no harm in matching them up, but i agree it is complex for a non-poweruser to configure.

surfdado:
I guess this is something a future balance-setup-wizard would take care of.

Mitchlol:
That would work, i guess i should actually get around to writing that. Got a few directionality related bugs to crush first, and i wanna do the I-term limiting. I also wanna dig a bit more into the lsm6ds3 and add support for the filters, but the documentation is terrible, and the defined constants in the lsm6ds3.h don't match the datasheet.

PR #511, opened 2022-07-20
Add support for the AS5x47U encoder

ZachOB:
This adds support for the SPI based [AS5147U](https://ams.com/en/as5147u) and [AS5247U](https://ams.com/en/as5247u) encoders. These are basically the same encoder, with the AS5247U just consisting of two AS5147U's stacked on top of each other in the same package.

This encoder is similar to the already supported AS5047 encoder, but includes extra hardware error detection and an 8 bit CRC instead of a parity bit, allowing it to reach automotive ISO26262 safety levels. I have also added an additional fault to better report internal errors reported by the encoder.

The encoder is connected to hardware SPI on the COMM port, requiring the UART app to be disabled and preventing use of ADC2, ADC3, and I2C. The SPI communication is implemented via DMA, resulting in extremely low CPU load. I've also tested to ensure the SPI and DMA are able to recover from all types of errors (disconnection, short to ground, and short to vcc of all 4 wires).

PR to update the VESC tool is [here](https://github.com/vedderb/vesc_tool/pull/237).


encoder/enc_mt6816.c
```diff
@@ -42,10 +42,14 @@ bool enc_mt6816_init(MT6816_config_t *cfg) {
...
-	palSetPadMode(cfg->nss_gpio, cfg->nss_pin, PAL_MODE_OUTPUT_PUSHPULL | PAL_STM32_OSPEED_HIGHEST);
-	palSetPadMode(cfg->mosi_gpio, cfg->mosi_pin, PAL_MODE_ALTERNATE(6) | PAL_STM32_OSPEED_HIGHEST);
+	palSetPadMode(cfg->sck_gpio, cfg->sck_pin,
+			PAL_MODE_ALTERNATE(cfg->spi_af) | PAL_STM32_OSPEED_HIGHEST);
+	palSetPadMode(cfg->miso_gpio, cfg->miso_pin,
+			PAL_MODE_ALTERNATE(cfg->spi_af) | PAL_STM32_OSPEED_HIGHEST);
+	palSetPadMode(cfg->nss_gpio, cfg->nss_pin,
+			PAL_MODE_OUTPUT_PUSHPULL | PAL_STM32_OSPEED_HIGHEST);
+	palSetPadMode(cfg->mosi_gpio, cfg->mosi_pin,
+			PAL_MODE_ALTERNATE(cfg->spi_af) | PAL_STM32_OSPEED_HIGHEST);
```
ZachOB, line 52: I started with this file as a base, and discovered that the alternate pin mode was incorrect (at least for the 75_300). I modified the hardware SPI config to include a pin alternate that is pulled from the hw_config

vedderb:
Looks nice, thanks for the PR!

PR #517, opened 2022-08-29
solve linking error when using gcc version 10

benzjimny:
When using gcc version 10 to compile the project, an error which states undefined symbol will occur during linking stage. The root cause is the inline functions. Compiling the project with gcc version 7 and 10 has been done with command **make all_fw**. The result is ok. There are some warnings with gcc version 10 and I'm going to fix them in the next commit.


kalvdans:
LGTM!

PR #533, opened 2022-09-27
FOC fault handling, terminal usability and flux linkage detection improvement.

TechAUmNu:
Add fault handling to all foc functions that touch the motor. Fault codes are now caught at the time they are generated and returned. Previously a lot of commands could be run even when a fault was active.
The conf_general_detect_apply_all_foc returns faults as -100 + fault for display in the foc wizard.  

Improve terminal usability by echoing commands and giving feedback on faults and missing / invalid arguments. 

Improve flux linkage measurement  and using the measured resistance and inductance to set the kp & ki.
I also found that increasing the switching frequency helped although not sure exactly if 40khz zvf is allowed on the dual hardwares.


vedderb:
Nice updates!

I will be away for a few days, then I will start reviewing and testing this.

About dual hardware, it is best to stay at 25 kHz to be safe.

One thing that would be great is if you fix the code indentation. Your updates have spaces for indentation, but all other code has tabs (I'm using a tab width of 4 btw). Most editors can probably choose to use tabs or spaces with some setting, and if you set it to tabs you can select the code you changed and use the auto-indent function to fix it (ctrl+i in eclipse and qt creator). In case you are wondering: vesc tool uses spaces and the firmware uses tabs. Don't ask me why :)

TechAUmNu:
Thanks, I will put in the vesc tool side tomorrow.

Ok will make it 25 kHz on the dual hardware.

I didn't notice the indentation was using spaces in qt until I committed! Will change it to tabs.

motor/mcpwm_foc.c
```diff
@@ -692,15 +692,28 @@ void mcpwm_foc_stop_pwm(bool is_second_motor) {
  *
  * @param dutyCycle
  * The duty cycle to use.
+ *
+ * @return
+ * The fault code
  */
-void mcpwm_foc_set_duty(float dutyCycle) {
+int mcpwm_foc_set_duty(float dutyCycle) {
```
vedderb, line 699: I think it is best if these don't return the fault code. The check is only done in the beginning anyway and if you are interested in that check you can just do it yourself. If the command itself can fail from the processing it does makes sense to return that result, but having it check some external state and returning that independently of what the command did just makes things messy and hard to follow.

Another reason for not adding extra function calls to these is that they might be called at a very high rate from threads that run control loops and we don't want to cause too much performance impact. If does not matter much now, but these small things add up over time in the code base.

Another thing: These commands bypass mc_interface, so you have to use them with a bit more responsibility in general. mc_interface usually does these high level checks for faults, timeout etc. and uses these functions accordingly.

TechAUmNu: That is a fair point, I just checked and all these ones are only used from mc_interface, so I will remove the checks in here as they are duplicated.
```
void mcpwm_foc_set_duty(float dutyCycle);
void mcpwm_foc_set_duty_noramp(float dutyCycle);
void mcpwm_foc_set_pid_speed(float rpm);
void mcpwm_foc_set_pid_pos(float pos);
void mcpwm_foc_set_current(float current);
void mcpwm_foc_set_brake_current(float current);
void mcpwm_foc_set_handbrake(float current);
```

The openloop functions don't have a corresponding mc_interface command so we can either add one with the same checks as the other commands to keep things consistent in behavior. Or keep the checks in these functions. 
```
int mcpwm_foc_set_openloop(float current, float rpm);
int mcpwm_foc_set_openloop_phase(float current, float phase);
int mcpwm_foc_set_openloop_duty(float dutyCycle, float rpm);
int mcpwm_foc_set_openloop_duty_phase(float dutyCycle, float phase);
```

vedderb: I think it is best to add these to mc_interface and use mc_interface_try_input in the same way as with the other functions. Initially I treated these as specialized functions for my testing, but if they are generally useful it is best to run them via mc_interface with the same checks as the other functions.

TechAUmNu:
I think this is now ready to go.

Associated change to display faults in the wizard https://github.com/vedderb/vesc_tool/pull/259

PR #542, opened 2022-10-11
Improve Sin/Cos Encoder, fix lowpass filter, fix scaling

marhmm-ducktrain:
Current Sin/Cos Encoder implementation has two issues:

1. `UTILS_LP_FAST(angle, angle_tmp, cfg->filter_constant);` : angle variable is always zero in this context. Therefore the returned angle is always cfg->filter_constant * angle_tmp (with default filter constant of 0.5). This also leads to a wrong Encoder Ratio which is twice as high as desired.
2. `UTILS_LP_FAST` in general is not able to filter wrapping numbers. For example, if last value is 350° and new value is 20°, the filter returns 185°, which is not the desired value of 5°. I added a UTILS_LP_FAST_360 which is capable of filtering wrapping values with 0 <= value < 360

### DIFF
--- a/encoder/enc_sincos.c
+++ b/encoder/enc_sincos.c
@@ -72,8 +72,9 @@ float enc_sincos_read_deg(ENCSINCOS_config_t *cfg) {
 			UTILS_LP_FAST(cfg->state.signal_low_error_rate, 0.0, timestep);
 
 			float angle_tmp = RAD2DEG_f(utils_fast_atan2(sin, cos));
-			UTILS_LP_FAST(angle, angle_tmp, cfg->filter_constant);
-			cfg->state.last_enc_angle = angle;
+			utils_norm_angle(&angle_tmp);
+			UTILS_LP_FAST_360(cfg->state.last_enc_angle, angle_tmp, cfg->filter_constant);
+			angle = cfg->state.last_enc_angle;
 		}
 	}
 
--- a/util/utils_math.h
+++ b/util/utils_math.h
@@ -98,6 +98,24 @@ void utils_rotate_vector3(float *input, float *rotation, float *output, bool rev
 #define UTILS_LP_FAST(value, sample, filter_constant)	(value -= (filter_constant) * ((value) - (sample)))
 
 /**
+ * A simple low pass filter able to filter wrapping values between 0 and 360
+ *
+ * @param value
+ * The filtered value.
+ *
+ * @param sample
+ * Next sample.
+ *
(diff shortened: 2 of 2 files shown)


vedderb:
This looks like an issue for sure, I haven't looked at this code much myself.

Isn't it better to run the regular low pass filter on sin and cos, then do atan2 with the filtered values?

vedderb:
What do you think about this?
https://github.com/vedderb/bldc/commit/d361823e853912df4b023687a522d963b8e45dad

vedderb:
I agree, here is another update:
https://github.com/vedderb/bldc/commit/56c5648770fd61154ede467208cf5c5023ec50f5

nitrousnrg:
> If we change some parts of this code, we maybe also could think about the following line: float sin = ENCODER_SIN_VOLTS * cfg->s_gain - cfg->s_offset;
> 
> In my opinion it should be: float sin = (ENCODER_SIN_VOLTS - cfg->s_offset ) * cfg->s_gain ;

We have been working on this bit code for the last couple of weeks, and I fully agree with this, I made that change myself, with the following addition:

Instead of a **gain** parameter, the code should get an **amplitude** parameter. This way, what we are doing is normalizing the sine to an amplitude of 1.0

The benefit of this is that the following SINCOS_MAX_AMPLITUDE and SINCOS_MIN_AMPLITUDE checks can be safely hardcoded and don't depend on the actual signal amplitude.

image

The module is key here, ideally it would be a flat line at 1.0.. but an offset error as small as 70mV produce +/-5° of position error (green trace). The best way is to plot the module (pink trace) and tune the parameters until its as flat as possible. 
image

The goal here is to tune the pink trace so stays within 0.9v and 1.1v, which represents as much error as you can handle.

This morning we've been testing plotting this module within vesc tool instead of using a DAC to avoid the need of a dac and scope and worked pretty well.

I never noticed the bug in the filter, what we did last week is move the filtering to the sin/cos analog signals, as in general its better a light filter at the input of a signal chain.

I wanted to have a motor running at full power before attempting a PR, as I'm concerned if this needs stronger hardware filtering that could require phase lag compensation in the code :/

vedderb:
The latest version has an amplitude-parameter instead of the gain parameters. The min and max errors are fixed at the moment, but they will be used after the amplitude is applied so they should be relative at least. I could also add a parameter to make the min and max error configurable.

I also tried a sin/cos-encoder the other day and it works really well for me.

PR #544, opened 2022-10-16
improve the logic for searching toolchain and python

benzjimny:
To keep the toolchain in the project folder is not a good practice. I prefer to use the system wide toolchain. And the default toolchain is a little old. Maybe we can create a wiki on where to download the latest toolchain and how to install it system wide. This commit will search the toolchain in system environment first, if it find nothing, it will search in the project or it will give an error to prompt execute make arm_sdk_install.


Teslafly:
The reason the tool chain is locked down is beacuse of this issue.
https://github.com/vedderb/bldc/issues/210

Using newer versions of the tool chain could cause different behavior.  Though I vaguely remember a pr that fixed this a little while back? 

That said, this is basically a safety critical application, and while it can't follow some guidelines for such a project, locking down the compiler version prevents many issues that are hard to find and debug. It keeps things as deterministic as possible. As much as one can in C. 
So even if the compiler bug is fixed, the compiler should still stay locked down, but possibly just updated.

Teslafly:
Is there any value in this PR still if the compiler version continues to be locked down?

vedderb:
I think we can close this one for now.

PR #548, opened 2022-11-06
Update gcc/gdb version

Teslafly:
Some tools have depreciated the older version of gdb we use and the issue causing https://github.com/vedderb/bldc/issues/210 on newer versions of gcc have been fixed. So it may be worth upgrading toolchains now.

Tested with vesc 6.0 mk4  in windows. Detecting and running a motor works.
Also successfully tested compilation in wsl linux. Mac not tested.

One things with the newer gcc version is it throws a lot more warnings about casts and implicit conversions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ workspace.code-workspace
 *.pro.shared
 
 .vscode
+*.log
+null.d
+nul.d
--- a/Makefile
+++ b/Makefile
@@ -73,24 +73,24 @@ endif
 ##############################
 .PHONY: help
 help:
-	@echo
+	@echo ""
 	@echo "   This Makefile is known to work on Linux and Mac in a standard shell environment."
-	@echo
+	@echo ""
 	@echo "   Here is a summary of the available targets:"
-	@echo
+	@echo ""
 	@echo "   [Tool Installers]"
 	@echo "     arm_sdk_install      - Install the GNU ARM gcc toolchain"
 	@echo "     qt_install           - Install the all tools for Qt"
-	@echo
+	@echo ""
 	@echo "   [Big Hammer]"
 	@echo "     all_fw               - Build firmware for all boards"
 	@echo "     all_fw_package       - Packaage firmware for boards in package list"
-	@echo
+	@echo ""
 	@echo "   [Unit Tests]"
 	@echo "     all_ut               - Build all unit tests"
 	@echo "     all_ut_xml           - Run all unit tests and capture all XML output to files"
 	@echo "     all_ut_run           - Run all unit tests and dump XML output to console"
-	@echo
+	@echo ""
 	@echo "   [Firmware]"
 	@echo "     fw   - Build firmware for default target"
 	@echo "                            supported boards are: $(ALL_BOARD_NAMES)"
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -16,3 +16,6 @@ workspace.code-workspace
 *.pro.shared
 
 .vscode
+*.log
+null.d
```
kalvdans, line 20: What is null.d?

vedderb:
I gave it a try now on the same hw. As the last time I tried the code got larger and slightly slower. For now I would prefer to stay with the old compiler for that reason. It would be great to know what is going on.

**Compiler  : 7.3.1**
```
   text	   data	    bss	    dec	    hex
 428988	   4300	 173812	 607100	  9437c
```

LBM speed test 3x:
("Seconds elapsed: " {5.257800})
("Seconds elapsed: " {5.276200})
("Seconds elapsed: " {5.272000})

ADC interrupt test 3x with motor running:
Latest injected ADC duration: 0.0239 ms
Latest injected ADC duration: 0.0239 ms
Latest injected ADC duration: 0.0243 ms

---

**Compiler  : 10.3.1**
```
   text	   data	    bss	    dec	    hex
 445040	   4304	 173788	 623132	  9821c
```

LBM speed test 3x:
("Seconds elapsed: " {5.361800})
("Seconds elapsed: " {5.375900})
("Seconds elapsed: " {5.373300})

ADC interrupt test 3x with motor running:
Latest injected ADC duration: 0.0239 ms
Latest injected ADC duration: 0.0239 ms
Latest injected ADC duration: 0.0251 ms

vedderb:
It would be good to check how the binary size and performance changes with later versions of gcc. Last time I tried the binary would get quite a bit larger and the code got slightly slower.

.gitignore
Teslafly: it's generated by the makefile extension.

kalvdans: I can't find any reference to null.d on the help page. It is strange that they choose NUL.D is a reserved filename on Windows. https://answers.microsoft.com/en-us/windows/forum/all/unable-to-rename-a-folder-to-com-errorthe/ff48c793-ab97-4881-952a-df195321c11c

Teslafly: yeah. i haven't been able to figure it out. i just know it gets generated when they generate the list of make options. and it's not consistent either.

kalvdans: Remove from PR then, and complain to microsoft :)

Teslafly:
looking at the notes on releases since the gcc-arm-none-eabi-7-2018-q2-update release, 

There is this which could be increasing the memory size.
Thumb1 code size regression due to new register allocation:
    https://gcc.gnu.org/bugzilla/show_bug.cgi?id=59535
which may be fixed now in the 10.3-2021.10 version?
    
Otherwise the other notes seem nonrelevant. And it's mainly changes to mainline gcc that are getting pulled in. Which would be exploit vulnerability fixes, memory safety improvements, bux fixes and tool improvements.

Teslafly:
Not a candidate for the fw6.0 release.
But keeping this open to potentially include in the future. unless this should really be closed?

PR #561, opened 2022-11-27
Adding passive (shorted) brake to FOC

betzuka:
Added new low level  FOC function ` mcpwm_foc_set_passive_brake() ` that shorts all phases through low side FETs, braking the motor. Intention is to use this for continuing to brake a propeller using zero power after actively stopping it using negative iq or other brake methods. If used at speed the braking effect maybe violent. It attempts to respect ABS_MAX_CURRENT and will raise a fault if exceeded.

The new function is exposed through lisp only at the moment.

### DIFF
--- a/lispBM/lispif_vesc_extensions.c
+++ b/lispBM/lispif_vesc_extensions.c
@@ -1252,6 +1252,13 @@ static lbm_value ext_set_brake_rel(lbm_value *args, lbm_uint argn) {
 	return ENC_SYM_TRUE;
 }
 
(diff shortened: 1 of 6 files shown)


vedderb:
I will wait until after the release before including this one. I think it will be better to be in the running state and not use the current controller as that will allow tracking the RPM based on the measured currents. In the stopped state the currents for the observer are set to 0, so the RPM and position will not be tracked like this.

Teslafly:
Now that 6.02 has been released. should this be looked at again?

Issue #557, opened 2022-11-21
Compiling in Windows (PowerShell) error


Talabo:
Compilation of the master-branch in Windows 11 (Windows PowerShell) seem to break and not work "straight from the shell". With the setup instructions from the masterbranch readme file.

This is the following result I receive:

> PS C:\Users\me\Downloads\bldc> make 100_250
> "********* BUILD: 100_250 **********"
> 
> 
>     Directory: C:\Users\me\Downloads\bldc\build
> 
> 
> Mode                 LastWriteTime         Length Name
> ----                 -------------         ------ ----
> d-----        18.11.2022     11:04                100_250
> 
> 
> make[1]: Entering directory 'C:/Users/me/Downloads/bldc'
> The syntax of the command is incorrect.
> Compiling crt0_v7m.s
> ChibiOS_3.0.5/os/common/ports/ARMCMx/compilers/GCC/crt0_v7m.s:1:0: fatal error: opening dependency file build/100_250/.dep/crt0_v7m.o.d: No such file or directory
>  /*
> 
> compilation terminated.
> make[1]: *** [ChibiOS_3.0.5/os/common/ports/ARMCMx/compilers/GCC/rules.mk:235: C:/Users/me/Downloads/bldc/build/100_250/obj/crt0_v7m.o] Error 1
> make[1]: Leaving directory 'C:/Users/me/Downloads/bldc'
> make: *** [Makefile:212: fw_100_250_vescfw] Error 2
> PS C:\Users\me\Downloads\bldc>

If i run `make all_ut` I get the following:

> process_begin: CreateProcess(NULL, g++ --version, ...) failed.
> Makefile:18: pipe: No error
> Makefile:24: Unknown compiler
> make[1]: *** No rule to make target 'C:/Users/me/Downloads/bldc/utils_math.c', needed by 'C:/Users/me/Downloads/bldc/build/unit_tests/utils_math/utils_math.o'.  Stop.
> make: *** [Makefile:312: ut_utils_math_elf] Error 2

Teslafly:
It's missing the `.dep` folder. If you manually make that as a workaround it should compile fine.

This does need to be fixed though.

Teslafly:
I think that the issue is this line:
`$(shell cmd /C if not exist "$(DEPPATH)" mkdir "$(DEPPATH)")`
it gives the error `The syntax of the command is incorrect.`

[...]

PR #562, opened 2022-11-28
fix issue #557

Jfriesen222:


Talabo:
Hi @Jfriesen222 !

Quick question, what significance does a "forward slash" have instead of "backslash"? (fix issue #557)

PR #566, opened 2022-11-30
Adding Flipsky 75100 Hardware

jaykup26:
Hardware configuration files for Flipsky 75100, 75100v2 and 75100 Aluminum PCB

More details can be found here

https://forum.esk8.news/t/how-to-update-firmware-on-the-flipsky-75100-foc-esc/61819

### DIFF
--- a/hwconf/flipsky/hw_75_100.c
+++ b/hwconf/flipsky/hw_75_100.c
@@ -1,0 +1,242 @@
+/*
+	Copyright 2018 Benjamin Vedder	benjamin@vedder.se
+	This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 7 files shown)


jaykup26:
> Should the v2 version have 2 different versions? The alu pcb version can do a lot more current safely than the TO220 based one.

While I agree that the Aluminum PCB version can probably handle 150A or more, Flipsky rates that controller at 120A burst on their website, so I'd be hesitant to make a version that goes over their specs.  I did include a no limit version for both the V1 and V2 so that could be loaded and any amp setting could be run.  The V2 and Aluminum PCB are otherwise identical from a vesc hardware config standpoint.

I just added a 0.0005 ohm shunt version of the V2 as a non-default firmware.  It also has limits disabled as those who are able to mod a controller are likely able to find safe limits on their own.

PR #569, opened 2022-12-02
[Feature] SIN/COS phase error compensation

totosolat:
Some sin/cos encoders do not provide perfec 90° phase between the two signals.

It is in particular the case of son/cos encoders made of two hall effects sensors placed in front of the rotor magnets. When working with high pole pairs number, very light placement error on the hall effect sensors can induce important electric angle phase error between sin and cos signals.

Perfect encoder would provide 90° phase between the two signals. Assuming we have a rotor angle **α** and a phase error **θ**, the encoder will output:
```math
ADC SIN → sin(α)
```
```math
ADC COS → cos(α+θ) = cos(α)cos(θ) - sin(α)sin(θ)
```
We can correct the output the following way:
```math
CORRECTED COS : cos(θ) = (cos(α+θ) + sin(α)sin(θ))/cos(α)
```

This feature has been tested on a motor with 10 pole pairs and an identified phase error of 8 deg electrical, we can see the inpact on FOC control.
Without phase error compensation:
phase_not_compensated2

With phase error compensation:
phasecompensated

Note: I am planning to provide a lisp example to help identify this phase error **θ** and/or working on automatic SIN/COS calibration to identify all SIN/COS parameters at the same time.


encoder/encoder_datatype.h
```diff
@@ -155,6 +155,9 @@ typedef struct {
 	float s_offset;
 	float c_offset;
 	float filter_constant;
+	float phase_correction;
+	float sph;
+	float cph;
```
kalvdans, line 160: Add a comment for each of these what state they contain.

totosolat: Is it what you expected?
https://github.com/vedderb/bldc/pull/569/commits/fa01773ab6cf8ef1ce10c7d998b24a32be6cdba6

kalvdans: Yes, thank you!

kalvdans: Also write that phase_correction is in degrees.

totosolat: Done
I also reverted some changes not related with this feature: https://github.com/vedderb/bldc/pull/569/commits/164bcf7052173a67eb98ed9285b1e4ddd5f634b0

PR #570, opened 2022-12-03
Fixup for commit b8e8952e - use reasonable defaults for imu filters

surfdado:
The default values should be kp=0.3 and beta=0.1 (just like they used to be in 5.3 and before)


vedderb:
These are not the default values, these are just higher gains that are used in the very beginning to move the filter towards the correct position faster. If you, for example, boot and the IMU is upside down and you have low gains it can take 10 seconds before it turns around to the correct orientation. By having these high gains in the beginning it will start by turning around quickly and then use the gains from the config. I think @Mitchlol wrote this code, maybe he has a comment on this.

surfdado:
you're right, I should have looked at your commit more carefully, sorry about that

PR #573, opened 2022-12-03
Disable field weakening when erpm < sensorless

surfdado:

### DIFF
--- a/motor/foc_math.c
+++ b/motor/foc_math.c
@@ -658,7 +658,9 @@ void foc_run_fw(motor_all_state_t *motor, float dt) {
 		float duty_abs = motor->m_duty_abs_filtered;
 
 		if (motor->m_conf->foc_fw_duty_start < 0.99 &&
-				duty_abs > motor->m_conf->foc_fw_duty_start * motor->m_conf->l_max_duty) {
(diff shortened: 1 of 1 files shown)


vedderb:
Why would you disable field weakening in the sensored region?

Teslafly:
If the sensored cutoff erpm is increased with a higher speed sensor into the region you want field weakening, this could be problematic. and probably needs more justification and nuance.

surfdado:
The reason I've added this change is that I have personally experienced massive torque loss when going up really steep inclines slowly, so this change fixed that problem for me. If I'm going 2-3mph I just don't see the benefit of having field weakening kick in, just wasn't sure if there was some corner case I wasn't considering

But I have to admit that I was approaching this with a pretty narrow mindset, thinking only about our Onewheel usecases and 2000erpm aka 4mph sensorless transition speed. I didn't necessarily expect this to get accepted into mainline, just wanted feedback to make sure I'm not doing something stupid/dangerous

nitrousnrg:
Big no-no for those using a resolver 100% of the time 🙂
Large inverters usually run 100% sensored and use field weakening.

vedderb:
A better way might be to limit the fw current to sqrt(i_max^2 - iq^2). Then the Q-axis gets priority on the current. I don't want to try that before the release though, so that is for the next beta.

PR #575, opened 2022-12-06
PPM/UART Remote Support for VESC Packages

NicoAleman:
Now able to fetch UART or PPM remote input (after deadband + throttle curve) from VESC Package through vesc_c_if. [Adding for support for a Float Package feature.](https://github.com/vedderb/vesc_pkg/pull/5)

Be sure to check over the logic included for app_ppm.c when config.ctrl_type == PPM_CTRL_TYPE_NONE, as it forces the next iteration of the main loop once the throttle input is gathered. Judging by app_nunchuk.c, which already does this, I assume it is fine, but there may be some application for PPM_CTRL_TYPE_NONE that requires continuing through the loop that I am not aware of. If so, just let me know and I can generalize it to simply store the value of ppm_servo_val once it's calculated, and save the extra needed logic for custom firmware.


vedderb:
I think it is better to do the deadband and throttle curve in the float app and add settings for them there. Otherwise users have to go and configure some other app which affects the float app.

vedderb:
I have added access to the remote state and ppm functions:
https://github.com/vedderb/bldc/commit/6482fcda3997c5639908cfb34b3777652de9893d
get_ppm also checks if the decoder is running and starts it if it isn't, so you don't even need to run the ppm-app. I have also added the utility function from the firmware so that you can include utils.h and get access to the deadband and throttle curve functions. Do you think that is ok?

NicoAleman:
I think this should be great, at least from just looking at it, thanks! The only potential concern I have is with PPM, as in my testing, I had to specifically report back a servo_val of 0 if the timeout threshold had been met, as it would otherwise often read a non-zero reading until the remote was powered on. But that's fairly specific and may not be an issue with this implementation, especially with the servodec_is_running() check. I'll give it a shot when I get a chance 👍

NicoAleman:
Testing rn and it seems solid for the most part, just have one issue. The current implementation does not map the PPM servo val around the configured pulse_center, just pulse_start and pulse_end, leaving the possibility (and likelihood) of it centering on a non-zero value at idle. 

[I have proposed a PR modifying the get_ppm() function](https://github.com/vedderb/bldc/pull/581#issue-1481309045), though if this behavior may be desired in some scenarios, perhaps this can be moved into it's own function. Should be a simple implementation either way, it seems to have done the trick for me and works great now! Thanks again for taking the time for this.

PR #586, opened 2022-12-27
Markdown format and release dates for the changelog

jaykup26:
This modifies the CHANGELOG to:
 - allow for issues, pull requests and commits to be automatically linked 
 - include the release dates
 - make it a little more readable 

This file could be renamed to CHANGELOG.md but another option to keep the filename the same is to simply add `<!-- vim: syntax=Markdown -->` at the top.  This enables markdown while keeping the filename the same.

An example can be seen here:

image

https://github.com/jaykup26/bldc/blob/changelog/CHANGELOG

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
-=== FW 6.00 ===
+<!-- vim: syntax=Markdown -->
+### 6.00
+#### Released 2022-12-08
 * Added stack checks.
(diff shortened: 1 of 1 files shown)


vedderb:
Looks good! Qt 5.15 also supports markdown, so after a few updates they also look good in VESC Tool:
image

PR #588, opened 2023-01-04
Adds vesc_c_if nvm access

Blevruz:
Adds a way for packages to edit sector 8 of flash memory.
The reasoning for choosing sector 8 is that this sector seems to only be used when downloading a new app, in which case it is probably fine to wipe a package's memory.

TODO:

- [x] test operations with a specific package
- [x] ensure code style compliance
- [x] make assorted PR for vesc_pkg (https://github.com/vedderb/vesc_pkg/pull/11)

### DIFF
--- a/flash_helper.c
+++ b/flash_helper.c
@@ -463,3 +463,35 @@ static void qmlui_check(int ind) {
 
 	code_checks[ind].check_done = true;
 }
+
+#define VESC_IF_NVM_REGION_SIZE	(ADDR_FLASH_SECTOR_9 - ADDR_FLASH_SECTOR_8)
+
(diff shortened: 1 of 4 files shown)


vedderb:
I would prefer to move the code to flash_helper and reuse the write_data function for the writing. Also, it would be better to write an array than writing one byte at a time, you can make the same range checks then too. Also, new functions should be added to the end of vesc_c_if - if you add them in the middle you break binary compatibility with all existing packages.

Blevruz:
Cursory testing was satisfactory enough (could do the read, write and wipe operations). Going to check style then squash commits, probably.

PR #590, opened 2023-01-16
Additional SVM variations

### LINKED ISSUE #589: Feature request: Alternating Reversing sequence & Null=V0 space vector PWM
I noticed on a lot of the manuals for the VESC6 boards mention that FOC only utilizes sinusoidal commutation.  I have done some work with NULL=V0 and alternating reversing sequence space vector pulse width modulation on [my Github](https://github.com/Rampagy/bldc-firmware) and would like to try and implement it here.  I tried looking through the code to see if I could find if was already implemented, but I was not able to.

If you're willing to provide an overview of the code architecture and where you think a good injection point for this would be, it would greatly help me in my endeavor.

https://youtu.be/5eQyoVMz1dY
https://youtu.be/hSXSY4LRizg
https://youtu.be/oHEVdXucSJs

These commutation methods provide ~15% more phase to phase voltage than sinusoidal commutation, and I think it would be a great addition to the project.

supported_commutation

Rampagy:
# Additional svm implementations

Fixes #589 
https://youtu.be/5eQyoVMz1dY
https://link.gale.com/apps/doc/A18320578/AONE

### DIFF
--- a/motor/foc_math.c
+++ b/motor/foc_math.c
@@ -253,7 +253,7 @@ void foc_svm(float alpha, float beta, uint32_t PWMFullDutyCycle,
 
 		case V7_ODD_V0_EVEN:
 		case NULL_V7:
-			tA = (uint32_t)(PWMFullDutyCycle+1);
+			tA = (uint32_t)(PWMFullDutyCycle);
 			tB = (uint32_t)(T0+T2);
 			tC = (uint32_t)(T0);
 			break;
@@ -284,7 +284,7 @@ void foc_svm(float alpha, float beta, uint32_t PWMFullDutyCycle,
 		case V0_ODD_V7_EVEN:
 		case NULL_V7:
 			tA = (uint32_t)(T0+T1);
-			tB = (uint32_t)(PWMFullDutyCycle+1);
+			tB = (uint32_t)(PWMFullDutyCycle);
 			tC = (uint32_t)(T0);
 			break;
 		}
@@ -314,7 +314,7 @@ void foc_svm(float alpha, float beta, uint32_t PWMFullDutyCycle,
 		case V7_ODD_V0_EVEN:
 		case NULL_V7:
 			tA = (uint32_t)(T0);
-			tB = (uint32_t)(PWMFullDutyCycle+1);
+			tB = (uint32_t)(PWMFullDutyCycle);
 			tC = (uint32_t)(T0+T2);
 			break;
 		}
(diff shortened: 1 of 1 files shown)


motor/foc_math.c
```diff
@@ -196,6 +196,10 @@ void foc_pll_run(float phase, float dt, float *phase_var,
 /**
  * @brief svm Space vector modulation. Magnitude must not be larger than sqrt(3)/2, or 0.866 to avoid overmodulation.
  *        See https://github.com/vedderb/bldc/pull/372#issuecomment-962499623 for a full description.
+ *        Null=V0 variation (reduces switching losses by up to 33%)
+ *        See the follwoing links for a full description: 
+ *        https://youtu.be/5eQyoVMz1dY
+ *        https://go.gale.com/ps/i.do?id=GALE%7CA18320578&sid=sitemap&v=2.1&it=r&p=AONE&sw=w&userGroupName=anon%7E469317a7
```
kalvdans, line 202:
```suggestion
 *        https://link.gale.com/apps/doc/A18320578/AONE
```
Seems to be the shorter link they like us to use in references.

util/utils_math.h
```diff
@@ -119,6 +119,8 @@ void utils_rotate_vector3(float *input, float *rotation, float *output, bool rev
 #define COS_MINUS_30_DEG		(0.86602540378)
 #define SIN_MINUS_30_DEG		(-0.5)
 #define ONE_BY_SQRT2			(0.7071067811865475)
+#define PI_OVER_3				(1.0471975511965977f)	// 60 degrees
+#define SIX_PI_OVER_3			(6.0f*PI_OVER_3)		// 360 degrees
```
kalvdans, line 123: So TWO_PI then?

Rampagy:
Adding picture for documentation purposes

image

Rampagy:
Don't merge this yet.  There's a bug at 300 degrees.

Rampagy:
> Don't merge this yet. There's a bug at 300 degrees.

Good for testing now.

Teslafly:
Doesn't vesc already use the ARS SVPWM type?
So you don't get any more voltage magnitude.

You do get to eliminate the dead time for the low side phase, but that can actually provide important balancing and make the dead time for all phases equal and easier to compensate for / ignore.
It also increases stress on the low side mosfet. (But enables asymmetrical hardware with more current on the bottom mosfet)

Have you tested this on any hardware at all? Tried with / without phase voltage filters?

This would also effect the voltage sensing for hardware with low side current shunts vs phase voltage shunts.

So this is not simple. and may not be desired in a lot of cases. It has actually been discussed before and not been found to be particularly valuable with it's tack on effects. At most this may be made selectable, but probably should not replace ARS SVPWM.

If you aren't you should join us in the vesc discord. ([Discord](https://discord.gg/JgvV5NwYts)) Especially if you are really into motor control. The #esc-firmware thread is probably where you want to be for discussing this in depth.

vedderb:
Nice work! I think you got the the VESC Tool update right too.

Before merging I would like to see some more testing. As it looks to me now from the testing you did the additional methods reduce stability and whether they help with spreading heat or reducing the effects of dead time is not really confirmed. It would be good to see some experiments that show advantages and how current sensing and voltage sensing with phase filters is affected.

PR #594, opened 2023-02-15
typo fix for custom encoder read callback

lrnd:
I believe there is a typo in encoder.c for the encoder read callback. The type needs to change to allow custom implementation


lrnd:
I have tested this change using a vesc_package to set custom encoder callbacks, it works

PR #597, opened 2023-03-01
Package C Interface: Add BMS values access

surfdado:
Follows the example of gnss, passing a pointer to a volatile struct. Will allow packages to take action based on BMS data -Example: pushback based on temperature, cell voltages, etc


surfdado:
just saw your other comment...

PR #599, opened 2023-03-02
IMU default filter now MEDIUM instead of LOW

surfdado:


surfdado:
Based on the experience from the past few months the majority of new users building balance vehicles will have a better out-of-box experience with the filter defaulting to MEDIUM.

PR #600, opened 2023-03-08
Add inductance detection command to lisp

Teslafly:
Adds a inductance detection command to lisp for running foc setup inductance measurement at a specified current. 

Useful for measuring the saturation curve of a motor.

Discussion on vesc discord around here:
https://discord.com/channels/904830990319485030/937351309340860426/1080813263098482759

The lisp command works as expected. However the inductance measurement doesn't seem to reach the target current most of the time. But that seems to more be a problem or misunderstanding with `mcpwm_foc_measure_inductance_current()`.

It's probably worth including this in the 6.02 release and then fixing any problems with the actual detection later. If any.

### DIFF
--- a/lispBM/README.md
+++ b/lispBM/README.md
@@ -2705,6 +2705,26 @@ This command is useful to update the configuration before starting the motor as 
 
 ---
 
+#### conf-measure-ind
+
+| Platforms | Firmware |
+|---|---|
+| ESC | 6.02+ |
+
+```clj
+(conf-measure-res target-current optSamples)
+```
+
+Measure motor inductance with target-current. The optional argument optSamples sets the number of samples to use (default 100).
+
+returns: ({ld_lq_avg} {ld_lq_diff} {actual_measurement_current} fault-code)
+
+Can not be used when the motor is running. The measurement current is not gaurenteed to reach the target, and the actual_measurement_current parameter should be used to verify the actual current used.
+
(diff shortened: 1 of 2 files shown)


Teslafly:
example command usage from https://discord.com/channels/904830990319485030/937351309340860426/1080814341760237588
```
(match (conf-measure-ind 1)
    (((? l) (? ld_lq_diff) (? curr))
        (print (list "Ind:" l "Saliency:" ld_lq_diff "Current:" curr))
    )
    
    ((? fault) (print (list "Measurement failed" fault)))
)
```

Unpacks data values and determines if command was success or returned fault.

vedderb:
image
I think you forgot something here :)

Another note: If restart_cnt changes while the command runs you should return without interacting with lbm. That means that someone streamed or uploaded new code and restarted lbm while the command was running (which is common during development) and the context that was blocked will not exist anymore.

Teslafly:
Oh yeah. I definitely missed that debug chunk. Thanks.

"If restart_cnt changes while the command runs you should return without interacting with lbm."

Would that cause a memory leak with `lbm_start_flatten(&v, 25)` if I don't finish the flatten and just return?
I suppose the `v` variable gets pushed out of scope, but is it more of a pointer to other memory that lbm_start_flatten assigns?

pushed 2 commits that should fix the issues.

vedderb:
Om restart lbm memory is completely reset, so there won't be any memory leak.

PR #601, opened 2023-03-08
Add dedicated adc app hwconf pin defines for reverse and cruise switches

Teslafly:
Adds defines for hwconf specified adc app reverse and cruise pins that don't clobber uart or ppm inputs. 

For hardware that implements these dedicated inputs.

Without a hardware config ready to commit that actually uses these. I added the commented out config as an example to the trampa75/300 hwconf using the swd pins.

### DIFF
--- a/applications/app_adc.c
+++ b/applications/app_adc.c
@@ -67,6 +67,15 @@ static volatile bool rev_override = false;
 static volatile bool cc_override = false;
 
 void app_adc_configure(adc_config *conf) {
+
+#ifdef HW_HARDWARE_ADC_APP_PINS
+	if (!buttons_detached) {
+		palSetPadMode(HW_REVERSE_SWITCH_PORT, HW_REVERSE_SWITCH_PIN, PAL_MODE_INPUT_PULLUP);
+		if ((conf->buttons >> 0) & 1) {
+			palSetPadMode(HW_CRUISE_SWITCH_PORT, HW_CRUISE_SWITCH_PIN, PAL_MODE_INPUT_PULLUP);
+		}
+	}
+#else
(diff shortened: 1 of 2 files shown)


Teslafly:
This can probably wait until after 6.02 is released as no hardware currently uses it.

Teslafly:
are there any problems with this that you see? with 6.02 released, is there a reason not to merge this?

vedderb:
I'm not so happy with the button implementation in the adc-app as it is and I don't want to encourage people to make dedicated ports for them at the moment. It would also require some updates in the help texts and probably make them even more confusing.

If you want to use different inputs for those functions you can do that in lisp. There could maybe be a config-option to use the swd-pins.

Teslafly:
> I'm not so happy with the button implementation in the adc-app as it is and I don't want to encourage people to make dedicated ports for them at the moment. It would also require some updates in the help texts and probably make them even more confusing.
> 
> If you want to use different inputs for those functions you can do that in lisp. There could maybe be a config-option to use the swd-pins.

The thing is I would rather not have the average person have to edit lisp in order to use the controller. The documentation of the changed pins would be from the "controller vendor". I don't think the vesc tool help text would need to be changed at all. 



The other thing I was eventually thinking of for the adc app is having the FW report a list of pin names to vesc tool, and you choose one for each function (cruise, reverse, etc) from a drop down. Then the config value is the index in that hardcoded list which maps to a pin/port value.

That way you can use any generic pin and it is always obvious what is connected to what from the users perspective. It would also get rid of half the control modes as separate reverse button modes are not required if you just set the reverse button input to nothing.

vedderb:
There are already some hardwares that have extra pins that can be used from lisp. In that case it is better to update the adc-app to use a pin from a list of available pins rather than hardcode these pin names just for the adc-app. Merging this before that is done will just make that transitions more painful as those config defines and documentation around them will become obsolete then.

PR #602, opened 2023-03-09
Add hwconf for Flipsky 75300 

Teslafly:
The Flipsky 75/300 uses low side shunts, has phase voltage filters, only uses 1 mosfet thermistor, and has a terrible current path.
It comes with the `75_300_R2` firmware installed.

Teardown and some reverse engineering:
https://discord.com/channels/904830990319485030/1069426843779866774/1076053154577256531

https://flipsky.net/collections/electronic-products/products/flipsky-fsesc-75300-84v-high-current-350a-esc-base-on-vesc-with-aluminum-case-water-cooling-enclosure-for-e-foil-fighting-robot-surfboard-agv-robot


jaykup26:
Looks good!

How is the input voltage measurement at 3.3v V_REG?  I needed to adjust the V_REG to 3.4081 to get it to read accurately on the Flipsky 75300

image

jaykup26:
Also should this be in the main Flipsky folder instead of a 75_300 sub folder?  Or should all the configs be in their own sub folder?

sp3swj:
Flipsky know about problem with missing temp probes from 10th January,  send him several info ;-)

Teslafly:
> Also should this be in the main Flipsky folder instead of a 75_300 sub folder? Or should all the configs be in their own sub folder?

Originally I was going to have a no_limits version and so put things in a folder. But given that the current path is already quite fragile for the rated current, decided that a no_limits version was unwise. 

I measured the 3.3v rail at 3.322v and when  my multimeter measures 45.8v. 
With vref = 3.3v the 75/300 measures 44.5vin.
With vref = 3.413v the 75/300 measures 46.0vin.

so it probably could use the 3.413V vref value.

Teslafly:
Moved config out of it's own folder and adjusted vref to 3.413v.

vedderb:
It is good to have a separate flipsky hardware file to avoid the confusion with the phase filters. You also need to put it in the package firmware script to include it in vesc tool.

It would be good to come up with a name though that is not as easy to confuse with the VESC 75/300. With this name I can see that people easily could put this firmware on the VESC and the other one on the flipsky. Maybe Something like FSESC_300_75, FSESC300_75V.

Teslafly:
> It is good to have a separate flipsky hardware file to avoid the confusion with the phase filters. You also need to put it in the package firmware script to include it in vesc tool.
> 
> It would be good to come up with a name though that is not as easy to confuse with the VESC 75/300. With this name I can see that people easily could put this firmware on the VESC and the other one on the flipsky. Maybe Something like FSESC_300_75, FSESC300_75V.

I currently have
```
#define HW_NAME					"FSESC_75_300"
```

The product name is:
"Flipsky FSESC 75300 84V High Current 350A ESC Base On Vesc"

Would you prefer something like:
```
#define HW_NAME					"FLIPSKY_FSESC_75300"
```

I can add this to the package firmware script.

Teslafly:
Changed hw name to `"FLIPSKY_FSESC_75300"` and added to packaging script.
Also added package folder to gitignore.

vedderb:
I didn't realize they have 84V in the description and call it 75/300. That makes it even more confusing. I don't like making the hardware name string too long either. What about FSESC_84300? Also, regarding the 84V claim, it is probably good to update HW_LIM_VIN to reflect that with some margin.

Teslafly:
image
i think the 75300 should be in the name  somewhere. since that is their main name for it. for better or worse. 
using the actual hardware values would invite more confusion.



The board uses 100v fets/caps.

vedderb:
In that case I would use FSESC75300 in one word without flipsky prepended  the way they do it there and reflect that in the file names. If the name is too long it will not fit on the can device panels. I also don't want the firmware version packet to grow too much.

Teslafly:
Name changed to "FSESC75300".
Is this good to merge?

sp3swj:
> Name changed to "FSESC75300". Is this good to merge?

Good name = exact product name

Below few info and remarks.

info from web page:
Continuous Current : 50V/300A,75V/200A
Max Current :350A

We can calculate :=> 15kW  continues power = 300A @ 50V = 200A @ 75V

4,2V  fresh loaded cell  and 20 cell give us 84A
max voltage 84V - thats mean we can power 20S - but we can not take from battery 200A ad 84V

~~ 3,5V heavy loaded cell  ==> 20*3,5 = 70V

350A - peeak current - but not specified for how many seconds :-( 

350A @ 50V give us 17,5 kW 

I have email from Flipsky info that  75300 should hold 20kW for 1-2 second - but not tested yet

SUMMARIES - in my opinion: 
-save continues current for 20S is 200A  
-save continues current for 16S is 300A  


name **FSESC75300**  fit super  :-)

PR #607, opened 2023-03-30
Add regen cutoff to prevent overcharging the battery during braking

nitrousnrg:
This change helps ensure that the battery remains in good condition, prolongs its overall lifespan and avoids hazardous scenarios of overcharge.

A soft cutoff for the regenerative braking is implemented, in which start and end thresholds should be set in VESC Tool. The end threshold should be set such that the cells don't exceed 4.2v in a typical Li-Ion battery.

The default values follow the default max voltage that was being used. This means that in setups with operating voltages above ~60V, these new fields with the default values will completely disable regenerative braking until the motor config is updated accordingly for the correct battery voltage.

The code is easy to test with power supplies, as it can now brake the motor without tripping an overvoltage. It was also tested with a 60v battery.

This commit needs its companion commit on vesc tool code, I'll make that PR now

### DIFF
--- a/confgenerator.c
+++ b/confgenerator.c
@@ -28,6 +28,8 @@ int32_t confgenerator_serialize_mcconf(uint8_t *buffer, const mc_configuration *
 	buffer_append_float32_auto(buffer, conf->l_max_vin, &ind);
 	buffer_append_float32_auto(buffer, conf->l_battery_cut_start, &ind);
 	buffer_append_float32_auto(buffer, conf->l_battery_cut_end, &ind);
+	buffer_append_float32_auto(buffer, conf->l_battery_regen_cut_start, &ind);
+	buffer_append_float32_auto(buffer, conf->l_battery_regen_cut_end, &ind);
 	buffer[ind++] = conf->l_slow_abs_current;
 	buffer_append_float16(buffer, conf->l_temp_fet_start, 10, &ind);
 	buffer_append_float16(buffer, conf->l_temp_fet_end, 10, &ind);
@@ -426,6 +428,8 @@ bool confgenerator_deserialize_mcconf(const uint8_t *buffer, mc_configuration *c
 	conf->l_max_vin = buffer_get_float32_auto(buffer, &ind);
 	conf->l_battery_cut_start = buffer_get_float32_auto(buffer, &ind);
 	conf->l_battery_cut_end = buffer_get_float32_auto(buffer, &ind);
+	conf->l_battery_regen_cut_start = buffer_get_float32_auto(buffer, &ind);
+	conf->l_battery_regen_cut_end = buffer_get_float32_auto(buffer, &ind);
 	conf->l_slow_abs_current = buffer[ind++];
 	conf->l_temp_fet_start = buffer_get_float16(buffer, 10, &ind);
 	conf->l_temp_fet_end = buffer_get_float16(buffer, 10, &ind);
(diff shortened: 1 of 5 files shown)


vedderb:
I think it is best to change the defaults to something high like 1000 and 1100 V, so that you always have braking if you don't change the parameters. A lot of people rely on the braking as their primary brake (which is a bad idea), so I don't want to surprise them with a non-existent brake after a firmware update.

Also, in practice I have never seen an issue with overcharging a battery with regenerative braking but I have seen many situations where an inconsistent or disabled brake would cause a dangerous situation. I worry less about wearing out the battery slightly faster than not having a brake.

Another potential issue with this setting is that people probably don't realize that it will completely disable the brake when the battery is full. That is one of the most common misunderstandings - that it is not possible to brake using the motor without charging the battery in the process.

nitrousnrg:
Sure, I'll move the defaults to 1000+v, I had the same concern which is why I did bring it up.

For some time I stayed in a home at the top of a hill with a Tesla rental. In the downhill it charged the battery 2 to 3% if it was not fully charged, and if it left the place fully charged it would reduce regen pretty dramatically. From a user perspective it sucks but its a safety matter.

tonymillion:
To jump in, I use an e-scooter with VESC and its usually the BMS's job to limit reverse charging, I've noticed it on a full battery where regen kicks in first, then drops out leading to the scooter freewheeling until I pull the brakes harder. 

If anything, I would request this be put behind a Boolean to completely disable it, lest it adds another failure point which could lead to a bad accident. 

I see the advantage if the battery is being used without a good quality balancing BMS but I (hope) thats not the usual case (for accident & potential fire safety)

nitrousnrg:
> To jump in, I use an e-scooter with VESC and its usually the BMS's job to limit reverse charging, I've noticed it on a full battery where regen kicks in first, then drops out leading to the scooter freewheeling until I pull the brakes harder.

Yes, most BMS's will protect from overcharging, but the protection actuation is sharp and sudden. Not to mention if it happens in freezing temps and you have to reach in panic for the mechanical brakes on a red light. We can have a more predictable behavior if the controller is aware of the limits to avoid a hard BMS trip, like every EV does.
I would even show a warning message "Regenerative Braking Temporarily Reduced"

> If anything, I would request this be put behind a Boolean to completely disable it, lest it adds another failure point which could lead to a bad accident.

I think the 1000V default setting that I committed earlier today is enough.

gowrav:
> I have tried using voltage regen cutoff start at 55v and cutoff end at 56v to reduce the regen current as the BMS I am working with cutsoff contactor if current is greater than -2A to safeguard the batteries.
> 
> The issue can be reproduced by reving the motor from zero to max rpm rapidly with throttle and suddenly releasing the throttle, if it doesn't cut off at first repeating it few times will do the trick.
> 
> I have tried setting Battery Current Max Regen to Zero 0 , keeping Maximum Braking Power also 0 , with ACD Control set to Current , but it doesn't seem to work , this is especially true when the motor is in Field Weakening region.
> 
> Doesn't seem to affect if field weakening is 0., however even at this point some spikes are observed for 0 BATTERY CURRENT MAX REGEN settings too.
> 
> Any pointers here could be helpful to solve this. The image below shows the Spikes observed for Battery Regen even when the Battery voltage is > 56v and BATTERY CURRENT MAX REGEN = 0 , shortly after the battery regen spike the BMS cutsoff the contactor causing overvoltage error as the regen energy now fills up the capacitor instead of draining to the batteries.
> Screenshot 2025-05-26 at 12 04 05 AM



> This change helps ensure that the battery remains in good condition, prolongs its overall lifespan and avoids hazardous scenarios of overcharge.
> 
> A soft cutoff for the regenerative braking is implemented, in which start and end thresholds should be set in VESC Tool. The end threshold should be set such that the cells don't exceed 4.2v in a typical Li-Ion battery.
> 
> The default values follow the default max voltage that was being used. This means that in setups with operating voltages above ~60V, these new fields with the default values will completely disable regenerative braking until the motor config is updated accordingly for the correct battery voltage.
> 
> The code is easy to test with power supplies, as it can now brake the motor without tripping an overvoltage. It was also tested with a 60v battery.
> 
> This commit needs its companion commit on vesc tool code, I'll make that PR now

This functionality is what was needed for me , however the battery still trips from regen even when set with BATTERY CURRENT MAX REGEN  = 0

PR #608, opened 2023-04-03
Add 5 milliohm shunt version for Flipsky 75100v2 Hardware

groybe:
For users of 5 milliohm shunts. Useful for large servo motors with lower currents.
A few people are using this for simracing wheels.


Toni67:
Would be a nice addition indeed, default shunts are cruelly lacking definition with these servos, allow the use of R005 would be very welcome.

PR #612, opened 2023-04-13
added read observer and encoder angle functions

hackeybabel:
+ Read observer angle
+ Read encoder angle
+ Read observer error


vedderb:
One of the lines has a syntax error and the extensions also need to be registered. It is also good to update the documentation when adding new extensions.

I made an attempt here to add all position sources and documenting them:
https://github.com/vedderb/bldc/commit/523e42b829b435549f19025ced9f6e9c84f37554

Can you try if that works?

PR #619, opened 2023-04-24
Luna UI updates

nitrousnrg:
Makes it work with the updated QML loader


nitrousnrg:
Hi, bump about this. Our bikes UI is broken without it.

Cheers

PR #622, opened 2023-05-01
Pin lock - RFQ

surfdado:
This isn't meant to be merged yet, mainly posted it for people to comment on or test independently.

Basic idea is to introduce a simple method to prevent accidental writes to the wrong VESC by using a 4-digit PIN to write-lock your firmware. Unlike pairing it is actually enforced by the firmware, so rogue 3rd party apps cannot circumvent it. Also, a 4-digit PIN is easy to remember.

PIN is stored in EEPROM just like the Odometer, however this implementation does clear the PIN during firmware writes.

Functionality is fully described here:
https://pev.dev/t/new-feature-spec-pin-locking/815/4

### DIFF
--- a/comm/comm_usb.c
+++ b/comm/comm_usb.c
@@ -90,7 +90,10 @@ static THD_FUNCTION(serial_process_thread, arg) {
 }
 
 static void process_packet(unsigned char *data, unsigned int len) {
+	commands_lock_writes(false);
+	// USB commands are not subject to PIN based write lock
 	commands_process_packet(data, len, comm_usb_send_packet);
+	commands_lock_writes(true);
 }
 
 static void send_packet_raw(unsigned char *buffer, unsigned int len) {
--- a/comm/commands.c
+++ b/comm/commands.c
@@ -70,6 +70,11 @@ static THD_FUNCTION(blocking_thread, arg);
 static THD_WORKING_AREA(blocking_thread_wa, 3000);
 static thread_t *blocking_tp;
 
+// PIN-Based Write-Lock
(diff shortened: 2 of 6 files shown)


Relys:
Great feature. First of all my question is why is the official UUID pairing even a thing if it's client side? That renders it completely pointless IMHO.

Second question, is there a timeout of how many pin combinations can be attempted before requiring a power cycle (i.e. preventing a brute-force attack)?

surfdado:
First question, idk - not my implementation. It was developed back when VESC Tool was the only tool/app available to prevent accidental pairing of the wrong remote.

Second, we can consider making it more hacker proof later, right now the aim is to keep it simple and idiot proof.

Relys:
I submitted a PR to @surfdado branch for this btw: https://github.com/surfdado/bldc/pull/4

"Adds a new command to COMM_WRITE_UNLOCK_CMD where you can append the pin before the actual command and payload which will handle writelock unlocking/locking while handling the command. Calls the command processing function recursively like COMM_FORWARD_CAN. Fixed some vulnerabilities with stack overflow by limiting recursion depth (which was also in COMM_FORWARD_CAN) as well as fixing brute force attempts for writelock."

surfdado:
> I submitted a PR to @surfdado branch for this btw: [surfdado/bldc#4](https://github.com/surfdado/bldc/pull/4)

I'll check it out as soon as I get a chance!

tonyt321:
would like this to be added officially would save a lot of confusion when connecting with multiple boards around

PR #624, opened 2023-05-03
Duplicated include; improved shutdown.c; store data even without switch

Gabrielerusso:
commands.c: 
- Delete duplicated include "mc_interface.h" 

shutdown.c / shutdown.h / main.c:
- Reducing possibility to write backup data frequently, now requires  minimum distance to be traveled before it enables storing, common condition if push to start is enabled with a low shutdown time, say 10s, or in any other condition where user turn on/off ESC a lot of times. Saves chip life at the cost of not having updated data during very short trips.
- Data is stored even when using "always on", basing storage condition on the inactivity time, this allower user that shutdown esc from the power source to keep data anyway or in case of a power loss.
- Introducing possibility to store data without a power-switch by using a combination of trip distance and inactivity timer, it estimates when the scooter it's stopped for a long period after a trip, now user to save trip data just need to leave the scooter still an X amount of time to make sure all data is stored. It can be **deactivated** by using "_always off_" in app setting  Shutdown Mode.

Compiled almost all version without issues, tested on my custom esc with different combinations (with power switch, without power switch) and it seems to work well.

### DIFF
--- a/hwconf/shutdown.c
+++ b/hwconf/shutdown.c
@@ -20,6 +20,7 @@
 #include "shutdown.h"
 #include "app.h"
 #include "conf_general.h"
+#include "mc_interface.h"
 
 #ifdef USE_LISPBM
 #include "lispif.h"
@@ -67,7 +68,11 @@ void shutdown_set_sampling_disabled(bool disabled) {
 }
 
 static bool do_shutdown(void) {
-	conf_general_store_backup_data();
+	// Store data if 1000m or more have been done to possibly reduce flash
+	// wear. Values could be lower than 1km, say 100m but suggested >500m.
+	if(mc_interface_get_distance_abs() >= 1000) {
+		conf_general_store_backup_data();
+	}
 #ifdef USE_LISPBM
 	lispif_process_shutdown();
 #endif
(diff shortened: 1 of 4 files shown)


Gabrielerusso:
Tested in those weeks with 4 escs and different configuration, 2 with power switch and 2 without, now i was able to save data, like the odometer without problems even in the version without the power switch.

hwconf/shutdown.c
```diff
@@ -67,7 +68,11 @@ void shutdown_set_sampling_disabled(bool disabled) {
 }
 
 static bool do_shutdown(void) {
-	conf_general_store_backup_data();
+	// Store data if 1000m or more have been done to possibly reduce flash
```
vedderb, line 71: I think it is safe to assume that writing the backup date is fine when a shutdown is triggered. Otherwise a lot of odometer data can remain unlogged.

Gabrielerusso: the amount of **_possible_** odometer data unlogged can be reduced by the distance, even 1m could be used even if it's  pretty useless as not even car odometers have such accuracy, say one could set 10m or 100m.
Anyway it was just a suggestion as we don't use external eeprom like in automotive odometers.

1zun4:
Please merge this. This solves my issue https://github.com/vedderb/bldc/issues/638 as well, because the only reason I need it to be able to store odometer data without software shutdown on the hardware.

hwconf/shutdown.c
1zun4: Yes, I think 1000m is fine for accuracy, we do not need the exact metres. It may shift over time, but from someone who does not have ANY mileage data, this is a very good start. The only thing that worries me is the flash enurance of the F405, could this become a problem?

PR #635, opened 2023-05-31
Perform FOC offsets calibration when running FOC detection

TechAUmNu:
Undriven offsets are not measured on startup as the motor might be spinning (reboot from crash, etc) So the only time undriven was measured was manually in VESC Tool.
This runs the full offset calibration before detecting the motor to improve detection and make sure the undriven offsets are measured.


vedderb:
Sorry for the delay on this one. We can add it, but I think it is best without including the undriven offsets as I have seen them make things worse on some setups.

PR #669, opened 2023-10-31
fixes detached ADC1 not being mapped

1zun4:
I have received multiple reports of the ADC not working correctly when the throttle input is disconnected by Lisp. After looking at the latest changes, I noticed that the pwr value was overridden after it was mapped, whereas before it was mapped. This only affected ADC1, as ADC2 had not been changed and was working as before.

I have now moved it up again and it should now work correctly. I have not tested it on my scooter yet, but will comment as soon as I do if the change has fixed it, but this is a very obvious logic flaw, so it should work.

Code changes: https://github.com/vedderb/bldc/compare/8c1e6c53de717b3feea6690d6b59f3c8d9e81a97...f08c8602fe1739fff4349d9ffc7989bf561c2229

from:
https://github.com/vedderb/bldc/blob/8c1e6c53de717b3feea6690d6b59f3c8d9e81a97/applications/app_adc.c#L202-L230
to:
https://github.com/vedderb/bldc/blob/f08c8602fe1739fff4349d9ffc7989bf561c2229/applications/app_adc.c#L202-L230

### DIFF
--- a/applications/app_adc.c
+++ b/applications/app_adc.c
@@ -201,6 +201,11 @@ static THD_FUNCTION(adc_thread, arg) {
 
 		range_ok = read_voltage >= config.voltage_min && read_voltage <= config.voltage_max;
 
+		// Override pwr value, when used from LISP
+		if (adc_detached == 1 || adc_detached == 2) {
+			pwr = adc1_override;
+		}
+
 		// Map the read voltage
 		switch (config.ctrl_type) {
 		case ADC_CTRL_TYPE_CURRENT_REV_CENTER:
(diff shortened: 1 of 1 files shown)


vedderb:
To me it makes more sense to not map the override-value and give an override of 0 to 1 rather than reverse the map. That should be done for ADC2 as well. How are you using the override? Is there a reason to simulate a throttle voltage rather than giving a value between 0 to 1?

1zun4:
Yes it absoluty does make sense that ADC1 is being mapped using the ADC app! We are using this to have a throttle input from a display which can vary depending on the throttle that is being used. In the past people had to change the value inside the Lisp script themselves, but since we changed to ADC this was finally being done using the app itself like a real hardware throttle.

vedderb:
That makes sense. Merging!

vedderb:
I moved it up even further as I think that makes sense when simulating a raw throttle, hope that works too: https://github.com/vedderb/bldc/commit/b1bad56859b691f69256c2a3de207a6f8a56e767

PR #671, opened 2023-11-06
When generating res_fw.qrc, only use compiled firmwares

alhirzel:
This prevents an error message when you only compile a few of the firmware variants, allowing `res_fw.qrc` to be generated for integration with VESC Tool.

It could be that this script should error when not all firmwares are present? Please advise if so.


alhirzel:
Thanks @vedderb!

PR #677, opened 2023-11-17
Add define to disable LEDs

Tomblarom:
I added a define to disable the use of LEDs if the hardware misses them. In my case there where none, in order to use all GPIOs as external inputs. I took `CAN_ENABLE` as example. Obviously by default `LED_ENABLE` defined as `1` and only set to `0`, if `HW_HAS_NO_LED` is set in the `hwconf/hw_xx.h`.

### DIFF
--- a/conf_general.h
+++ b/conf_general.h
@@ -109,6 +109,18 @@
 #endif
 
 /*
+ * Enable green and red LED
+ */
+#ifndef LED_ENABLE
+#define LED_ENABLE					1
+#endif
+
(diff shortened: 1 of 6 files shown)


Tomblarom:
According to @vedderb another solution would be to simply empty out the corresponding variables, like this:

```
#define LED_GREEN_ON()
#define LED_GREEN_OFF()
#define LED_RED_ON()
#define LED_RED_OFF()
```

PR #694, opened 2023-12-25
Package C-Interface: add APIs for HW/FW version/names

surfdado:
This API has been inserted after "foc_set_openloop_duty_phase" in order to make it possible to safely identify v6.05 vs v6.02 firmware.


surfdado:
I just realize that Lisp sysinfo has all this already

PR #695, opened 2023-12-31
COMM_SHUTDOWN command for shutdown/restart

surfdado:
cmd[0] = 111 // magic number
cmd[1] = [force] // 0=normal, 1=force (allow while motor is spinning) 
cmd[2] = [restart] // 0=shutdown, 1=restart

Obviously, shutdown only works for hardware with shutdown support

### DIFF
--- a/comm/commands.c
+++ b/comm/commands.c
@@ -1621,6 +1621,26 @@ void commands_process_packet(unsigned char *data, unsigned int len,
 		conf_custom_process_cmd(data - 1, len + 1, reply_func);
 	} break;
 
+
+	case COMM_SHUTDOWN: {
+		int ind = 0;
+		int force = data[ind++];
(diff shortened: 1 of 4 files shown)


vedderb:
I think the magic number is not needed really. If you connect something that sends undefined commands that pass the checksums you are in trouble anyway.

surfdado:
Well, since my pin lock PR hasn't gone anywhere there is actually one app with hundreds of users out there that uses that COMM id for pin lock - so the problem is not just hypothetical.

If we agree to define COMM_SHUTDOWN as 155 then we can avoid the magic nr I guess.

surfdado:
Just force-pushed a fixup to resolve merge conflicts...

vedderb:
You could put a pin-lock command in-between as a placeholder for now.

surfdado:
Okay, done and pushed

PR #697, opened 2024-01-02
Resample shutdown before disabling gates in ALWAYS_OFF mode

### LINKED ISSUE #686: [LispBM] Shutdown event issues
As a follow-up to #666, I found that when using `shutdown-hold`, depending on how fast the power switch is cycled, shutdown respectively shutdown cancellation is not working reliably.

For example, when delaying shutdown with `shutdown-hold true`, then switching the power button on again while still performing the desired shutdown operation that would end with a `shutdown-hold false`, the VESCs will stay on as expected but (iiuc) the gate remains disabled, so it is no longer possible to drive the motors. It also appears that once in this state, the power switch is no longer functioning, so the VESCs will remain on (i.e. cannot be turned off) regardless of power switch state unless manually removed from power.

To illustrate:

VESC-shutdown drawio

Perhaps the power switch is no longer sampled, and gate reactivation logic does not trigger because the shutdown operation now takes longer? Mostly guessing at this point, though :)

dcodeIO:
Fixes #686 by resampling the shutdown switch in ALWAYS_OFF mode after cleanup code has run.

Previously, gates were unconditionally disabled, even if the shutdown switch has been turned back to ON state during a potentially long-running `shutdown-hold true` ... `shutdown-hold false` sequence, in effect shutting down the system immediately on subsequent shutdown switch activations instead of firing `event-shutdown` again.

### DIFF
--- a/hwconf/shutdown.c
+++ b/hwconf/shutdown.c
@@ -72,7 +72,7 @@ void shutdown_hold(bool hold) {
 	m_shutdown_hold = hold;
 }
 
-static bool do_shutdown(void) {
+static bool do_shutdown(bool resample) {
 	conf_general_store_backup_data();
 #ifdef USE_LISPBM
 	lispif_process_shutdown();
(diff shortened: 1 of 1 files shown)


vedderb:
Nice work!

PR #703, opened 2024-01-27
Add ability to use custom pins in lisp

TechAUmNu:
To use, add defines in hw config
// Lisp custom pins
#define PIN_HW_1_GPIO 	GPIOC
#define PIN_HW_1 		13

#define PIN_HW_2_GPIO 	GPIOB
#define PIN_HW_2 		2

Ideally this would be expandable to more than 2 pins.


lispBM/c_libs/vesc_c_if.h
```diff
@@ -222,6 +222,13 @@ typedef enum {
 	VESC_PIN_HALL5,
 	VESC_PIN_HALL6,
 	VESC_PIN_PPM,
+#ifdef PIN_HW_1
```
vedderb, line 225: This enum should always have the pins defined even when not present on the hardware, otherwise things won't be aligned if more pins are added later. Otherwise the PR looks good.

TechAUmNu: Thanks for the review, made it so the enum always defines the pins now.

PR #705, opened 2024-02-09
Add ability to calibrate current sensors, simplify ifdefs in foc, higher resolution sampled current

TechAUmNu:
Changes:
-Allows for individual calibration of each current sensor. Hw_conf can define CURRENT_CAL1,2,3... to set individual calibrations for each current sensor. By default no calibration is used. (x1.0)
-Simplifies a little of the foc code to reduce the number of ifdefs, by making hw.h set GET_CURRENT3() as 0 when HW_HAS_3_SHUNTS is not defined
-Only puts the sample into sampled data variable (ADC_curr_norm_value) after doing sensor picking, I think it is a bit better for reading now.
-Sampled data now stores raw or scaled current values based on the selection. Actual currents are stored at 8x resolution (same as total current now) which allows viewing currents that are oversampled. 

Background:
Current sensors should ideally have exactly the same resistance and match perfectly. In reality the measurements can actually vary quite a lot due to board layout, soldering, temperature, noise on sensor pins, etc. This results in multiple of the same esc measuring different values for R and L. In addition to the absolute resistance or gain being different between boards it can also be different between phases. e.g. with DC mode you can measure different current when commanding +10A or -10A.
Due to bad current sharing between shunts there can also be large variation between phases in how they measure current. e.g. with 3 phases side by side, the shunts of the outside phases nearest the middle phase will carry more current than the outer shunts due to lower inductance / resistance across the pcb or busbars. If all phases have 4 shunts and measure kelvin across the same shunt on each phase, say the leftmost shunt, then the left phase will measure less current than the right phase. 

Testing:
-I have tested on FOC, HFI, DC and BLDC modes on 2 low side, 3 phase and 3 low side shunts. 

**Would be good to check this doesn't affect anything for dual hardware.** 
For dual parallel the phases are controlled as one, so calibration is applied to the group.
For dual motor each motor has its own calibration values.

How to use:
Currently my method to calibrate is looping a known current through each phase output to ground using a calibrated power supply, the low side mosfet is just held on. Another way is to command a certain current and measure it with a calibrated current sensor.
Then the correction is saved in external / internal eeprom to persist across firmware updates.  I currently do this using a LISP script.


vedderb:
This one is a bit tricky, but probably worth having. I'm traveling now, but I will review it when I get back home.

vedderb:
I did some testing on dual hardware with both 2 and 3 shunts and it seems to work. I still have to test on dual hardware with phase shunts, but I don't have one at hand right now. I didn't do any testing with dual parallel as it is a bit inconvenient to set up, but I hope that someone who uses it gives it a try before the release of 6.05.

PR #707, opened 2024-02-14
Advanced traction control - work with single ESC too

Gabrielerusso:
Adds a Traction control that uses derivate of rpm (rate of change of rpm) to calculate the new current instead of the classic difference of the RPM.
For the moment it's possible to setup it via lisp console, some configuration are commented as the apps doesn't have the two new field.

It's completely configurable via LISP and has a quick "level" setting like automotive motorcycles.

This is separate from the rpm-difference TC, and you can chose to use both, or to use only one:
1. Advanced TC + Standard TC (only with multiple ESC)
2. Adavanced TC with any setup
3. Standard TC (only with multiple ESC)

- **tc_max_rpm_rate** : rate of change in erpm/s for small motors (<12") should be between 3000 and 150000, the higher the number more sharp will be the response.
- **tc_level** : how much the traction controll will affect riding in %, <1.0 means no effect, 99.99 means maximum effect; common values are between 20% and 90%, an higher percentage means a more aggressive traction control.

_TODO: maybe calculate this for each other ESC in case of multiple ESC, by now it's only done on the master_

_Note: i think this could also be used, if set correctly, to limit acceleration to preserve battery life while keeping motor current high for when it's needed, as it will limit only the acceleration but not the current at a steady speed._


vedderb:
The confgenerator-file should not be edited manually as it is autogenerated from vesc tool when editing the configuration xml files.

What this looks like to me is a limit on the RPM rate of change. Maybe it helps when wheels start to spin, but I'm not sure that is what usually is meant by traction control. If a RPM derivative limit is useful it should be added to the General -> RPM settings and implemented in the mc_interface-function update_override_limits. Doing it that way means that all apps and inputs can take advantage of it.

Gabrielerusso:
> The confgenerator-file should not be edited manually as it is autogenerated from vesc tool when editing the configuration xml files.

Didn't know that, just changed files to test it out.

> What this looks like to me is a limit on the RPM rate of change.

Yes, this is a RPM rate of change limiter.

> Maybe it helps when wheels start to slip, but I'm not sure that is what usually is meant by traction control.

We usually use wheel slip for ABS (i noticed TC is not used while braking) and TC, wheel slip ratio, and is defined (more or less) as $slip=({RPM_{driven} \over RPM_{undriven}} - 1)*100$ or more precisely $slip=({Speed_{wheel} \over Speed_{vehicle}} - 1)*100$ .  Other than that accelerometers are used for Stability Control and other things we don't care much as the moment. 
_Note that we could use this info for another PR to abstract the "TC ERPM difference limit" now used in the TC for a more generic Slip % that would work to any configuration._

In our case something similar is done using a RPM thresold and doing the RPM difference of the motors **BUT** as we are not using an undriven wheel, and the rpm is not always equivalent to the real ground speed it basically acts like a RPM sincronizer more than an actual TC. This mean that if both (2 wheel used as reference) wheels start slipping (and they do in ice/snow/mud) we won't have TC anymore but a simple syncro between the two wheels, and that will lead to slipping, more than a TC the actual is a limited slip differential.

Car and motorcycles uses perforated wheels mounted on disc brakes (and other sensors) to calculate slip, but this is almost impossible for our use cases. 
To compensate for that the rpm rate of change ($dRPM \over dt$) limiter could be very helpful if set-up correctly to limit the $dRPM \over dt$ to the actual physical maximum value.

Example: maximum possible for that motor in optimal working condition is ${dRPM \over dt}=10{RPM \over s}$ if ${dRPM \over dt}>10{RPM \over s}$ this mean we are surely slipping (or running without any load) , if we are ${dRPM \over dt}<=10{RPM \over s}$ this mean we **PROBABLY** are inside our normal working range. 
Reading from accelerometer could be used to be more precise, as if $dRPM \over dt$ is increasing and rate of change of accelerometer in one specific direction is not increasing then this mean we are slipping in that direction, but not everyone has an IMU installed.

Gabrielerusso:
just to recap:

- the actual TC it's not a real TC but something like a  limited-slip differential, a wheel speed synchronizer with a limited speed difference.

- my "advanced TC" is not a real TC either, it's a RPM rate of change limiter.

- we could implement the single setting for those two things, using their real names, and also implement a more realistic TC with a slip % setting.

- we could just add the RPM rate of change limiter, leaving the current "TC" as it is.

Gabrielerusso:
just to recap:

- the actual TC it's not a real TC but something like a  limited-slip differential, a wheel speed synchronizer with a limited speed difference. 

- my "advanced TC" is not a real TC either, it's a RPM rate of change limiter.

1) implement the single setting for those two things, using their real names, and also implement a more realistic TC with a slip % setting. In this case we would have 3 settings: 
     1.  limited-slip differential max ERPM (actual TC max ERPM, in ERPM);
     2.  RPM max rate of change (in ERPM/s);
     3.  Traction Control Slip Thresold (in % where 0 means maximum effect, 100% minimum effect) or call it TC Level and do the 
          opposite with 0% no effect (disabled) and 100% maximum effect;

2) we could just add the RPM rate of change limiter, leaving the current "TC" as it is. In this case we would have 4 settings: 
     1.  TC enabled (0/1 , current setting - no editing);
     2.  TC max ERPM (in ERPM, current setting - no editing);
     3.  ERPM max rate of change (in ERPM/s);
     4.  ERPM rate of change limiter level (in % , 0 means no effect, 100 means maximum effect);

Gabrielerusso:
- [X]  RPM rate of change limiter moved to mc_interface and max_erpm_rate to mc_configuration struct.
- [X] enable/disable TC from Lisp
- [X] Using a combination of slip ratio and "slip differential" for Traction Control
- [ ] Test new changes

FDResearch:
Did you ever get a chance to test this further?

Gabrielerusso:
> Did you ever get a chance to test this further?
Yes, tested this summer, now i'm back to test but this repo is full of old things so i'll close this one and start a new clean one only with the working and essential stuff

PR #709, opened 2024-02-20
FOC speed control: Limit m_speed_pid_set_rpm to achievable RPM when ramp rate is non-zero

ljungholm:
- If m_speed_pid_set_rpm is allowed to exceed foc_openloop_rpm before index is found, then there will be a jump in torque when index is found. 
- If m_speed_pid_set_rpm is allowed to exceed [l_min_erpm, l_max_erpm] then there is a wind up effect since it will take time for m_speed_pid_set_rpm to return to valid range due to ramp limit.

### DIFF
--- a/motor/foc_math.c
+++ b/motor/foc_math.c
@@ -444,7 +444,7 @@ void foc_run_pid_control_pos(bool index_found, float dt, motor_all_state_t *moto
 	}
 }
 
-void foc_run_pid_control_speed(float dt, motor_all_state_t *motor) {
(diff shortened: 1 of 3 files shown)


ljungholm:
I created an account in order to post on https://vesc-project.com/forum as requested in CONTRIBUTING, but I do not have permission to create a new post and can't find where to request that permission.

jaykup26:
The VESC project forum is pretty inactive these days.  Check out the VESC Discord instead

https://discord.gg/JgvV5NwYts

PR #713, opened 2024-03-02
fix ZV freq for auto detection

hackeybabel:
lower the zv freq for flux linkage measurement as the hardwares with phase shunts and v0-v7 enabled will trigger watchdog reset as you cannot run more than 30khz with v0-v7 enabled.

### DIFF
--- a/conf_general.c
+++ b/conf_general.c
@@ -1818,7 +1818,7 @@ int conf_general_detect_apply_all_foc(float max_power_loss,
 #ifdef HW_HAS_DUAL_MOTORS
 	mcconf->foc_f_zv = 25000.0;
 #else
-	mcconf->foc_f_zv = 40000.0;
+	mcconf->foc_f_zv = 25000.0;
 #endif
 	mc_interface_set_configuration(mcconf);
 


TechAUmNu:
Better to just make the temporary config turn off V0 & V7 sampling.

hackeybabel:
> Better to just make the temporary config turn off V0 & V7 sampling.

But if with dual motors 25khz works fine why not use it with single motor as well.

TechAUmNu:
I did a lot of testing and found that 40khz worked better when detecting low indictance motors. The 25khz is just a compromise because of the limitations of dual hardware. 

https://github.com/vedderb/bldc/pull/533

TechAUmNu:
Also only the wizard will set this to 40khz you can run it at whatever frequency you want from the foc page buttons. 

So by just making sure that the conflicting setting is disabled temporarily during the wizard there is no change to functionality. 
Does your hwconfig have v0&v7 on by default?

vedderb:
Is this ok?
https://github.com/vedderb/bldc/commit/adfe4a736ea060a9fb462f82a719cb714d94a063

TechAUmNu:
That looks good

PR #723, opened 2024-05-15
Make the command for getting the Python version quite

Lauszus:
"python" is not available on my system only "python3", so this will cause output stderr even though the Makefile later checks if "python3" exist


kalvdans:
Just unconditionally use `python3`. I don't know any Python3 system that only has  a `python` executable without a `python3`.

That said, patch looks good.

kalvdans (approved):

PR #725, opened 2024-05-15
Hw-RPS_FFB

adilrepas:
I made a controller for direct drive force feedback steering wheel which is used to play racing simulator games.
I posted a picture of the controller on the Vesc discord


adilrepas:
I realize there are some errors

PR #736, opened 2024-06-30
New hwconf files: ADV200+500 and Thor300

surfdado:
Floatwheel ADV200 and ADV500
Fungineers Thor300

REMOVED >> Also added a generic buzzer on the servo port in a non-intrusive way. Only enabled if SERVO_BUZZER is defined in the hw header. <<

Thor uses it to give feedback when pressing the momentary button to power off the board.


vedderb:
I don't like the buzzer thing, it adds files, functions, names and noise for just toggling a pin at boot on some hardware. Also, many buzzers require that you actually PWM them.

I would suggest the following:

bild

Replace this line with a macro HW_BOOT_DELAY() that is defined to be chThdSleepMilliseconds(500) by default in hw.h. Then hwconfigs can override that define and used it to beep a buzzer, flash LEDs etc.

Regarding this change, what are you trying to achieve by removing it? Most apps and other threads should already be up and running before this sleep finishes. It only delays the boot ready CAN-frame.
bild

surfdado:
your concerns should all be addressed now

surfdado:
@vedderb could you merge this now?

PR #750, opened 2024-07-27
update RSR_DD hw and adding RSR_DD v2.1

adilrepas:
update RSR_DD hw and add RSR_DD v2.1 with different current sensors


adilrepas:
@vedderb 
I hope it doesn't miss the release

PR #755, opened 2024-08-15
hwconf/*: disable DAC buffer to increase accuracy

c0d3z3r0:
According to the STM32F4 datasheet, the ADC is equipped with a buffer for being able to drive external loads directly. This buffer negatively impacts accuracy, though. The offset can be as high as 0.2V.

Thus, disable the buffer by setting the BOFFx flag.

This was verified on a STM32 nucleo64 F446RE by reading the DAC voltage back via ADC and comparing results with and without BOFFx=1. The output voltage was verified (calibrated DMM BM789) to be correct with BOFFx=1.

Reference:
STM DS8626 STM32F40xxx Rev 9
 - 5.3.25 DAC electrical characteristics
 - Figure 53. 12-bit buffered non-buffered DAC

### DIFF
--- a/hwconf/other/hw_uxv_sr.c
+++ b/hwconf/other/hw_uxv_sr.c
@@ -265,7 +265,7 @@ THD_FUNCTION(dac_thread, arg) {
 	chRegSetThreadName("DAC");
 	chThdSleepMilliseconds(3000);
 	RCC_APB1PeriphClockCmd(RCC_APB1Periph_DAC, ENABLE);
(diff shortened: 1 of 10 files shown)


TechAUmNu:
The question is if there is enough current capability on the DAC without the buffer to actually drive the load? Some shunt amps load the reference pin. 

Also since the input current is calibrated anyway does this actually help in practice?
Does it reduce the noise on the reference voltage?

PR #770, opened 2024-09-29
Zero vector frequency randomization

lolwheel:
Adds Zero vector randomization so the acoustic profile of FET switching is less annoying.

I tested this change on my Floatwheel and made sure that `all_fw` target builds without issues.

I don't have a good way of testing this change on a dual-motor setup, any tips on how to do this would be appreciated.

### DIFF
--- a/confgenerator.c
+++ b/confgenerator.c
@@ -64,6 +64,7 @@ int32_t confgenerator_serialize_mcconf(uint8_t *buffer, const mc_configuration *
 	buffer_append_float32_auto(buffer, conf->foc_current_kp, &ind);
 	buffer_append_float32_auto(buffer, conf->foc_current_ki, &ind);
 	buffer_append_float32_auto(buffer, conf->foc_f_zv, &ind);
+	buffer_append_uint32(buffer, conf->foc_f_zv_bandwidth_pct, &ind);
 	buffer_append_float32_auto(buffer, conf->foc_dt_us, &ind);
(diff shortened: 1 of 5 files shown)


vedderb:
This one has many potential dangers and will almost certainly not work with dual motors. I also don't know how much the rand-function will impact performance. It will take me quite significant effort to test all of this. Before I do that, have you confirmed that it actually has any benefits? Is it more silent? Does it help with EMI? If it is just an "interesting feature" without any observed worthwhile benefits I don't think it is worth adding as it several potential problems that I might not even be able to find yet and it has impact on the performance.

lolwheel:
I agree that this is a risky change. In my opinion it does provide improvements in the sound characteristics of the motor.

`rand()` is usually pretty fast but I think I agree with you that its performance is not a given. I can probably just inline a simple LCG random number generator in the code, it's like 2 lines.

I do not have a way of measuring EMI explicitly, neither is this my goal with this change.

What are your concerns regarding the change not working in dual motor setup? I'll be happy to address them better.

vedderb:
My floatwheel has an almost entirely silent motor (maybe not all of them are like that). Having that almost non-existent sound change characteristics slightly is not a good enough motivation for me for a change with this impact. Maybe I'm wrong here though and my floatwheel is an exception, so if you have some demo and good arguments I'm prepared to change my mind here.

How does it change the HFI-sound? That is one area where there potentially is something to gain.

Here is one comment about it from the discord that is not very convincing:
bild

PR #780, opened 2024-11-23
Adding a persistent and resettable tripmeter

r3n33:


r3n33:
Moving trip meter to another device/repo

PR #786, opened 2024-12-03
Update Doc for added support for WRGB color order for WS2814 chip.

Relys:


Relys:
https://github.com/vedderb/vesc_express/pull/50

PR #797, opened 2025-01-22
Add option for 16Mhz XTAL as well as 8Mhz

etharooni:
Custom VESC hardware uses 16Mhz crystal.  I needed support from this in its specific hwconf file.


zamonary1:
I think that's a good change to be implemented. The changed code seems to implement the default behavior for 8mhz crystal and keep the compatibility with existing hwconf files, while providing an option to use other HSE values for hardware developers.

PR #802, opened 2025-03-04
support canbus error returns when transmitting

sdb4:
return the results of the lower-level can_transmit functions so that calling code can check if a transmission error occurred

### DIFF
--- a/lispBM/c_libs/vesc_c_if.h
+++ b/lispBM/c_libs/vesc_c_if.h
@@ -394,8 +394,8 @@ typedef struct {
 	// CAN
 	void (*can_set_sid_cb)(bool (*p_func)(uint32_t id, uint8_t *data, uint8_t len));
 	void (*can_set_eid_cb)(bool (*p_func)(uint32_t id, uint8_t *data, uint8_t len));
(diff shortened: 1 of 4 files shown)


lispBM/c_libs/vesc_c_if.h
```diff
@@ -394,8 +394,8 @@ typedef struct {
 	// CAN
 	void (*can_set_sid_cb)(bool (*p_func)(uint32_t id, uint8_t *data, uint8_t len));
 	void (*can_set_eid_cb)(bool (*p_func)(uint32_t id, uint8_t *data, uint8_t len));
-	void (*can_transmit_sid)(uint32_t id, const uint8_t *data, uint8_t len);
-	void (*can_transmit_eid)(uint32_t id, const uint8_t *data, uint8_t len);
+	msg_t (*can_transmit_sid)(uint32_t id, const uint8_t *data, uint8_t len);
```
vedderb, line 397: Changing the return types in vesc_c_if will break binary compatibility with code built against the old version. Also, msg_t is not available in vesc_c_if when using it on its own. You could make a wrapper around the old type in lispif_c_lib and keep vesc_c_if unchanged.

PR #805, opened 2025-03-19
Update hw_mksesc_84_200_hp.h

belge96:
battery voltage reading fix pretty much


belge96:
meant to change the 84_100_HP rather than 200, albeit i'm pretty sure it has the same issue

PR #807, opened 2025-03-22
Modify incorrect # endif and default configuration

makerbase-motor:
Hello, respected Mr. Benjamin:
I need to make the following modifications based on known issues:
1. Modify the incorrect # endif
2. Modify the default configuration to avoid new users mistakenly configuring reverse current.
3. Modify hardware parameters and fix voltage reading errors.


vedderb:
The indentation should be 4 spaces per tab. Can you make sure that your editor is set to that?

makerbase-motor:
OK， I have already made revisions to the indentation

PR #812, opened 2025-04-12
HWCONF - JetFleet F6 Variants

conradopoole:
- Added JetFleet HW confs for the 20s, 24s, and 32s variants.
- Modified package_firmware.py to add all the variants under JetFleetF6

### DIFF
--- a/package_firmware.py
+++ b/package_firmware.py
@@ -105,9 +105,9 @@ package_dict["TRONIC_250R"] = [['TRONIC_250R', default_name]]
 package_dict["X12_PRO24"] = [['x12_pro24', default_name]]
 package_dict["X12_PRO30"] = [['x12_pro30', default_name]]
 package_dict["Thor300"] = [['Thor300_20s', default_name]]
-package_dict["JetFleetF6"] = [['JetFleetF6_20s', default_name],
-                    ['JetFleetF6_24s', default_name],
-                    ['JetFleetF6_32s', default_name]]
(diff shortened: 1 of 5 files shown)


package_firmware.py
```diff
@@ -105,7 +105,9 @@ def get_git_revision_short_hash() -> str:
 package_dict["X12_PRO24"] = [['x12_pro24', default_name]]
 package_dict["X12_PRO30"] = [['x12_pro30', default_name]]
 package_dict["Thor300"] = [['Thor300_20s', default_name]]
-package_dict["JetFleetF6"] = [['JetFleetF6', default_name]]
+package_dict["JetFleetF6"] = [['JetFleetF6_20s', default_name],
+                    ['JetFleetF6_24s', default_name],
+                    ['JetFleetF6_32s', default_name]]
```
vedderb, line 110: This does not look right to me. For the firmware matching in vesc tool to work each of these should be on its own line. You only combine them like this if you want to have multiple firmware options for the same hardware name. Maybe I don't fully understand the intention here...

conradopoole: It's just that the HW is pretty much the same besides the FETs supporting different max voltages, they are all JetFleet F6, so I thought I would called them all JetFleetF6 and then just choose the variant in the submenu 20s, 24s or 32s.

But if it makes more sense to follow having a complete separate entry in the dictionary for each variant, let me know and I will update it.

conradopoole: I just made the change to split it back to 3 entries following your recommendation

PR #840, opened 2025-07-05
Compile with newlib-nano

lukash:
With enabled floats for scanf and printf. Saves about 22kB of flash (with GCC 10). Haven't tested what this actually does on GCC 7.

I tested flashing and running motor control only briefly for now, will test more later. The float support for printf/scanf seems to be working fine in the terminal. I am not aware of other unsupported functionality in newlib-nano.

When changing to GCC 14, the size increased slightly (by <800B). The generated code size actually decreased significantly, but I guess new stuff was added to newlib again. Then again, I'd assume there are good reasons for that and expect the new newlib-nano to just work better than the 10 years old one...

I'd like to ask to get this in and start compiling with GCC 14 if everything works properly. There have been great improvements to GCC over the recent years. Meanwhile, GCC 7 has been deprecated and is being (or has been) phased out of most distros.

On a bit of a side note, changing from GCC 7 to GCC 10 for the on-CPU IMU bitbanging communication led to 5-10% decrease in CPU load. Might get even better with GCC 14, I haven't tested that yet.

Note: I don't think the way I added this to the ChibiOS makefile rules is the correct way to do it, but I haven't found a clean way to pass it in from the outside. May not bother you as much as it bothers me :smile:.


vedderb:
That looks like a nice size reduction that is worth using. For now I will leave things as they are until the 6.06-release, but for 7.0 I will merge it and do some testing with the later compiler versions. It is interesting that the cpu usage went down in your tests, last time I tested things not only grew in size but they also slowed down on later GCC versions.

lukash:
Did some testing and with this float formatting via `VESC_IF->printf()` actually works in the package. Formatting 6 ints with this and GCC 14 takes a bit over 150B less than on GCC 10 without this change.

PR #876, opened 2025-11-27
fast filtering on phase current values

Gabrielerusso:
Implemented light filtering on phase current readings

- Added a 3‑sample sliding window average followed by two cascaded first‑order low‑pass filters.
- Designed to be computationally efficient with minimal phase delay.
- Filtering smooths noisy current measurements while raw values are still used for overcurrent protection, so safety mechanisms remain unaffected.

Also does sampling only when needed, saving some clock cycle.

### DIFF
--- a/motor/foc_math.h
+++ b/motor/foc_math.h
@@ -44,6 +44,8 @@ typedef struct {
 	float phase_sin;
 	float i_alpha;
 	float i_beta;
+	float i_alpha_filter;
+	float i_beta_filter;
 	float i_abs;
 	float i_abs_filter;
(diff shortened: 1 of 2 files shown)


Gabrielerusso:
I tested those modifications for more than 10000km on 3 differents ESC with firmware 6.05 and 6.06, i've now rebased the modifications on the current master version for easier testing by other users.

I've seen lower current (and power) peaks registered due to filtering, this does not impact real power consumption but just dampen spikes that are mostly not real but artifacts.
If someone is willing to try it and give a feedback it would be appreciated.

PR #877, opened 2025-12-20
Add the crash_diag terminal command for diagnosing resets and crashes

lukash:
Changes the fault handlers so that on a fault all the relevant registers (those dumped on the stack by the MCU and the fault information registers) are stored in a RAM struct (in the .noinit section) before resetting. Same with the Chibios halt, stores the pointer to the reason string before resetting.

On a bootup, the RAM struct is only initialized if Power On reset flag is set, and the reset flags are also stored in the struct.

This is all the possible information about why a reset could have occurred, which is then exposed via the `crash_diag` terminal command. Example output:
```
-> crash_diag 

Reset flags: Reset Pin,Software
System crashed, registers:
r0: 0x00000000
r1: 0x080728a6
r2: 0x00000000
r3: 0x20000800
r12: 0x50000020
lr: 0x080145a5
pc: 0x080145a8
psr: 0x61000000

cfsr: 0x00010000
hfsr: 0x40000000
mmfar: 0xe000ed34
bfar: 0xe000ed38
afsr: 0x00000000
shcsr: 0x00000000
```

With this output and the firmware .elf file (and in case of a package crash the package .elf) it should be possible to find the causes of a lot of in-the-field crashes and resets in general.

### DIFF
--- a/main.c
+++ b/main.c
@@ -294,10 +294,14 @@ static bool should_reset_config(void) {
 }
 
 int main(void) {
-	if (RCC->CSR & RCC_CSR_PORRSTF) {
-		// In case of Power On reset erase the whole struct
+	if (crash_info.magic != CRASH_INFO_MAGIC) {
+		// Erase the struct when the contents did not survive: first power-up
+		// from cold, SRAM decay after a long power loss, or a firmware update
+		// moving the section.
 		memset(&crash_info, 0, sizeof(crash_info));
+		crash_info.magic = CRASH_INFO_MAGIC;
 	}
+	crash_info.boot_count++;
 	crash_info.reset_flags = RCC->CSR;
 
 	// Clear the reset flags
@@ -481,6 +485,7 @@ static void stop_motor_and_reset(void) {
 void main_system_halt(const char *reason) {
 	crash_info.halt_reason = reason;
 	crash_info.type = CRASH_HALT;
+	crash_info.crash_boot = crash_info.boot_count;
 
 	stop_motor_and_reset();
 }
@@ -518,6 +523,7 @@ void fault_handler_c(uint32_t *hardfault_args) {
 	crash_info.registers.shcsr = SCB->SHCSR;
 
 	crash_info.type = CRASH_REGISTERS;
+	crash_info.crash_boot = crash_info.boot_count;
 
 	stop_motor_and_reset();
 }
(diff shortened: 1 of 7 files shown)


lukash:
@vedderb I've revised the PR, fixed a few issues and extended it a bit (to better cover mainly power delivery issues).

Especially in light of the mysterious IMU-related reset I'd really like to have this in 7.01, but it can be immensely useful in general to anyone dealing with crashes or resets.

I would actually like to extend this in the future for VESC Tool to check for the crash info and report it via a popup dialog. As of now, people having crashes/resets are completely in the dark, they can't really easily know if they even got a crash or what caused their controller to reset, which is pretty bad in a PEV, and especially if we're talking Onewheels (a crash is guaranteed).

I realize flash usage is a concern, so I tried to keep the flash increase minimal, but obviously code like this is going to take up some more space. However, I looked into flash usage and I'll submit a couple PRs that can free up thousands of kB (this PR adds ~1200 kB IIRC).

Note the last commit, which in addition to provide data from the PVD also enables it - the code was there to handle the fault, but the PVD itself wasn't enabled.

main.c
```diff
@@ -107,7 +109,7 @@ static THD_FUNCTION(flash_integrity_check_thread, arg) {
 
 	for(;;) {
 		if (flash_helper_verify_flash_memory_chunk() == FAULT_CODE_FLASH_CORRUPTION) {
-			NVIC_SystemReset();
+			chSysHalt("Flash corruption detected.");
```
vedderb, line 112: chSysHalt is problematic here. If the motor is running the PWM updates will freeze and result in a blown power stage almost instantly. If flash is indeed corrupted and we don't trust the machine code anymore the best thing that can be done here is a reset, which is what it did before.

vedderb:
Nice update, I think this information is worth having for sure. I had a quick look and gave one comment, but I will have a more careful later.

main.c
lukash: I believe this is not the case. We define `CH_CFG_SYSTEM_HALT_HOOK(reason)` as `main_system_halt(reason)` and that calls `stop_motor_and_reset()` at the end, so it's the same path as all the crash handlers etc. So this should be correct IMO.

vedderb: I think you are correct, this change should be fine then.

PR #892, opened 2026-03-11
Encoder: Add mechanical slip detection 

fbg-jpg:
This code is meant to fix situations where the magnet gets loose and causes the motor to rotate very fast. The solution is for encoders of type AS5047 and it compares the FOC observer phase against the physical encoder phase at high speeds and if they are not in sync than a loose magnet fault is displayed and the motor will stop.
https://github.com/vedderb/vesc_tool/pull/453


fbg-jpg:
I added new changes to match your suggestions. Now, the logic is configurable from the VESC Tool under the FOC -> Advanced tab. Also, the logic is now common for all encoders.

PR #899, opened 2026-04-04
Fix CAN baud rate being changed by unrelated CAN traffic

snowzach:
Move CAN_PACKET_UPDATE_BAUD handler into the sender-validated block so it only processes frames addressed to this VESC (id match) or broadcast (id=255). Previously it was in the unfiltered switch block, allowing any CAN frame with a matching packet ID (e.g. from a BMS) to inadvertently change and persist the baud rate.


vedderb:
Adding the length-check is fine, but the other things will break backwards compatibility which is something I really don't like doing. I also don't see this as a critical problem as you need to ensure that all things on a CAN-bus are compatible with each other anyway. There are many frames that can cause all sorts of problems (set current, reset things etc) if you have random but valid frames on the bus.

PR #902, opened 2026-04-26
Default phase filter settings in Makerbase 75_100 configs

jwidess:
This commit adds an #ifdef macro to the Makerbase `hw_mksesc_75_100_core.h` and `hw_mksesc_75_100_v2_core.h` hardware config headers to default phase filters to the appropriate setting (usually off) based on HW config. 

I added this as restoring defaults currently inherits the global fallback of true from `mcconf_default.h`. However, for config variants without phase filters, this seems to only affect the stored/default config value and not the actual behaviour.

I've compiled and flashed this to my own [MKSESC 75100](https://www.aliexpress.us/item/3256804277669942.html) from AliExpress, and it works fine, correctly defaulting phase filters to off on defaults.

### DIFF
--- a/hwconf/makerbase/75_100/hw_mksesc_75_100_core.h
+++ b/hwconf/makerbase/75_100/hw_mksesc_75_100_core.h
@@ -259,6 +259,13 @@
 #ifndef MCCONF_FOC_SAMPLE_V0_V7
 #define MCCONF_FOC_SAMPLE_V0_V7			false	// Run control loop in both v0 and v7 (requires phase shunts)
 #endif
+#ifndef MCCONF_FOC_PHASE_FILTER_ENABLE
+#ifdef HW_HAS_PHASE_FILTERS
+#define MCCONF_FOC_PHASE_FILTER_ENABLE	true	// Use phase voltage filters when available
+#else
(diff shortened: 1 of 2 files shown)


vedderb:
Have you confirmed that this actually makes any difference? If the hardware has no filters the setting should not do anything.

jwidess:
> Have you confirmed that this actually makes any difference? If the hardware has no filters the setting should not do anything.

Apologies, I should have clarified that yes, you are correct, this functionally does not change anything, as if the HW config does not enable phase filters, this setting is inert. It is only meant to default the "_Enable Phase Filters_" setting to false on Makerbase ESCs when they are not enabled to provide a bit of peace of mind and reduce confusion for new users. 

**EDIT**: This confusion is mainly aided by the [notes on sellers' pages](https://www.aliexpress.us/item/3256811594513844.html#nav-description:~:text=Phase%20filering%20is%20not%20available%20for%20MKSESC75100%20!) saying things like,
> Phase filering is not available for MKSESC75100 !
> Please turn off the phase filter function and choose "false" in vesc_tool when use firmware version on or above 5.3 (VESC_TOOL 3.01). Without turning off the phase filter will result in esc damage. Please do not restore the default parameters when using the wizard interface.

However as the firmwares properly disable phase filters this isn't true, but may confuse newbies.

PR #903, opened 2026-05-10
Fix LSM6DS3 polling, increase bus speed to 700 kHz for hardware that is known to support it

lukash:
This has cost me a lot of effort due to the MCU reset issue (which is real in the end) and the uncertainty of the 700 kHz bus speed working on all hardware.

Regarding the resets, I thought it was a hardware issue for a while (the controller that did it originally started to spuriously reset on impact), but I got the same thing on a different controller. I'll post more about it on discord. Trying to mitigate it by capping the IMU Sample Rate to well tested values (hundreds of kms ridden), but it doesn't integrate all too well in VESC Tool right now (unlike mcconf it doesn't check for a value being overridden).

Fot configuring the 700 kHz bus speed (which helps _a lot_), only enabling it for the controllers with verified pullup values, more support can be added. I'd like to do it for JF and Thors too (very common in Onewheels), in discussion with Fungineers right now.

Overall, had a very positive feedback from the few Onewheelers who tried it, they report a big increase in high-speed stability. Would like to get it in 7.0 if I'm in time and build more on top of it for future releases (Thor 301 and 400 support hardware SPI, I wanna add interrupt pin support and ultimately a modern IMU).

Note: Please don't squash the commits when merging


vedderb:
Nice work on this one and thanks for putting in all the effort!

@surfdado and @Mitchlol
Can you have a look at this too? To me it looks good and I'm happy to get this into firmware 7.0.

hwconf/shaman/Little_FOCer/hw_Little_FOCer_core.h
```diff
@@ -243,6 +243,7 @@
 #define LSM6DS3_SDA_PIN         2
 #define LSM6DS3_SCL_GPIO        GPIOA
 #define LSM6DS3_SCL_PIN         15
+#define LSM6DS3_SPEED_700KHZ	1
```
vedderb, line 246: Setting this one to 1 here is a bit confusing as you check of it is defined and not if it equals 1. This means that even if you set it to 0 it will still run on 700kHz as it still is defined then.

I would propose just defining it and not setting it to 1.

Mitchlol:
Sorry I don't have context to judge this, I haven't looked at an IMU datasheet or this code in years. I would need a pairing session explaining it to me to understand what these changes are doing and why.

vedderb:
> Sorry I don't have context to judge this, I haven't looked at an IMU datasheet or this code in years. I would need a pairing session explaining it to me to understand what these changes are doing and why.

That is how I feel about the IMU code too, that is why I'm asking for some help. I'm also happy to just merge it and give it a quick try on our hardware.

lukash:
> Sorry I don't have context to judge this, I haven't looked at an IMU datasheet or this code in years. I would need a pairing session explaining it to me to understand what these changes are doing and why.

I'm happy to explain anything. The commit messages should have the important info but some context may be missing, I can also fill it in if you point out deficiencies.

That said I think I have a good enough understanding and the code was kind of tested to death due to the issues that delayed posting this, I've been running this in my boards for a year.

But anyway feel free to ask for any form of explanation, just right now I'm stressed for time, leaving home for a week and gonna be busy most of the week after that as well.

hwconf/shaman/Little_FOCer/hw_Little_FOCer_core.h
lukash: Agreed, I'll address it later if that's fine, as I said gonna be busy the next two weeks, I rushed posting this a bit.

PR #909, opened 2026-06-10
hw_mksesc_84_100_hp.h change to correct voltage reading

WeDo-ESC:
Minor change to vin_r2 to correct voltage divider reading, thoroughly tested also fixes a lot of hang ups  before correction controllers would crash and have multiple voltage errors minor change seems to have fixed.


WeDo-ESC:
closing due to commit errors

PR #916, opened 2026-07-07
Add support for MagnTek MT6835 21-bit SPI magnetic angle encoder

adilrepas:
This PR adds support for the MagnTek MT6835 21-bit magnetic angle encoder
via hardware SPI, standard 4-Wire SPI


vedderb:
Looks good!

Can you put the types ENCODER_TYPE_MT6835 at the end of the structs? That will avoid breaking old motor configurations.

adilrepas:
Okay, I'll do it when I get home. 
I forgot to change it after testing, I was using the unmodified version of Vesctools, so to test the encoder, I added it to the list of encoders already in Vesctools.

adilrepas:
> Looks good!
> 
> Can you put the types ENCODER_TYPE_MT6835 at the end of the structs? That will avoid breaking old motor configurations.

I moved ENCODER_TYPE_MT6835 to the end of the enum. Also fixed a compile warning

PR #917, opened 2026-07-08
IMU device stack refactor, new features and devices

lukash:
This is a big one, my apologies for that. Feels like it should be split into multiple PRs, but some of the late commits tie up some loose ends and what would be left is just some short commits enabling hardware features. I'm also (again) short on time and wanted to post this rather than sit on it. But I can definitely split it up to smaller PRs if desired.

Hopefully it'll be digestable. Self-contained clean single-change commits from beginning to end, so readable commit by commit or as a whole. A lot of the churn is updates to all hwconfs. Definitely recommend reading the main device/transport setup files in /imu first to get the idea.

Heavily AI-assisted (otherwise it'd take me many weeks to finish something like this), carefully designed by me along with the AI and most of the code carefully reviewed and also cross-checked with AI. What I didn't review in detail is the legacy ICM and MPU IMUs migration to the new structure. AI has gotten really good at validating code changes without changing behavior and I pretty much fully trust it on that. I don't know what is the status of the old IMUs, considering them not very important nowadays, but best to test them if you got the hardware and can be bothered. I think it's well worth the cleanup to migrate them and assume the small risk something could break.

What the PR brings:
- Big deduplication of code (the sampling thread, transport layer implementation and a lot of the setup boilerplate)
- A new set of macros for configuring the IMU which allows much more straight-forward device-based dispatch (that's the update to all hwconfs), with fallback definitions to support the old macros (for any out of tree configs out there)
- DRDY interrupt support added as a generic module which can be easily enabled for any IMU
- Support for the LSM6DSV32X IMU (shows it's now quite a bit easier to add a new IMU behind any of the three possible transports)
- Generic `imu_status` and `imu_read_reg` commands, `imu_status` being especially nice and printing almost all the info (just missing the BUS speed, possible to add later):
```
-> imu_status 

Model         : LSM6DS3
Variant       : TR-C
Transport     : spi-hw
Running       : yes
Sample Rate   : 3310 Hz
Read fails    : 0
```

Biggest concern is probably the changes to EXTI interrupt handling needed for the DRDY support. I've done my best covering all bases but you know more than me about the SPI Encoder to see if the changes won't mess it up. Hopefully it'll be good. Note I'm removing the `HW_ENC_EXTI_CH`/`HW_ENC_EXTI_ISR_VEC` macros and I also found `HW_ENC_EXTI_CH2`/`HW_ENC_EXTI_ISR_VEC2` macros on some dual-encoder controllers, but those are unused throughout the codebase. I left those in.

Not sure how the new set of IMU config macros will be received, but IMO the simplified setup is well worth the change and there shouldn't be any breakage with the legacy macros support.

A small hurdle when doing this was the fallback I2C commit which landed about a week ago. I kept the added support for that in a separate commit. Gotta say I'm not a fan of that mechanism, the commit ([imu: Add IMU_FALLBACK_ macros to support an optional secondary bus](https://github.com/vedderb/bldc/commit/7da5da236fd3fad9053510197049f3aba03d4e50)) shows it adds quite a bit of churn and doesn't look very clean. Have you considered doing it via a different HWNAME (e.g. `duet_xs-v2`)? I think it would be cleaner...

I'm probably forgetting to mention a lot of important things but the text is too long already. All the commits are well described and list important changes and some very minor lost functionality (e.g. around terminal commands, the old IMUs etc.).

Tested hardware (on almost the final version, I'll be testing more and I'll update the list):
- JetFleetF6_32S
- Thor301
- Thor400v2
- Lightning400 (with DRDY)
- Little_FOCer
- LSM6DSV32X IMU connected externally via HW SPI and DRDY

Anyway I'm pretty happy with how this came out, this in my eyes sets the IMU support straight and makes it ready for HW SPI and the DRDY interrupt on any hardware on which the maker would like to get first class fast IMU support. Few minor improvements possibly remain (like the IMU thread priority), but those should be small changes.

Anything you need me to change or do to make this easier to get in, don't hesitate to let me know.


vedderb:
You have to understand that this is almost impossible for me to review and test... Also, braking the I2C fallback code means that at least the Duet XS will be broken after this change. I made it into the same hwconfig precisely because I wanted to avoid having different configs and firmware for it when the PCBs already are difficult to tell apart when we flash and test them. Messing with DMA priorities also has potentially dangerous side effects for motor control.

I will look into it more carefully soon, but my initial reaction is that there is no chance that I can review, test and merge this without spending a full week or two of full time effort on it, which is difficult to justify when there are millions of other things that I'm falling behind on... Also, there are many many hardware that potentially will break that I can't even test as I don't have access to them...

vedderb:
> Big deduplication of code (the sampling thread, transport layer implementation and a lot of the setup boilerplate)

Yet there is far more code now than there was before...

> DRDY interrupt support added as a generic module which can be easily enabled for any IMU

Is that the main new feature? Is there a simpler way of enabling this other than this enormous change?

> Support for the LSM6DSV32X IMU (shows it's now quite a bit easier to add a new IMU behind any of the three possible transports)

Same question here... Is there a simpler way to enable this? The LSM6x code already had abstractions for the three types of transport and there was no duplication between them.

This also follows the pattern of wasting my time using AI of recently. Here are a million things I had AI make and I trust AI, so you deal with the vast effort of reviewing and testing it...

lukash:
> You have to understand that this is almost impossible for me to review and test...

I understand, sorry about that. I was hoping you'd see the overall benefit of the cleanup and I'm testing what I can myself. Definitely possible to split into smaller PRs, more work for me though especially assuming I'd need to make them fully self-contained (and in theory more testing needed that way).

> Also, braking the I2C fallback code means that at least the Duet XS will be broken after this change. I made it into the same hwconfig precisely because I wanted to avoid having different configs and firmware for it when the PCBs already are difficult to tell apart when we flash and test them.

I haven't broken the I2C fallback, it's in place and should be working (obviously couldn't test). I assumed you'd want to keep it just wanted to bring it up... On a very small side note I think the PCBs should be distinguished with a version since the SPI allows higher IMU speeds which can make a difference for the user, they should be able to tell the versions apart. Which speaks for a special HWNAME. But anyway, keeping the fallback, it's not a problem.

> Messing with DMA priorities also has potentially dangerous side effects for motor control.

Can't say I know too much about that. I only raised the one FOC ADC priority, there was nothing else higher. Seems this should be no problem.

> Also, there are many many hardware that potentially will break that I can't even test as I don't have access to them...

I think this is specific to the ICM and MPU IMUs? I have no clue how relevant those are nowadays? The rest I'm sure we can verify on other hardware - it's pretty much just the IMU code which is being changed (in addition to the EXTI interrupts but I think we can test that too as it's new stuff?).

lukash:
> > Big deduplication of code (the sampling thread, transport layer implementation and a lot of the setup boilerplate)
> 
> Yet there is far more code now than there was before...
> 
> > DRDY interrupt support added as a generic module which can be easily enabled for any IMU
> 
> Is that the main new feature? Is there a simpler way of enabling this other than this enormous change?
> 
> > Support for the LSM6DSV32X IMU (shows it's now quite a bit easier to add a new IMU behind any of the three possible transports)
> 
> Same question here... Is there a simpler way to enable this? The LSM6x code already had abstractions for the three types of transport and there was no duplication between them.
> 
> This also follows the pattern of wasting my time using AI of recently. Here are a million things I had AI make and I trust AI, so you deal with the vast effort of reviewing and testing it...

The change is a big cleanup, deduplication and proper structure design. It is obviously possible to do the same support through the "minimum change" path. However that slowly deteriorates the codebase, brings edge cases and more bug potential. And, to be honest, it's not how I want to work, I want to do things properly and in a clean way. I think this PR is a huge improvement in that regard. Was kinda hoping you'll see it in a similar way...

vedderb:
The best way to not brake old boards that used to work is to not touch their implementation. The minimal change approach I follow is to keep changes as contained as possible and carefully test everything they touch on every board. This has some implications how how the codebase evolves compared to regularly starting over and changing everything, but it does not make a huge difference. For example, after this change which touches everything about the IMUs, the FOC code and the encoder code and ALL hardware files it is not obvious to me that it now is easier to add support for a new IMUs. If anything, there are now more abstractions to understand and deal with. Before there was already the abstraction of extracting the samples from the chip and handing them to the IMU filter. Now the abstraction include the transport layer heavily and there might be more transports than just I2C and SPI to consider. With the old approach it would be trivial to add a new bus if needed, but now you have wrap it in abstractions if a new IMU has something other than I2C or SPI.

Another thing to consider is: WHY would you want to support lots of IMUs? Ideally there is only one IMU you use for new designs and when that becomes obsolete (such as the BMI160, MPU9150 etc.) you add support for the next best one and keep support for the old ones as well as you can. Old boards are inherently difficult to support as they become more and more rare. I feel strongly about not braking old boards that used to work and there are only two ways of avoiding that:

1) Test every change carefully on every old board
2) Avoid changing things affecting old boards that you can't test easily

As 1) takes vast work and effort I usually stick with 2) unless there is a very good reason to potentially break old boards. And looking at this change for the MPU9150 there is enormous risk of braking old boards and "I trust AI to not break them" is simply not good enough to me.

On another topic, the usual things I ask about are also not covered:

* How does the binary size change?
* How does the memory usage change?
* Do all hardwares compile?
* How is the performance affected?

If this change e.g. reduces binary size, reduces memory usage or speeds things up it becomes far more appealing, but from just looking at the changes I suspect that it just makes the binary and memory usage grow.

lukash:
> The best way to not brake old boards that used to work is to not touch their implementation. The minimal change approach I follow is to keep changes as contained as possible and carefully test everything they touch on every board. This has some implications how how the codebase evolves compared to regularly starting over and changing everything, but it does not make a huge difference. For example, after this change which touches everything about the IMUs, the FOC code and the encoder code and ALL hardware files it is not obvious to me that it now is easier to add support for a new IMUs. If anything, there are now more abstractions to understand and deal with. Before there was already the abstraction of extracting the samples from the chip and handing them to the IMU filter. Now the abstraction include the transport layer heavily and there might be more transports than just I2C and SPI to consider. With the old approach it would be trivial to add a new bus if needed, but now you have wrap it in abstractions if a new IMU has something other than I2C or SPI.

I kinda understand why you have that PoV but I disagree with it (based on my long experience in the industry). But I feel we've had arguments exactly like this before and I don't want to get into another one. While I'd like you to see my point of view I was never able to convince you and don't think this time would be different, so I think I'm just giving up. IMO the new code is much cleaner and speaks for itself.

> Another thing to consider is: WHY would you want to support lots of IMUs? Ideally there is only one IMU you use for new designs and when that becomes obsolete (such as the BMI160, MPU9150 etc.) you add support for the next best one and keep support for the old ones as well as you can. Old boards are inherently difficult to support as they become more and more rare.

You've partly said it yourself. You need to keep adding more IMUs, due to obsolescence. Since you keep support for the old ones, we end up supporting a lot of IMUs. There's also part scarcity. The reason I'm adding the new one (LSM6DSV32X) is it's much better specced and has bigger scale and better vibration rejection, which I think matters for Onewheels. We want to move forward, improve behavior, get rid of nose hunting and vibrations.


> I feel strongly about not braking old boards that used to work and there are only two ways of avoiding that:
> 
>     1. Test every change carefully on every old board
> 
>     2. Avoid changing things affecting old boards that you can't test easily
> 
> 
> As 1) takes vast work and effort I usually stick with 2) unless there is a very good reason to potentially break old boards.

Understood. But long-term, you are fighting the dilemma of aged suboptimal codebase and not being able to innovate properly due to not wanting to change things. At some point you need to solve this dilemma.

One way could be to drop support for the old stuff at some point. Old HW can still use older VESC releases for the rest of its lifetime... (but as I said, don't know how relevant the old IMUs are nowadays and if it's too early to drop support).

> And looking at this change for the MPU9150 there is enormous risk of braking old boards and "I trust AI to not break them" is simply not good enough to me.

Fair enough. Do you have any hardware to test it? IMHO if you don't, it's a bad situation either way.

> On another topic, the usual things I ask about are also not covered:
> 
>     * How does the binary size change?
> 
>     * How does the memory usage change?
> 
>     * Do all hardwares compile?
> 
>     * How is the performance affected?
> 
> 
> If this change e.g. reduces binary size, reduces memory usage or speeds things up it becomes far more appealing, but from just looking at the changes I suspect that it just makes the binary and memory usage grow.

Haven't had the time to check, I don't expect any sizable savings in any of the above, except IMU performance should increase due to the improvements (which are arguably doable in the old codebase, just duplicate across 4 -> 5 IMUs separately and in slightly different ways). I can check all of those later but don't think there'll be any significant movement either way.

vedderb:
I downloaded the changes into a branch and I'm testing them now. I was excited to see that the binary size actually went down slightly. Will report back when I have done more testing on VL hardware.

vedderb:
BTW, did you solve the mystery of the firmware update crashing the CPU when using SPI1 for the IMU?

lukash:
> BTW, did you solve the mystery of the firmware update crashing the CPU when using SPI1 for the IMU?

Ah, yes. Meant to tell you, sorry. The CAN process was getting stuck on the DMA because the buffers were on it's stack, which is in CCM RAM (solved by placing the buffers in the transport instance in the PR). Trying to remember, I think that's all there was to it in the end, the DMA sharing is just about the priority and working for me without an issue now on Thor400v2 (rode it for about 10km so far).

lukash:
I'm leaving for an extended weekend trip so not gonna be very responsive until Tuesday.

I think I owe another apology for dumping this on you like this, I am also really time-limited and while I tried my best on the code itself, cutting corners on the process :sweat_smile: Anyway, no rush on you right now, you can think it through. I just really wanted to get it out so you can start having a look / thinking about it, and know not to make unnecessary changes that'd cause me conflicts. If you accept it, we can figure the details on how to proceed later...

I do value your time, thank you.

vedderb:
I did some more testing now and so far I really like it.

* All firmwares still build without warnings or errors
* The binary size went down slightly
* The Duet XS60 works with both the regular and fallback mechanism
* The BMI160 with SPI still works on an old ESC from my junk box
* The MPU9250 also works, including the magnetometer, on another old ESC from my junk box
* The hw and bb SPI transports work on VL hardware

The code is much cleaner like this for sure and I like that there only is one file with the IMU thread now that deals with all implementations. Super nice work!

I haven't been able to test the implications of changing the DMA priorities yet, but I think I will merge it anyway and start testing. This is just the beginning of beta 7.01, which is a good time to introduce a change like this.

lukash:
Thank you! I'll have some small followups (someone already noticed the deinit interface hook is not being called and the deinit is done elsewhere - not an immediate functional issue but I'll sort it out), I have still some HW to enable faster I2C on etc., also wanna look into the IMU thread priority but that may not be for immediate inclusion.

Anyway wanted to ask what is the tentative plan for 7.01 beta so I can plan accordingly and if you could coordinate the release with me, would like to make sure everything is in place for the IMUs on onewheels.

vedderb:
Generally I keep the beta updated to the latest master, but I haven't made the config split yet for 7.01 as I was still improving and backporting to 7.00. Probably I will do it this weekend. I can let you know when the first beta is out.

vedderb:
Beta 7.01 is now ready and this PR is included!

PR #923, opened 2026-07-25
Enable 700 kHz I2C on more Onewheel-relevant controllers

lukash:
Enable 700 kHz I2C on most Onewheel-relevant controllers left out of the original PR. The pullups on all of them were verified with the manufacturers.

Thor300 / JF F6 are potentially problematic with 10k pullups, never manifested an issue in practice though. The I2C BB lib also enables the internal 40k pullups, which bring the overall resistivity down a bit more.

I also inquired about the ADV500 (I think it's pretty safe to assume the pullups are the same as on ADV200 which is a copy of Little FOCer v3 but wanted to confirm) but still waiting for a reply so posting these now and I'll post ADV500 once I get a response.

Please don't squash commits.

I don't see a 7.01 branch yet so assuming it'll be branched off master later and these will get in.


lukash:
After the confirmation of ADV500 pullups, adding the enablement for that too (plus a small cleanup).

PR #933, opened 2026-08-02
Reduce flash usage

lukash:
The SUBALIGN commit is a massive saving in flash usage. Maybe I'm not aware of some optimizations or other reasons why this was in place. The situation is also complicated by a GCC quirk forgetting the -falign-functions if you use `#pragma GCC optimize`. I've added explicit alignment to some hot functions where the removal of the SUBALIGN(16) broke their alignment due to that quirk.

The `no-math-errno` looks free.

If you'd just want to keep one of the commits (or any other edits, obviously), let me know.

Please don't squash commits :grin:

### DIFF
--- a/encoder/encoder.c
+++ b/encoder/encoder.c
@@ -469,7 +469,9 @@ void encoder_set_custom_callbacks (
 
 #pragma GCC pop_options
 
-float encoder_read_deg(void) {
+// The optimize pragma above disables -falign-functions=16 for the whole file,
+// even after pop_options. Align the hot function explicitly.
+__attribute__((aligned(16))) float encoder_read_deg(void) {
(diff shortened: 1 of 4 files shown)


vedderb:
This is a nice find, didn't realize that subalign makes such a difference! It comes from the example linker script I used years ago and I never questioned it.

I did some profiling with the foc_profile.lbm script and it seems like the interrupts finish at the same time as before (maybe even slightly faster), so there seems to be no downside to performance.

The performance of the FOC interrupt finishing in time is a bit of a problem generally as well, so if you test before and after with foc_profile.lbm while the motor is running it would be great. Any changes that make the interrupt finish faster are very welcome.

PR #934, opened 2026-08-02
Update GCC to 14

lukash:
I've wanted to update the GCC for some time... I finally went to look for what causes the flash usage increase in recent GCCs, it's basically big chunks of code being pulled even though they are not really needed. The branch updates the GCC, fixes the warnings that pop up (I'd prefer the 3rd party libraries themselves were updated, but I'm not gonna do that due to time constraints and not knowing how to test them, the actual fixes are not large in size and shouldn't be much of an issue just patching the warnings in the old lib versions) and remove the extra unneeded code being pulled in.

In summation this PR reduces flash usage by close to 2.5 kB. In combination with https://github.com/vedderb/bldc/pull/933 the reduction is even higher (I think - AI measured the numbers, I ended up not verifying to make 100% sure).

The in-repo toolchain setup was untested though, please test it if I can ask that. It's nontrivial running downloaded binaries for me, I could work around this, but it'd take more time for me.

The toolchain update should eventually enable more codebase modernization and potential optimizations/speedups, like LTO (probably only after a ChibiOs update). (in addition to being used for other parts of the ecosystem, like packages)

Regarding stability concerns: I'm sure recent GCCs are in a very good shape with a huge user base and I believe they are in general very stable. Historically it may have not been the case, but that was long time ago... I've been running the firmware compiled with GCC 10 for years and I don't think there'll be any breakage going from 10 to 14. I'll be switching to 14 shortly (having this merged would make things a bit smoother).


vedderb:
Nice work! If you resolve the conflict I will merge it and start testing.

lukash:
> Nice work! If you resolve the conflict I will merge it and start testing.

Thanks! Yeah I got the conflict fixed and went to re-check my binary size claims, only to find out I mixed up the results between GCC 7 and 14. The binary size actually increases by about 1k with this PR. The savings in better compile optimizations is outweighed by some added "bloat" (well, not sure to which extent it's bloat and to which extend it's "needed functionality") to newlib. IMO still well worth the update (there's no long-term way forward with an old GCC), but quite a bit less attractive.

Also went to run the foc-profile, actually found it looks tiny bit slower (only ran it once, not sure there's variance but I wouldn't expect it), which has really surprised and disappointed me. So planning on having a bit closer look on that.

Pushed the rebase.

vedderb:
Yeah those results are roughly what I saw last time I tried. It got a little bit slower and much bigger at the time, which is why I wasn't at all excited to update. With your tweaks the extra 1k size is not a problem IMO, but I don't like that it is running slower. Will be interesting to see if you can find out what is causing that. Until then I think I will stick with the older GCC.

lukash:
I've found the regression (although spent way too much time on this :sweat_smile: ). It's in the added top commit (I also rebased, sorry for the force-push noise, the rest of the commits is unchanged).

foc-profile comparison (mean of 10 measured samples) Max FZV:
- GCC 7.3: mean 41442 Hz (spread 40698–42296 Hz)
- GCC 14.3 (default options): 40327 Hz, −2.7 % (spread 37940–41791 Hz)
- GCC 14.3 -freorder-blocks-algorithm=simple: mean 41656 Hz, +0.5 % (spread 41176–42042 Hz)

GCC 14's default stc basic-block reordering scatters the hot FOC code over a wider address range. Executing from flash the added taken branches to uncached flash lines cost more execution time and it varies pass-to-pass. The default options build shows both a lower mean and much larger run-to-run jitter. With the "simple" block reordering the interrupt runs slightly better than GCC 7 performance and has the tightest timing distribution.

Also the image shrinks by ~2.2 KB with this option so it's again slightly smaller than with GCC7, so win-win.

I ended up testing this while running Refloat (to exercise the instruction cache with more code from different parts of flash being executed) and ran into an inconsistency in the FOC interrupt execution time which I tracked to https://github.com/vedderb/bldc/pull/936. The first tone played from the package would knock down the performance by ~5% and it would never restore which took some time to differentiate from the impact of the optimizations I was testing.

I suggest to also check the performance impact of -freorder-blocks-algorithm on the lisp interpreter, there should definitely be some markable change (well, with the whole branch as well).

Also note the second to last commit ("Suppress new linker RWX segment warning") will break building with old GCCs which don't know the switch, so maybe leave that out (that one-line warning is quite harmless) and add it after a while when 3rd parties have migrated to new GCCs. Let me know and I'll drop it from the series.

lukash:
I also thought this could be informative for you, AI output warning. This is a dump of what I tried and a few other opportunities flagged by AI.

> Tried, no measurable effect:
> - Grouping the FOC interrupt call tree in flash. Everything the interrupt calls per cycle was placed into a dedicated section, laid out back-to-back (normally these functions end up scattered across ~90 KB of flash, vs. the 1 KB flash cache). Measured zero difference in every scenario tried — idle, motor running, and with a busy lisp package evicting the cache between interrupts. Call-target distance apparently just doesn't matter on this cache.
> - Branch prediction hints. cold on mc_interface_fault_stop (makes every fault-check branch in the interrupt predicted not-taken) plus unlikely on the phase-override branches. Occasionally looked like a small win but never reproducibly above the measurement noise (~±15 cycles).
>
> Checked, nothing there:
>
> - The -Os pragma regions in mcpwm_foc.c / mc_interface.c contain almost nothing the interrupt executes — only trivial getters, where -Os compiles to the same code as -O2 anyway.
>
> Untested, the remaining candidates:
>
> - Running the interrupt code from RAM — would remove flash wait states entirely, and the measurements say code fetch is a meaningful fraction of the interrupt time. Needs ~12–16 KB of RAM for the call tree plus linker/startup support, so it's a real trade on RAM-tight boards.
> - Targeted source tweaks using the existing FOC_PROFILE_LINE_FINE markers to find them; the obvious remaining costs are the float divisions/sqrts (14 cycles each) and timer register reads.
>
> Tried and best avoided:
>
> - LTO: the link fails (fault_handler_c is referenced only from inline asm) and code size grows.
> - -O3 or more inlining: bigger code is the wrong direction for a flash-cache-bound interrupt.
> - Fast-math options beyond -fno-math-errno: -ffinite-math-only breaks the NaN-based checks (UTILS_NAN_ZERO), and reassociation changes the filter/observer numerics.
>
> General takeaway from all the measurements: the interrupt's speed is governed more by flash code layout than by instruction count, but beyond "keep the layout compact" (which the reorder flag does) the effects behave like luck — equivalent layouts can differ by a few percent per link. That's also where the run-to-run jitter of the default GCC 14 build comes from.

lukash:
And while I'm at it, in similar vein what I got for the flash size optimizations (again, AI output pretty much verbatim), in case you'd find it useful:

> All measured on GCC 14:
> - global -Os (alignment kept): −36728 B. slows everything incl. FOC loop + lisp eval — not shippable as-is
> - hybrid: -Os except motor/math/lispBM at -O2: −7912 B. near-zero performance impact: covers ChibiOS, comm, libcanard, imu, encoders, confgenerator — none cycle-critical
>   - …additionally lispBM at -Os: ~−27000 B more. lisp eval slowdown (10–30 % typical for -Os); a real trade for package-heavy boards
> - drop -falign-functions=16: −6984 B. hot-function entry alignment gone (the flag exists for FOC fetch determinism)
> - -falign-functions=8 instead: −4184 B. halves the padding, keeps 8-byte entry alignment; roughly stacks with hybrid -Os
> - LTO: dead end. measured: link fails (fault_handler_c referenced only from naked asm), .text2 blows up to 99.6 % from cross-TU inlining, and flash2 saved nothing anyway
>
> The decomposition of the global -Os number is telling: ~27 K of it is lispBM alone, ~2 K motor/math, and ~8 K everything else. So the hybrid is the standout: −8 K for a ~6-line change (target-specific CFLAGS + -O2 overrides after the rules.mk include, which I verified works with this build system), touching nothing performance-sensitive. The lispBM knob is a separate, conscious decision. One caution if implemented via upstream's #pragma GCC optimize style instead of makefile flags: the pragma-before-first-function quirk strips -falign-functions TU-wide, so hot functions sharing a file need the aligned(16) treatment we already gave mc_interface/encoder.
>
> Not measured but estimable, in descending value:
>
> - blackmagic opt-out flag (~24 K): a per-board HW_NO_SWD_PROG-style exclusion — it's dead weight for boards that never flash other targets. Feature decision, moderate makefile+bm_if stubbing work.
> - Drop -u _scanf_float (~1–2 K): requires converting %f sscanf to strtof across ~10 files first (terminal, apps, DRV drivers, bm_if) — otherwise float parsing silently breaks.
> - newlib scanf/stdio chain (~0.9 K: ungetc/fflush/findfp): only reachable by hacky overrides of newlib internals; not recommended.

(I was suprised by the LTO result but didn't investigate)

vedderb:
Very nice work on this one, this definitely make it worth upgrading! Also many important insights on performance bottlenecks and memory usage from the work you did here.

Improving performance even more soon would be nice as it is a bit of a bottleneck especially on dual motor hardware. Might be worth sacrificing some RAM for hot parts of the code if that helps.

vedderb:
Downloading gcc works fine for me with make arm_sdk_install. It is also very convenient that changing to the old release also uses the old gcc automatically. That makes backporting features much easier.

Now I will do some testing and compare the performance.

vedderb:
My FOC performance tests and memory usage line up with your results. I also tested the lispBM-performance using speed_test_tak_q2.lbm and I saw good improvements there.

Before
```lisp
("Time tak:" 3.414500f32)
("Time q2:" 8.831300f32)
("Time gc:" 0.001180f32)
```

After
```lisp
("Time tak:" 3.072800f32)
("Time q2:" 8.148500f32)
("Time gc:" 0.001160f32)
```

lukash:
Nice ~10% improvement on the LISP, next time I'll know what to run. Happy the new GCC worked out!

PR #935, opened 2026-08-05
Add BM60 and BM75 firmware targets

bmyyqzs:
## Summary
- Add BM60 and BM75 hardware configurations.
- Register BM60 and BM75 in firmware packaging.
- Prefer exact board source files in the Makefile target lookup before falling back to shared core files.

## Testing
- Not run in this planning pass.


Makefile
```diff
@@ -129,8 +129,11 @@ define FIND_TARGET_C_CODE
    # Remove `_no_limits`
    $(eval ROOT_TARGET_NAME = $(subst _no_limits,,$(2)))
 
-   # Look for `*_core.c` file
-   ifneq ("$(wildcard $(3)/hw_*_core.c)","")
+   # Prefer the exact board source file when it exists.
```
vedderb: If you have separate boards not sharing a core-file you should put them in separate directories. I'm afraid that this change will break some other builds otherwise.

bmyyqzs:
The requested change has been addressed. The Makefile is unchanged now, BM60 and BM75 are in separate hardware directories, and the PR has been squashed into one commit. Both `make BM60` and `make BM75` build successfully.

